PR #6, opened 2024-10-25
Add release workflow

ThoFrank:
Adds a release workflow which is triggered every time a tag matching `v*.*` is pushed.

Documentation matching the commit is built and attached.

Release Notes are empty for now. We need to think about how we want to handle them in the future.
It is always possible to add them manually after the action ran.

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -1,7 +1,6 @@
-name: Build and Deploy Documentation
+name: Build Documentation
 
-on: [push, pull_request]
-
+on: [workflow_call, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


marcmo (approved):

.github/workflows/doc-build.yml
```diff
@@ -1,7 +1,6 @@
-name: Build and Deploy Documentation
-
-on: [push, pull_request]
+name: Build Documentation
 
+on: [workflow_call, pull_request]
```
marcmo, line 3: why did you remove the `on push`?
when exactly does the `workflow_call` get triggered? I remember to read that people had problems with the `workflow_call` trigger

ThoFrank: It's removed here, because I added it to the `Publish Documentation` workflow.
So this way on a PR the documentation is built, but not published anywhere.
And only on a push to main the `Publish Documentation` workflow is called (which triggers the build by using the `Build Documentation` workflow).

ThoFrank:
Yes `v1.0.0` works too. You can see that in my fork.
`v*.*` just enforces there is at least one `.`.

.github/workflows/doc-build.yml
ThoFrank: workflow_call is required for reusable workflows:
https://docs.github.com/en/actions/sharing-automations/reusing-workflows#creating-a-reusable-workflow

PR #12, opened 2024-11-06
GitHub action for code coverage 

SuhashiniNaik:
This GitHub Action automates the generation and publishing of code coverage reports of the modules. It captures code coverage data using lcov, filters out third-party, mock files, gtest folders and system libraries.
Then it generates an HTML report that is deployed to GitHub Pages. 

Changes made:
Added code-coverage.yml
Added CodeCoverage.cmake
Made changes in CMakeLists.txt

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -1,0 +1,67 @@
+name: Code Coverage Report
+
+on: [workflow_call, push, pull_request]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Debug
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Setup cmake
+      uses: jwlawson/actions-setup-cmake@v2
+      with:
+        cmake-version: '3.22.x'
+  
+    - name: Test
+      # Execute tests defined by the CMake configuration.
+      run: |
+          cmake -B cmake-build-unit-tests -S executables/unitTest -DBUILD_UNIT_TESTS=ON -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}}
+          cmake --build cmake-build-unit-tests -j4
+          ctest --test-dir cmake-build-unit-tests -j4
+    
+    - name: Install lcov
+      run: sudo apt install lcov
+    
+    # no-external command excludes the system libraries in the root directory 
+    - name: Capture code coverage
+      run: |
+        echo "Capturing code coverage..."
+        lcov --no-external --capture --directory . \
+             --output-file cmake-build-unit-tests/coverage_unfiltered.info
+
+    - name: Filter out 3rd party and mock files
+      run: |
+        echo "Filtering out 3rd party and mock files from coverage data..."
+        lcov --remove cmake-build-unit-tests/coverage_unfiltered.info \
+             '*libs/3rdparty/googletest/*' \
+             '*/mock/*' \
+             '*/gmock/*' \
+             '*/gtest/*' \
+             '*/test/*' \
+             --output-file cmake-build-unit-tests/coverage.info
+
+    - name: Generate HTML coverage report
+      run: |
+        echo "Generating HTML coverage report..."
+        genhtml cmake-build-unit-tests/coverage.info \
+                 --output-directory cmake-build-unit-tests/coverage
+    
+    
+    - name: Zip the coverage report
+      run: |
+        mv  cmake-build-unit-tests/coverage code_coverage
+        zip -r code_coverage.zip code_coverage
+
+    - name: Upload code coverage artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: code_coverage
+        path: code_coverage.zip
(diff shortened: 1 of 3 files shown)


.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Filter out 3rd party and mock files
+      run: |
+        echo "Filtering out 3rd party and mock files from coverage data..."
+        lcov --remove cmake-build-unit-tests/coverage_unfiltered.info \
+             '*libs/3rdparty/googletest/*' \
+             '*/mock/*' \
+             '*/gmock/*' \
+             '*/usr/include/*' \
+             '*/gtest/*' \
+             --output-file cmake-build-unit-tests/coverage.info
```
marcmo: rather than exclude a whole list of locations, only exclude the mock/test patterns and instead use options like:
```
	•	--directory: Specifies the directory containing .gcda and .gcno files.
	•	--capture: Captures coverage data.
	•	--include/--extract: Includes files matching specific patterns.
	•	--remove: Excludes files matching specific patterns.
	•	--no-external: Excludes coverage data for files outside the specified directory.
```

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,66 @@
...
+
+    - name: Build
+      run: cmake --build cmake-build-posix --target app.referenceApp -j --config ${{env.BUILD_TYPE}}
+
+    - name: Test
+      # Execute tests defined by the CMake configuration.
+      run: |
+          cmake -B cmake-build-unit-tests -S executables/unitTest -DBUILD_UNIT_TESTS=ON
+          cmake --build cmake-build-unit-tests -j4
+          ctest --test-dir cmake-build-unit-tests -j4
```
marcmo: since we run the tests in this action, we should remove the test execution in the `build.yml` file

ThoFrank:
I think this action should be split into parts. Similar to the documentation action:

1. Build the coverage, zip it and upload the artifact. Should only run on PR's. This way you can still view the output by downloading the zip.
2. Download the zip, unpack it and publish it. Should only run on push to master (and as a dependency to action 1).

Action 2 could be combined with the doc publish action.

Additionally action 1 could also be integrated into the release workflow. But probably this should be done in another PR, because the permissions in the release workflow need to be fixed.

ThoFrank:
@marcmo Right now the link is only available in the action log. But we could look into posting the link as a comment to the pr.

For an example see [here](https://github.com/eclipse-openbsw/openbsw/actions/runs/11724559687/job/32658791048) -> `Upload docs`

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,60 @@
+name: Code Coverage Report
+
+on: [push, pull_request]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Release
```
vladyslavmarkovaccenture: Configuring the project with
```
cmake -B cmake-build-unit-tests -S executables/unitTest -DBUILD_UNIT_TESTS=ON
```
Without setting build type to "Debug" will give the following output
```
CMake Warning at openbsw/admin/cmake/CodeCoverage.cmake:214 (message):
  Code coverage results with an optimised (non-Debug) build may be misleading
Call Stack (most recent call first):
~/openbsw/CMakeLists.txt:35 (include)
  CMakeLists.txt:21 (include)
  ```
  Implying that the project should be configured with build type set to "Debug" in order to generate code coverage data.

Comment (marcmo) [reply .github/workflows/code-coverage.yml]: yes...please set the build type to `Debug`

File: .github/workflows/code-coverage.yml (hunk 1 above)
Comment (SuhashiniNaik) [reply .github/workflows/code-coverage.yml]: Yes Oliver, I have excluded the mock/test patterns but I also had to exclude 3rd party files because they were not getting excluded by using --no-external command in the latest commit. It works for excluding system libraries though

File: .github/workflows/code-coverage.yml (hunk 3 above)
Comment (SuhashiniNaik) [reply .github/workflows/code-coverage.yml]: Yes thank you for letting me know

File: .github/workflows/build.yml [hunk 4]
```diff
@@ -11,21 +11,43 @@ jobs:
...
+        uses: actions/cache@v3
+        id: cache
+        with:
+          path: cmake-build-posix
+          key: ${{ github.ref }}-cmake-build-posix
 
+      - name: Configure CMake
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: cmake -B cmake-build-posix -S executables/referenceApp -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}
+
```
marcmo: one question: where do we define the BUILD_TYPE?

SuhashiniNaik: On line number 7 in build.yml  we define the build BUILD_TYPE for the entire workflow

vladyslavmarkovaccenture:
Since we don't have debug symbols in the debug build now the code coverage info might be imprecise. This is fixed in #23 along with failing unit test. I suggest merging it first.

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,67 @@
+name: Code Coverage Report
+
+on: [workflow_call,push, pull_request]
```
marcmo: space after commas

marcmo (approved):
LGTM

PR #19, opened 2024-11-12
Add workflow overview

ThoFrank:
Visualization of our GH action strategy.


ThoFrank:
also added the trigger conditions for the merge queue

netomi:
We tried to capture the rules how GitHub references jobs to status checks here: https://otterdog.readthedocs.io/en/latest/reference/organization/repository/status-check/

ThoFrank:
I looked at that page already. Probably it's a good idea to just use unique names for all jobs.

hbragge-acn (approved):

PR #20, opened 2024-11-13
Add doxygen github action to generate documentation and coverage.

shamitha-shashidhara:
Generated comprehensive documentation using Doxygen comments. Generated documentation coverage numbers using Covrxygen. Generating DoxygenWarningLog.txt, which includes warnings about undocumented elements.

Files Added
doxygen-build.yml: Workflow file to build the Doxygen documentation.
doxygen-publish.yml: Workflow file to publish the Doxygen documentation.
Doxyfile: Configuration file for Doxygen.
DoxygenMainPage.md: Main entry page for Doxygen  documentation.
README.md: Updated to include instructions on how to generate and view the Doxygen documentation.


frankdarcyacn:
We need to be careful that `Doxyfile` in this PR does not conflict with the one currently in gerrit's referenceApp workspace main branch - that version has `GENERATE_XML=NO` whereas in this PR `GENERATE_XML=YES` (as needed to generate coverage).
Maybe its easiest to set  `GENERATE_XML=YES` in gerrit first, but that would mean that XML is always generated, even when we just want HTML, and it would take twice as long (roughly). 
@marcmo - do we know when gerrit will next be synced to github?
Alternatively, perhaps we could leave is at `GENERATE_XML=NO` and if we want to generate XML for coverage then override it on the command line like this...
`( cat Doxyfile ; echo "GENERATE_XML=YES") | doxygen`
Just a thought.

PR #23, opened 2024-11-19
Fix debug symbols on debug build

vladyslavmarkovaccenture:
This commit fixes unit-tests AND _code coverage_ info as it re-adds debug symbols to debug build (for _code coverage_) and causes _googletest's_ `EXPECT_DEATH` to fail as expected while running `CrcTests.CalculateCRCWithInvalidData`. Further changes in a separate PR might be required for `estd_assert` macro for unit test builds.


marcmo (approved):
LGTM

PR #27, opened 2024-11-22
Add cache for build artifacts

shamitha-shashidhara:
**Used hash key concept:** The cache key is based on the source files and toolchain. If there are changes in these, a new build is triggered; otherwise, the cache is used.


marcmo (approved):
LGTM

PR #8, opened 2024-10-27
C++20 support

rolandreichweinbmw:
As discussed in https://github.com/eclipse-openbsw/openbsw/issues/7 , fixes build issues with C++20

PR #28, opened 2024-11-25
Reapply CPP20 compatibility and enable check in CI

ThoFrank:
Changes from previous PR #8 were overwritten. This reapplies them and enables CI builds for cpp 14,17 and 20.


ThoFrank:
@mthiede-acn2 could you please take a look? The changes from #8 are all done in one commit, but I had to patch one more file.

.github/workflows/build.yml
```diff
@@ -8,6 +8,7 @@ jobs:
     strategy:
       matrix:
         platform: [posix, s32k148]
+        cpp-standard: [14, 17, 20]
```
marcmo, line 11: this will of course extend our build times. Would probably be enough to just run the compiler front end without generating code. We can use the compiler’s syntax-checking mode or run only the front-end to parse by using the `-fsyntax-only` flag.

ThoFrank: I don't think ``-fsyntax-only`` is supported by CMake. It would probably require a bigger hack. I tried it naively by adding ``-DCMAKE_C_FLAGS`` which resulted in CMake complaining about a broken compiler which doesn't produce binaries.

We could add ccache in the future. This could help bring down compilation times. Somebody already created a [ccache-action](https://github.com/hendrikmuhs/ccache-action) which combines ccache with github's cache.

marcmo:
Regarding the C++ 20 compatibility changes, some of them seem to be highly controversial. 
why is removing `volatile` valid in all those places?
does it really make sense to get rid of compound assignments?
I'd rather NOT add C++ 20 support as long as we don't have the need...and then only do it after careful consideration.

marcmo (changes requested):
do not just remove volatile

ThoFrank:
I just reapplied the changes from #8. They were already approved and merged before the migration to GitHub.
So I didn't think it would be controversial. I combined all the changes from #8 into one commit. My changes are done in other commits

PR #38, opened 2024-12-19
Add workflow badges to README.md

SuhashiniNaik:
- Removed coverage.js that was previously written for converting` coverage.info` to `Json` 
- Updated README.md with links to workflow badges

Output of badges for reference:
image

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -60,9 +60,23 @@ jobs:
           genhtml cmake-build-unit-tests/coverage.info \
                   --output-directory cmake-build-unit-tests/coverage &&
 
-          # Zip the coverage report
           mv cmake-build-unit-tests/coverage code_coverage &&
-          zip -r code_coverage.zip code_coverage
+          zip -r code_coverage.zip code_coverage &&
+
+          #Badges
+          LinePercentage=\$(lcov --summary cmake-build-unit-tests/coverage.info | grep 'lines' | awk '{print \$2}') &&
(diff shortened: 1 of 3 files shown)


coverage.js
marcmo, line 1: Please put any additional files into the .github folder, not in the root

marcmo: I took a look at the parse stuff, I don't think we should use the `lcov-parse` javascript library unless we really need it. Another dependency and seems we can also get the data we want easier, e.g. like this
`lcov --summary coverage/lcov.info | grep lines | awk '{print $2}'`
isn't that all we need?

marcmo (changes requested):

coverage.js
SuhashiniNaik: Yes the command ` lcov --summary coverage/lcov.info | grep lines | awk '{print $2}' ` we can get the coverage percentage but to convert this into a workflow  badge and to be dynamic we will need to use [shields.io](https://shields.io/) a third party app to host our Coverage badge. If that is okay? Can I go ahead with it

Also I used the` lcov-parse`  javascript  library  to convert the` lcov.info `into a JSON, since the JSON will be used to make badge and it will be dynamic.

marcmo: let's try to use shields.io and see how that looks

SuhashiniNaik: Yes, I have used shields.io to generate the badge in `code-coverage.yml` and the link has been added to `README.md`

marcmo (approved):
LGTM

PR #39, opened 2024-12-19
Add GitHub action for IWYU

SuhashiniNaik:
This GitHub Action automates the analysis of C++ code to ensure proper header file usage using the [Include What You Use tool. IWYU identifies unnecessary includes and suggests the minimal set of headers required for a clean and efficient codebase.

-Output has been uploaded as an artifact


.github/workflows/iwyu.yml
```diff
@@ -0,0 +1,61 @@
...
+      run: echo "::add-matcher::./.github/problem-matcher.json"
+
+    - name: Export CXXFLAGS
+
+    #Since Clang 17 is installed in usr/bin, it will search for built-ins in /usr/lib/clang/17/include.
+    #The purpose of this step is to help the clang toolchain used by IWYU to locate the correct system
+    #header directories.
+
+      run: |
+        cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON executables/referenceApp
```
vladyslavmarkovaccenture: I suggest using -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE=iwyu

vladyslavmarkovaccenture (commented):
I wonder if we have a separate directory for the tools. Other than that LGTM.

.github/workflows/iwyu.yml
vladyslavmarkovaccenture: it looks like iwyu is not that stable yet as it produces compiler errors when built w/o setting CXXFLAGS to clang's system includes. And it's easier to be done using the iwyu_tool.py than CMake flag for it for now. We might change it to use CMake flag later when iwyu works without workarounds. Currently it only checks targets which can be built with clang only. Hence - no check for s32k1 target for now.

SuhashiniNaik: Hi @vladyslavmarkovaccenture 

[https://github.com/eclipse-openbsw/openbsw/actions/runs/13675127926](url)
This is the updated action, with warnings of IWYU listed as annotations

PR #40, opened 2024-12-20
Integration Tests CI

ThoFrank:
GH runners can't create a vcan interface. This makes it imposible to directly run the integration tests.

As a workaround this PR uses the NixOS testing library to spawn a VM (nested in the GitHub provided vm) in which we can enable socketcan and thus run the tests.

A MacOS based runner is required because the referenceApp freezes during on a vm nested on a linux host. @frankdarcyacn is investigating why, but this is really hard to debug

TODO: Clean up the git history

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -6,31 +6,22 @@ jobs:
   treefmt:
     runs-on: ubuntu-latest
     steps:
-      - name: Create user program directory
-        run: |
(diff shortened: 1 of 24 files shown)


ThoFrank:
This PR was created under the false assumption, that vcan isn't supported on the GitHub runners. #53 shows that it is possible. Thus this PR is no longer needed.

PR #52, opened 2025-01-20
Add BenchmarkSystem that implements benchmarking

marcmo:
originates from CR #124398


marcin-wierzbicki:
I fixed minor issue with printing percentages:
```
-        "Load test: CPU idle: %d.%d %% (load) (LOAD_COUNTER_LIMIT: 0x%x)\n",
+        "Load test: CPU idle: %d.%02d %% (load) (LOAD_COUNTER_LIMIT: 0x%x)\n",
```

executables/referenceApp/application/src/systems/BenchmarkSystem.cpp
```diff
@@ -0,0 +1,308 @@
...
+constexpr uint32_t SYSTEM_CYCLE_TIME  = 2000; /* milliseconds */
+constexpr uint32_t LOAD_COUNTER_LIMIT = 0x1800;
+} // namespace
+
+namespace systems
+{
+
+BenchmarkSystem::BenchmarkSystem(
+    ::async::ContextType const context,
+    ::lifecycle::ILifecycleManager& lifecycleManager,
```
hbragge-acn: Unused parameter

executables/referenceApp/application/include/systems/BenchmarkSystem.h
```diff
@@ -0,0 +1,156 @@
...
+
+#include <async/AsyncBinding.h>
+#include <async/FutureSupport.h>
+#include <console/AsyncCommandWrapper.h>
+#include <lifecycle/AsyncLifecycleComponent.h>
+#include <lifecycle/console/LifecycleControlCommand.h>
+#include <mcu/mcu.h>
+#include <runtime/StatisticsContainer.h>
+
+#define GET_HW_COUNTER() (DWT->CYCCNT)
```
hbragge-acn: I guess this is defined as a macro just to clarify that "CYCCNT" is a HW counter? It would be more common to not have a macro, but maybe just a comment above the first usage telling what it is. This habit originates from Misra that discourages using function-like macros.

executables/referenceApp/application/CMakeLists.txt
```diff
@@ -26,6 +26,10 @@ add_executable(
     ${app.referenceAppExtraSources}
     src/main.cpp)
 
+if (BUILD_BENCHMARK)
+    target_sources(app.referenceApp PRIVATE src/systems/BenchmarkSystem.cpp)
```
hbragge-acn, line 33: Should we maybe add a check here that prevents the benchmarking build on Posix? This might be slightly out of place, but currently it will fail due to a compiler error which is also not good. So you could just add:
`if (BUILD_TARGET_PLATFORM STREQUAL "POSIX")
    message(FATAL_ERROR "Benchmarking not supported on POSIX")
endif ()`

executables/referenceApp/application/include/app/BenchmarkLogger.h
```diff
@@ -0,0 +1,10 @@
+// Copyright 2024 Accenture.
+
+#ifndef GUARD_A2388DF9_4B35_42D5_8B75_DFC18C21C372
```
hbragge-acn: We should use "#pragma once" now

executables/referenceApp/application/src/app/app.cpp
```diff
@@ -70,10 +73,16 @@ LifecycleManager lifecycleManager{
...
 ::estd::typed_mem<::systems::RuntimeSystem> runtimeSystem;
+#endif
 ::estd::typed_mem<::systems::SysAdminSystem> sysAdminSystem;
-::estd::typed_mem<::systems::DemoSystem> demoSystem;
 ::estd::typed_mem<::systems::SafetySystem> safetySystem;
+#ifndef BENCHMARK
+::estd::typed_mem<::systems::DemoSystem> demoSystem;
+#else
+::estd::typed_mem<::systems::BenchmarkSystem> benchmarkSystem;
+#endif
```
hbragge-acn: Please do some regrouping to keep this clearer, eg:
#ifdef BENCHMARK
::estd::typed_mem<::systems::BenchmarkSystem> benchmarkSystem;
#else
::estd::typed_mem<::systems::RuntimeSystem> runtimeSystem;
::estd::typed_mem<::systems::DemoSystem> demoSystem;
#endif
::estd::typed_mem<::systems::SysAdminSystem> sysAdminSystem;
::estd::typed_mem<::systems::SafetySystem> safetySystem;

libs/bsw/runtime/include/runtime/RuntimeMonitor.h
```diff
@@ -86,6 +86,13 @@ class RuntimeMonitor
 
     void enterTask(size_t const taskIdx)
     {
+#ifdef BENCHMARK
```
hbragge-acn: Ifdefs make the code less testable so I think we should use them sparingly in the common code under libs/bsw. Since this file isn't even 200 lines long, maybe you could just create a copy like RuntimeMonitorSimple.h instead of modifying the original?

doc/learning/benchmark/index.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Benchmark mode
+==============
+
+Previous: :ref:`learning_hwio`
+
+The reference app can be built to run benchmark tests. There is a BenchmarkSystem class that
+implements the following tests:
+
+* interrupt latency test - to measure the time between raising the interrupt line and entering the ISR,
```
hbragge-acn: Bullet point lists don't need commas at the end of entries

hbragge-acn (approved):
Looks good

marcin-wierzbicki:
fixed code formatting

PR #53, opened 2025-01-23
Add github action to run pytest on posix build

frankdarcyacn:
This adds a github action to run integration tests on the posix build.
The `ubuntu-latest` runner was built with SocketCAN support but is missing the kernel module file for `vcan`.
Running `sudo apt install linux-modules-extra-$(uname -r)` adds this kernel module file, then `vcan0` can be set up and the tests work.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build S32k and posix platform
 
-on: [push, pull_request]
+on: [workflow_call, push, pull_request]
 
 jobs:
   run-command:
--- a/.github/workflows/run-pytest-on-posix.yml
+++ b/.github/workflows/run-pytest-on-posix.yml
@@ -1,0 +1,56 @@
+name: Run pytest for posix platform
+
+on: [push, pull_request]
+
+jobs:
+    run-pytest-on-posix:
+      runs-on: ubuntu-latest
+
+      steps:
+
+        - name: Install missing kernel module file for vcan
+          run: sudo apt install linux-modules-extra-$(uname -r)
+
+        - name: Install can kernel modules
+          run: |
+            sudo modprobe can
+            sudo modprobe can-raw
+            sudo modprobe vcan
+
+        - name: Set up vcan0 interface
+          run: |
+            sudo ip link add dev vcan0 type vcan
+            sudo ip link set vcan0 mtu 16
+            sudo ip link set up vcan0
+
+        - name: Show all interface info, vcan0 should be set up
+          run: ip a
+
+        - name: Checkout repository
+          uses: actions/checkout@v4
+
+        - name: Set up CMake
+          uses: jwlawson/actions-setup-cmake@v2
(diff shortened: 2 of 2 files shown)


marcmo (approved):
LGTM

shamitha-shashidhara:
Hi @frankdarcyacn 
Instead of running the Build POSIX target in the pytest workflow, we can utilize the cache from the build.yml workflow right?

frankdarcyacn:
> Hi @frankdarcyacn Instead of running the Build POSIX target in the pytest workflow, we can utilize the cache from the build.yml workflow right?

Hi @shamitha-shashidhara, I see your point, yes, it should be possible to use a cached build output from another action. 
But I took a look at `build.yml`, am I correct in thinking that it builds 6 different ways as defined by this...
```
      matrix:
        platform: [posix, s32k148]
        cpp-standard: [14, 17, 20]
```
and that it will only rebuild if particular files have changed, as defined by `hashFiles` below?
```
          key: ${{ runner.os }}-cmake-${{ matrix.platform }}-${{ matrix.cpp-standard }}-${{ hashFiles('**/*.cpp', '**/*.h',  '**/*.cmake', '**/*.txt', '**/*.c', '**/*.s', 'admin/cmake/ArmNoneEabi.cmake') }}
```
eg. In the [this commit](https://github.com/eclipse-openbsw/openbsw/commit/fda81b1d4143f883c2c7e4ca38082579ba8a7f0f) it looks like the builds were not rebuilt in the [action run](https://github.com/eclipse-openbsw/openbsw/actions/runs/12889036302)? 
If so, then I think it would be safer if  `build.yml` always builds for all targets, in case there are other file types, now or in the future, that affect the build?
Looking at  `build.yml` makes me ask myself, should the python tests be run against the posix build with a particular cpp-standard? I presume the 3 different standards are specified just to check there are no compilation errors or warnings with any of them. 
Hi @marcmo, would you have opinions on the questions above?

marcmo:
the 3 different standards are for making sure we build with those compilers.
we did not find a way of just running the compiler frontend without producing the object files...that would have been preferred.
The caching has to work...if we need to modify it, we should do so....
@shamitha-shashidhara @SuhashiniNaik would you please verify this? 
if possible we should use the caching also for the tests of course!

shamitha-shashidhara:
The cache will restore builds only if there are changes in`.cpp`,` .h`,`arm`, or `CMake` files that affect posix and s32k builds. If there are changes that do not affect the build, such as documentation changes or .yml changes, the cache will reuse the output of the already existing cache, which saves time. The [commit](https://github.com/eclipse-openbsw/openbsw/commit/fda81b1d4143f883c2c7e4ca38082579ba8a7f0f) mentioned by @frankdarcyacn has only `.yml` changes and `README` updates, so the posix and s32k builds were not triggered.

frankdarcyacn:
I've updated to use the cached posix build when running pytest.

PR #57, opened 2025-02-05
Add safety modules

marcmo:
originates from CR #125941


libs/safety/safeMemory/include/safeMemory/mpu.h
```diff
@@ -0,0 +1,267 @@
...
+        IP_MPU->CESR = MPU_CESR_VLD(1);
+#endif
+    }
+
+    static bool isGlobalEnabled()
+    {
+#ifdef PLATFORM_SUPPORT_MPU
+        return ((IP_MPU->CESR & (MPU_CESR_NRGD_MASK | MPU_CESR_VLD_MASK)) == 0x01U);
+#endif
+        return false;
```
rolandreichweinbmw, line 155: Maybe guard via #else #endif since this second return might become unreachable?

libs/safety/safeMemory/include/safeMemory/mpu.h
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    static void setDescriptor(uint8_t const slot, tDescriptor const& d)
```
rolandreichweinbmw: If the sole purpose of uint8_t instead of size_t is saving space: Have you measured the difference? Using uint8_t typically compiles into even bigger code than uint32_t or size_t on 32 bit ARM due to additional conversion for non accu size. On the stack, probably takes same space due to alignment?

libs/safety/safeMemory/src/safeMemory/MemoryProtection.cpp
```diff
@@ -0,0 +1,197 @@
...
+{
+    // clang-format off
+    Mpu::globalDisable();
+    Mpu::setWordInDescriptor<0U, RegionDescriptor_AccessControl,
+        safety::mpuWord2<Mpu::R, Mpu::W, Mpu::R, Mpu::W, Mpu::R, Mpu::W, Mpu::R, Mpu::W,
+            Mpu::SM_RWX, Mpu::UM_RWX, Mpu::SM_RWX, Mpu::UM_RWX,
+            Mpu::nPI, Mpu::SM_UserMode, Mpu::UM_RWX,
+            Mpu::nPI, Mpu::SM_UserMode, Mpu::UM_nXnWnR>::word>();
+#ifdef PLATFORM_SUPPORT_MPU
+    for (uint8_t i = 0U; i < ((sizeof(memoryProtectionConfigurationRam)) / (sizeof(Mpu::tDescriptor))); ++i)
```
rolandreichweinbmw: See above, maybe size_t instead of uint8_t? Similar for the other loop variables below.

libs/safety/safeMemory/include/safeMemory/mpu.h
alinemundt: done

libs/safety/safeMemory/include/safeMemory/mpu.h
alinemundt: done

libs/safety/safeMemory/src/safeMemory/MemoryProtection.cpp
alinemundt: done

alinemundt (approved):

PR #58, opened 2025-02-07
Switch from ESTD to ETL

rolandreichweinbmw:
As being discussed in https://github.com/eclipse-openbsw/openbsw/discussions/51 , this is a first request for review via pull request.


marcmo:
thanks for your PR. It's impressive...it's massive! there is a lot of code that still needs formatting and it does not yet compile—those 2 things we need to solve first.

rolandreichweinbmw:
Regarding code formatting: The CI check "treefmt" suggests using treefmt, but the documented use of "treefmt" (binary from github repo) gives me:

$ treefmt
2025/02/08 20:03:46 INFO using config file: /home/rr/openbsw/.treefmt.toml
Error: failed to create composite formatter: failed to initialise formatter cmake: formatter command not found in PATH
$

Building treefmt from sources also results in errors, both with "go build", "nix-build" and "nix build".

What would you recommended now?

rolandreichweinbmw:
Regarding compile: For me, it compiles and runs (on C++14).

Do you mean the C++20 check in the CI? For now, I intentionally configured ETL for C++14 which enables definition of e.g. char8_t which is built-in in C++20.

Or do you mean a different compile error I'm not aware of?

I compiled:
- Unit tests
- POSIX
- S32K148

rolandreichweinbmw:
BTW: I'm using my fork/branch of ETL at https://github.com/rolandreichweinbmw/etl/tree/dev-rr

It includes some additions that I already contributed upstream, but no ETL release happeded yet.

executables/referenceApp/application/src/systems/DoCanSystem.cpp
```diff
@@ -70,25 +72,27 @@ void DoCanSystem::initLayer()
...
+        ::etl::ref(doCanTransceiver),
+        ::etl::ref(_tickGenerator),
+        ::etl::ref(_transportLayerConfig),
         ::util::logger::DOCAN);
 }
 
 void DoCanSystem::init()
 {
-    _classicAddressingFilter.init(::estd::make_slice(_addresses), ::estd::make_slice(_codecs));
+    _classicAddressingFilter.init(
```
christian-schilling: Are there no comparable `make_span` helpers in `etl` that offer type deduction? I think it's quite beneficial to have/use those.

executables/unitTest/common/include/util/EcuIdList.h
```diff
@@ -52,7 +53,9 @@ class EcuIdList
 
     uint16_t getSize() const
     {
-        return (fBufferLength == 0U) ? 0U : ::estd::read_be<uint16_t>(fpData);
+        return (fBufferLength == 0U)
```
christian-schilling: This is quite a step back in readability. I wonder if this can be expressed more concisely with `etl`.
One of the things we tried to achieve with estl was to eleminate the need to use `reinterpret_cast` from most code as far as possible by providing safer alternatives. One reasoning for this is also that `reinterpret_cast` often triggers warnings with safty related rulesets and using a safe wrapper api seems like a better approach over always adding suppression comments for those warnings.

libs/bsw/asyncFreeRtos/include/async/TaskInitializer.h
```diff
@@ -248,8 +258,9 @@ void TaskInitializer<Adapter>::create(
     TaskFunctionType const taskFunction,
     TaskConfigType const& config)
 {
-    ::estd::slice<uint8_t> bytes = ::estd::make_slice(stack).template reinterpret_as<uint8_t>();
-    ::estd::memory::align(alignof(StackType_t), bytes);
+    ::etl::span<uint8_t> bytes = ::etl::span<uint8_t>(
```
christian-schilling: Another case of `reinterpret_cast` being reintroduced in place of an api specifially designed to avoid it. (`reinterpret_as`, which is safer as it also includes the length adjustment calculations)

libs/bsw/docan/include/docan/can/DoCanPhysicalCanTransceiver.h
```diff
@@ -137,12 +139,12 @@ SendResult DoCanPhysicalCanTransceiver<Addressing>::startSendDataFrames(
...
     FrameIndexType const lastFrameIndex,
     FrameSizeType const consecutiveFrameDataSize,
-    ::estd::slice<uint8_t const> const& data)
+    ::etl::span<uint8_t const> const& data)
 {
     (void)lastFrameIndex;
     if (!_sendPending)
     {
-        ::estd::slice<uint8_t> payload = ::estd::slice<uint8_t>::from_pointer(
+        ::etl::span<uint8_t> payload(
```
christian-schilling, line 146: We chose to use `from_pointer` intead of overloading the constructor to make this less-safe and therefore less preferable way of creating slices/spans easier to spot.

libs/bsw/docan/test/gtest/src/docan/addressing/DoCanNormalAddressingTest.cpp
```diff
@@ -36,8 +36,9 @@ TEST(DoCanNormalAddressingTest, testEncodeSingleFrameWithAddress)
     EXPECT_EQ(0x1238934U, canId);
     // expect valid encoded single frame
     uint8_t const expectedPayload[] = {0x33};
-    ::estd::slice<uint8_t> payload  = payloadBuffer;
-    EXPECT_TRUE(::estd::memory::is_equal(expectedPayload, payload));
+    ::etl::span<uint8_t> payload    = payloadBuffer;
+    EXPECT_TRUE(
```
christian-schilling: I think here the `estl` api was way better. It also has different behaviour, checking the equality of the sizes. However it does look like etl has `etl::eqal`, why not use that here?

libs/bsw/io/examples/MemoryQueueExample.cpp
```diff
@@ -145,9 +149,10 @@ bool receiveCanFrame(CanFrame& frame, ::io::IReader& reader)
         return false;
     }
     // Copy data to frame. We expect a big endian 32bit id followed by the actual data.
-    frame.id = ::estd::memory::take<::estd::be_uint32_t>(data);
-    ::estd::memory::copy(frame.data, data);
-    frame.data.trim(data.size());
+    frame.id = *reinterpret_cast<::etl::be_uint32_t*>(data.data());
+    data     = data.subspan(sizeof(etl::be_uint32_t));
+    ::etl::mem_copy(data.begin(), data.end(), frame.data.begin());
+    frame.data = frame.data.subspan(0, data.size());
```
christian-schilling: `.trim()` as an inplace operation was introduced because some compilers used significantly more stack space when using `.subslice()`. This may or may not be an issue with the compilers being used in the future, but it certainly was quite significant in the past.

libs/bsw/io/include/io/ForwardingReader.h
```diff
@@ -84,7 +85,7 @@ inline ::estd::slice<uint8_t> ForwardingReader::peek() const
 inline void ForwardingReader::release()
 {
     // If _destinationData is an empty slice, copy will not copy any data.
-    (void)::estd::memory::copy(_destinationData, _sourceData);
+    (void)::etl::mem_copy(_sourceData.begin(), _sourceData.end(), _destinationData.begin());
```
christian-schilling: `estd::memory::copy` is safer here as it also checks the destination length, which `etl::mem_copy` obviously can't.

libs/bsw/uds/mock/gmock/src/uds/connection/ResponseMock.cpp
```diff
@@ -23,20 +23,20 @@ bool PositiveResponse::appendUint8(uint8_t const data) { return appendData(&data
...
+    auto const v = ::etl::be_uint32_t(data);
+    return appendData(static_cast<uint8_t const*>(v.data() + 1), 3) == 3;
 }
 
 bool PositiveResponse::appendUint32(uint32_t const data)
 {
-    auto const v = ::estd::be_uint32_t::make(data);
-    return appendData(v.bytes, 4) == 4;
+    auto const v = ::etl::be_uint32_t(data);
+    return appendData(static_cast<uint8_t const*>(v.data()), 4) == 4;
```
christian-schilling: I think the `estl` api is more clear here.

libs/bsw/io/include/io/ForwardingReader.h
filipe-cuim: if we want safe copies, etl also offers that in copy_s in etl/algorithm.h
Also etl::mem_copy shouldn't check the length since it is a low level copy (basically memcpy but with pointers of the same time), so in this case it makes sense.

libs/bsw/io/examples/MemoryQueueExample.cpp
filipe-cuim: I have no knowledge about this issue, but subspan seems more clear than trim. I know estd also has subslice, but they seem to be doing the same thing (both etl::span::subspan and estd::slice::subslice).
Also the fact that trim may reset the size of the slice to zero seems to be strange in my view (even though it is documented).
But really not sure about this point.

libs/bsw/uds/mock/gmock/src/uds/connection/ResponseMock.cpp
filipe-cuim: I think the static_cast wasn't needed here. The method etl::unaligned_type_common::data() already returns a const uint8_t*.
Honestly since data is used on most containers to get a pointer to its raw buffer, I think data seems more aligned with the rest of the apis.
About the make method, seems to be a bit redundant.

libs/bsw/docan/test/gtest/src/docan/addressing/DoCanNormalAddressingTest.cpp
filipe-cuim: agreed. We could have used etl::equal here.

rolandreichweinbmw:
> I'm raising a few points where I see `etl` being a step back from `estl` to put them up for discussion. Apart from those we should think about how to treat etl and it's development. Do we want to upstream changes? Maintain patches? Keep it as is? Maintain an "add-on" library for missing pieces (like the ones I commented on).
> 
> Also I'm wondering if we really gain a lot by adopting etl. Just because we can does not mean we should.

During the process for this PR, I already contributed some additions to ETL to easen its use in OpenBSW, and the author is quite responsive:

https://github.com/ETLCPP/etl/issues?q=is%3Apr%20author%3Arolandreichweinbmw

Most of it he already merged, but unfortunately, they are not yet in the master branch or released yet due to the ETL development process.

However, I'm maintaining the changes relevant to OpenBSW in:

https://github.com/rolandreichweinbmw/etl/tree/dev-rr - see also the individual feature branches.

I think for this purpose, we can maintain a fork or branches for changes not yet released in upstream ETL even though we should contribute all of it.

One point for ETL is that some of the features I would like to contribute are already based on ETL. And ETL is being used in projects where I would like to merge OpenBSW. Apart from ETL being established in the industry and public since many years.

libs/bsw/docan/include/docan/can/DoCanPhysicalCanTransceiver.h
filipe-cuim: Makes sense. Could be something to suggest also for ETL

libs/bsw/asyncFreeRtos/include/async/TaskInitializer.h
filipe-cuim: Agreed! But I wonder if this could be done in another way. Or if span could be construct in a different manner, because etl also offers reinterpret_as in etl/span.h. Haven't had the time to try that out myself.

executables/referenceApp/application/src/systems/DoCanSystem.cpp
rolandreichweinbmw: I agree that they are really helpful.

There are deduction guides to the etl::span ctors:

https://github.com/ETLCPP/etl/blob/99d75375060b61aae49bb2fc06349b45cb2c3380/include/etl/span.h#L750

Unfortunately, only available w/ ETL_USING_CPP17

Maybe we can consider switching to C++17 in OpenBSW soon? :-)

For C++14, I can provide etl::make_span helper as we know from estd.

Update:
See: https://github.com/rolandreichweinbmw/etl/tree/span-make-span
And: https://github.com/ETLCPP/etl/pull/1027

Now using it accordingly here.

filipe-cuim: Agreed not sure why it was not added to etl, because they have make_array for example. Could be a suggestion for ETL

executables/unitTest/common/include/util/EcuIdList.h
filipe-cuim: Agreed. I think the goal of ETL is also to achieve that "less usage of reinterpret_cast". Probably there is some other workaround for this specific use-case. Didn't try that out myself though.

rolandreichweinbmw: I agree!

That's why I contributed two alternative solutions to ETL some time ago:

* unaligned_type_ext: https://github.com/rolandreichweinbmw/etl/tree/unaligned-type-ext
    separate type; read/write e.g. via etl::be_uint16_t_ext{data}

* unaligned_type::at_address: https://github.com/rolandreichweinbmw/etl/tree/unaligned-type-at-address
    pseudo constructor helper; e.g. etl::be_uint32_t::at_address(data)

I haven't used it yet in the above OpenBSW line since upstream has not decided yet which one to merge.

libs/bsw/asyncFreeRtos/include/async/TaskInitializer.h
rolandreichweinbmw: I agree. I already contributed span::reinterpret_as() some time ago at:

https://github.com/rolandreichweinbmw/etl/tree/span-advance-copy-reinterpret-as

(Not yet available in upstream release, but we can already use it if you like as it is available in the ETL copy in this PR.)

libs/bsw/docan/test/gtest/src/docan/addressing/DoCanNormalAddressingTest.cpp
rolandreichweinbmw: Agreed. Adjusted accordingly.

libs/bsw/io/include/io/ForwardingReader.h
rolandreichweinbmw: Right.

When I previously contributed this addition to ETL:

https://github.com/ETLCPP/etl/pull/1024
https://github.com/rolandreichweinbmw/etl/tree/span-advance-copy-reinterpret-as

... then I better also use it. :-)

Adjusting accordingly.

libs/bsw/uds/mock/gmock/src/uds/connection/ResponseMock.cpp
rolandreichweinbmw: Agree. I removed static_cast.

rolandreichweinbmw:
Just updated the branch and PR with new ETL upstream update, based on new ETL version 20.40.0, but with yet unmerged changes upstream that we need for easy switchover of OpenBSW to ETL. (As before, copied into the openbsw branch "use-etl".)

New ETL branch with those additions for reference is: https://github.com/rolandreichweinbmw/etl/tree/openbsw-replace-estd

rolandreichweinbmw:
Can you please do a new review?

marcmo:
@rolandreichweinbmw could you please mark the PR as draft?

rolandreichweinbmw:
> @rolandreichweinbmw could you please mark the PR as draft?

I'm missing the permissions to do so. But as a committer, you can add this I guess?

rolandreichweinbmw:
Just pushed an update, rebased to current main, but no significant changes. Please review.

libs/bsw/io/include/io/VariantQueue.h
```diff
@@ -118,39 +168,41 @@ struct variant_q
...
+            ::etl::type_list_contains<TypeList, T>::value,
+            "type must be a part of the variant type list");
 
-        ::estd::memory::take<uint8_t>(buffer)
-            = static_cast<uint8_t>(::estd::index_of<TypeList, T>::value);
-        ::estd::memory::take<T>(buffer) = t;
+        buffer[0] = static_cast<uint8_t>(::etl::type_list_index_of<TypeList, T>::value);
+        buffer.advance(1);
+        buffer.reinterpret_as<T>()[0] = t;
+        buffer.advance(sizeof(T));
```
hbragge-acn, line 188: Sorry if this has been discussed before, but is there really no equivalent for estd::memory::take in etl? I know it's almost unused in openbsw for now, but I noticed that it's very popular in some of our project specific code. Maybe there are alternative ways, but it seems to be very helpful when accessing many different, sequential things in a buffer. Maybe we should consider adding it in etl. Other than that, the patch looks really impressive so far (I still have a couple more modules to check).

rolandreichweinbmw: Yes, I'm aware that take<>() is a useful convenience function. And technically, it would be easy to add it to ETL (as I did already with other ESTD interfaces). However, I found the following issue when thinking about adding it to ETL: take<>() removes the respective part from the span, and then returns a reference to it, letting the caller manipulate this memory which is strictly not part of any object (span) anymore.

rolandreichweinbmw: By reinterpret-casting in a byte stream, take<>() can also violate strict aliasing rules, as rejected by upstream ETL in a similar case: https://github.com/ETLCPP/etl/pull/999

hbragge-acn: Interesting insights, thanks. Seems to have been working well enough for us. But since there aren't many usages yet, I think it's okay to do it manually for now like you did.

christian-schilling: _If_ OpenBSW moves to ETL, we could/should extract parts like this to it's own library. There is no reason to add everything that is present in estd but missing etl to the etl.

libs/bsw/uds/src/uds/connection/PositiveResponse.cpp
```diff
@@ -29,33 +29,36 @@ bool PositiveResponse::appendUint8(uint8_t const data) { return appendData(&data
...
+    auto const sourceData = ::etl::span<uint8_t const>(data, length);
     if (fBuffer.size() < sourceData.size())
     {
         fIsOverflow = true;
         return 0;
     }
-    auto const result = ::estd::memory::copy(fBuffer, sourceData);
-    fBuffer.advance(result.size());
-    return static_cast<uint16_t>(result.size());
+    if (::etl::copy(sourceData, fBuffer))
```
hbragge-acn: The copy cannot fail because buffer sizes have already been checked above, making the "return 0" unreachable, right? I would just remove the if.

rolandreichweinbmw: Correct. Adjusting accordingly. Thanks for the note!

libs/bsw/io/include/io/VariantQueue.h
rolandreichweinbmw: I actually separated out the part of estd that stays to util/estd. Since ETL is a header only library, and OpenBSW relies on certain compile units for:
* HW timer access in for chrono.h
* assert()

those are in util/estd for now.

However, to prevent confusion about multiple libs, we should try to keep as much as possible in ETL and "upstream" it.

hbragge-acn (approved):
I've gone through the whole thing and didn't see anything else worth complaining about. Looks like a step forward, all things considered. Also liked the voluntary cleanup you did here and there. Kudos for the massive effort

rolandreichweinbmw:
Can you please have a look at the failing check "[eclipsefdn/eca](https://api.eclipse.org/git/eca/status/gh/eclipse-openbsw/openbsw/58) — The author(s) of the pull request is not covered by necessary legal agreements" because it worked before and I haven't changed anything in my account.

rolandreichweinbmw:
Rebased and adjusted documentation.

rolandreichweinbmw:
Updated to new ETL 20.41.1. The only deviation left is the etl::inplace_function Function Wrapper (not yet available in released ETL).

Also, libs/3rdparty now includes the whole copy of ETL 20.41.1, not just the headers in etl/include/etl/*

jiangyilism (approved):

rolandreichweinbmw:
Using etl::delegate and etl::closure in async Function/Call now.

PR #59, opened 2025-02-11
Fix compilation issues with clang on Linux

marcmo:
originates from CR #126489


platforms/posix/bsp/bspInterruptsImpl/src/interrupts/suspendResumeAllInterrupts.cpp
```diff
@@ -3,14 +3,17 @@
 #include "interrupts/suspendResumeAllInterrupts.h"
 
 #include <FreeRTOS.h>
+#include <pthread.h>
```
rolandreichweinbmw: How about using STL's &lt;thread&gt; instead?

ThoFrank: Thanks! Good idea

PR #62, opened 2025-02-21
Fix CI script regarding failed unit tests

rolandreichweinbmw:
Besides the CI script now failing on failed unit tests, the -j from ctest was removed also, because otherwise the test summary was truncated from the output, and the tests are run sequentially anyway.


marcmo (approved):
good catch

PR #63, opened 2025-02-24
SafeSupervisor: Fix format string to print int

rolandreichweinbmw:


libs/safety/safeLifecycle/src/safeLifecycle/SafeSupervisor.cpp
```diff
@@ -101,8 +101,8 @@ void SafeSupervisor::handle(Event const& event)
         {
             Logger::warn(
                 SAFETY,
-                "SafeSupervisor: Received unknown event with ID %s",
-                ::estd::to_underlying(event));
+                "SafeSupervisor: Received unknown event with ID %d",
+                static_cast<int>(::estd::to_underlying(event)));
```
mthiede-acn2, line 105: the change to %d is correct, but is the static_cast necessary?

rolandreichweinbmw: Good question. Depending on how the format string handling is defined, it might be best to cast to the official type that "%d" is interpreting.

aschaal: %d is officially int, so for me this looks good.

We can change it in future if we want, but your fix definitely improves the code, so the commit can be approved.

aschaal (approved):

PR #68, opened 2025-02-28
make S32K148EVB buildable with clang

jiangyilism:


platforms/s32k1xx/hardFaultHandler/src/hardFaultHandler.s
```diff
@@ -64,22 +64,22 @@
 customHardFaultHandler:
 
     // Save flags for dump in advance
-    mrs     R12, PSR
+    mrs     R12, APSR
```
jiangyilism: Please help review this change.

jiangyilism: update: Changed to XPSR. Since the gnu assembler undocumented behavior make PSR an alias to XPSR.
ref: https://reviews.llvm.org/D39712?id=121835
This undocumented behavior still exists in gas-2.44.  `mrs R12, PSR` and `mrs R12, XPSR` are assembled to the same machine code.

Filelists.cmake
```diff
@@ -33,6 +33,8 @@ add_compile_options(
 add_compile_options(
     "$<$<COMPILE_LANG_AND_ID:C,Clang,GNU>:-O2;-g3;-Werror;-Wall>")
 
+add_link_options(-Wl,--noinhibit-exec)
```
jiangyilism, line 40: This command line option is from https://reviews.llvm.org/D41046?id=132044 .
Current linker script makes sections .data_rom and .data overlapping. gnu ld accepts it without warning but other linkers consider it a suppressible error.

vladyslavmarkovaccenture:
Important: If it now compiles s32k1 with clang, then we need to update the verifier in the same PR accordingly.

jiangyilism:
s32k1xx.verifier.toml specify the toolchain file admin/cmake/ArmNoneEabi.cmake , which builds with arm-none-eabi-g++ . This PR do not intend to change the default arm cmake toolchain file. It only fixes build errors for clang.

openbsw users can build with their own toolchain file with other compilers than gcc or clang.

@vladyslavmarkovaccenture  Are you suggesting to add a admin/cmake/ArmNoneEabi-clang.cmake toolchain file and a s32k1xx-clang.verifier.toml ?    Maybe even modify .github/workflows/build.yml to add a s32k148-clang build ?
If this is the case, we will need to discuss the default llvm toolchain suggested by openbsw. Currently I build with https://github.com/arm/arm-toolchain but other llvm-based arm toolchain might prefer different cmake toolchain file setting.

vladyslavmarkovaccenture:
@jiangyilism 

> openbsw users can build with their own toolchain file with other compilers than gcc or clang.

We don't use any toolchain files to build with clang for example. But we didn't build s32k1 with clang so far.
Testing if unit tests are building and if posix app is building can be done with setting CXX / CC variables while running cmake like so:
`CXX=clang++17 CC=clang-17 cmake -S executables/referenceApp -B build`
The main problem is that we still have no verifiers for this for some reason.
> @vladyslavmarkovaccenture Are you suggesting to add a admin/cmake/ArmNoneEabi-clang.cmake toolchain file and a s32k1xx-clang.verifier.toml ? Maybe even modify .github/workflows/build.yml to add a s32k148-clang build ? If this is the case, we will need to discuss the default llvm toolchain suggested by openbsw. Currently I build with https://github.com/arm/arm-toolchain but other llvm-based arm toolchain might prefer different cmake toolchain file setting.

Yes, modifying or adding verifier.toml files **and** GitHub workflows is necessary; otherwise, compiler errors will be reintroduced. I will speak with the guys responsible for verifiers on our side. I need to know if we are using *verifier.toml files on jenkins still. We should add verifiers for Clang in **both** Jenkins and GitHub workflows.

jiangyilism:
@vladyslavmarkovaccenture  Toolchain file `admin/cmake/ArmNoneEabi-clang.cmake` is now added. Also refactor `.github/workflows/build.yml` to include s32k148-clang build.  

Notes: 
1. Some github workflow jobs are renamed
2. `admin/cmake/ArmNoneEabi.cmake` is renamed to `admin/cmake/ArmNoneEabi-gcc.cmake`

How to test/verify with `s32k1xx.verifier.toml` ?

platforms/s32k1xx/bsp/bspMcu/include/3rdparty/cmsis/cmsis_compiler.h
```diff
@@ -39,14 +39,7 @@
...
 /*
- * LLVM/Clang Compiler
- */
-#elif defined ( __clang__ )
-  #include "cmsis_clang.h"
-
-
-/*
- * GNU Compiler
+ * GNU and LLVM/Clang Compiler
```
matthiaskessler: When we added cmsis, we only added the files needed to support gcc. Since we try to not modify 3rdparty code, IMHO, we should try to add the respective file from the original cmsis package. The details about the version used can be found in NOTICE.md.

jiangyilism: Added `3rdparty/cmsis/cmsis_clang.h` from CMSIS v6.1.0
Since currently all my CI/workflows are blocked on this repo, I can only trigger runs on my fork https://github.com/jiangyilism/openbsw/actions/runs/13789390165

Is there a way to unblock github workflows of this PR?

jiangyilism:
Rebased on main branch.
pytest run: https://github.com/jiangyilism/openbsw/actions/runs/13807706258

jiangyilism:
> Rebased on main branch. pytest run: https://github.com/jiangyilism/openbsw/actions/runs/13807706258

@frankdarcyacn Please help review my last commit to enable pytest runs on both gcc and clang posix builds.

.github/workflows/run-pytest-on-posix.yml
```diff
@@ -60,4 +61,5 @@ jobs:
           if: ${{ steps.restore-cached-posix-build.outputs.cache-hit == 'true' }}
           run: |
             cd test/pyTest
+            sed --in-place 's/cmake-build-posix/&-${{ matrix.compiler }}/' target_posix.toml
```
frankdarcyacn: I'd rather not use `sed` to locally change a line in `target_posix.toml`. It looks like the kind of thing that gets missed if things change in the future. In my experience, if I use a regular expression to solve something I ultimately find I've introduced about 5 more problems. Alternatively you could create a symbolic link `cmake-build-posix/` pointing to the new directory name `cmake-build-posix-gcc/` or `cmake-build-posix-clang/`. But I'm not convinced these tests need to be run twice.

frankdarcyacn:
> > Rebased on main branch. pytest run: https://github.com/jiangyilism/openbsw/actions/runs/13807706258
> 
> @frankdarcyacn Please help review my last commit to enable pytest runs on both gcc and clang posix builds.

Hi, I see you have the integration tests (those using `pytest`) working with both `gcc` and `clang` builds. But is that really needed? 
Is it sufficient to run those tests once? I mean, we don't run those tests for every build created using the different standards (14, 17, 20). Those tests are checking that the `referenceApp` does what it should. If the purpose is to check for differences in code behaviour with different compilers then would it make more sense to run the unit tests of each `openbsw` library built with both compilers? And perhaps compare performance differences in those unit tests?

Personally, I would just run `pytest` once, for the build it currently uses. I'd make the change you have to fetch the cached build using the new key name and just add a symbolic link `cmake-build-posix/` pointing to the new directory name `cmake-build-posix-gcc/` to make it work (instead of the `sed` line). Does that make sense?

jiangyilism:
@frankdarcyacn Can remove clang from strategy-matrix-compiler array for now, if current pytest does not test anything toolchain-dependent.
We will face the same question once we try to add arm/s32k148 pytest to github workflow.

I will remove clang from the array but keep the overall yaml structure. So if someone want to run it on private fork then the change will be minimal.

I will try the toml config file symlink approach too.

jiangyilism:
@frankdarcyacn
1. It uses symlink builddir approach now. Thanks for the advice.
2. If we want to reduce unnecessary build triggering, maybe consider skip building c++17 ? c++14 (minimum) and c++latest (20 currently) should be enough, unless c++latest will introduce big abi breakage?
3. The latest commit of this PR also fixes run-pytest-on-posix workflow. Without it, currently it always passes without actually running pytest. Due to cache/restore error: https://github.com/eclipse-openbsw/openbsw/actions/runs/13784603965/job/38549722772    . @marcmo Do you think we need a separate PR to fix run-pytest-on-posix?

frankdarcyacn:
> 3. The latest commit of this PR also fixes run-pytest-on-posix workflow. Without it, currently it always passes without actually running pytest. Due to cache/restore error: https://github.com/eclipse-openbsw/openbsw/actions/runs/13784603965/job/38549722772    . @marcmo Do you think we need a separate PR to fix run-pytest-on-posix?

That workflow run was marked as success overall despite showing...
> Error: Failed to restore cache entry. Exiting as fail-on-cache-miss is set. Input key: Linux-cmake-posix-20-0654f9e821872b3255d6aa4789bcaa3fd559530cba5b0c2cf3ba40311953a675
I see that it should have been looking for a key with `gcc` in it, and that's fixed now.
But I would have expected the workflow to fail because of `fail-on-cache-miss: true` - that didn't happen. To me that looks like a bug in cache action itself.

.github/workflows/run-pytest-on-posix.yml
```diff
@@ -37,10 +40,8 @@ jobs:
           id: restore-cached-posix-build
           uses: actions/cache/restore@v3
           with:
-            path: |
-              cmake-build-posix
-              cmake-build-s32k148
-            key: ${{ runner.os }}-cmake-posix-20-${{ hashFiles('**/*.cpp', '**/*.h',  '**/*.cmake', '**/*.txt', '**/*.c', '**/*.s', 'admin/cmake/ArmNoneEabi-gcc.cmake') }}
+            path: cmake-build-posix-${{ matrix.compiler }}
+            key: ${{ runner.os }}-cmake-posix-${{ matrix.compiler }}-20-${{ hashFiles('**/*.cpp', '**/*.h',  '**/*.cmake', '**/*.txt', '**/*.c', '**/*.s', 'admin/cmake/ArmNoneEabi-${{ matrix.compiler }}.cmake') }}
```
marcmo: @SuhashiniNaik @shamitha-shashidhara could you please take a look at this, there are cache problem.
see https://github.com/eclipse-openbsw/openbsw/pull/68#issuecomment-2721160142

shamitha-shashidhara:
> https://github.com/actions/cache/issues/1369 this says it is fixed in cache v4.0.2, although the issue is still "open"


@frankdarcyacn you are right that it was fixed in versions 4 and above. The cache version we are using in our build.yml and run-pytest-on-posix.yml is version 3 (`actions/cache@v3` and `actions/cache/restore@v3`). We need to change it to `actions/cache@v4` and `actions/cache/restore@v4`. When I tested on `v4`, `fail-on-cache-miss: true` seemed to work as expected.
https://github.com/esrlabs/openbsw/actions/runs/13893335012/job/38868925837

Let me know if I need to create a separate PR for this, or if it will be fixed in this PR itself.

.github/workflows/run-pytest-on-posix.yml
shamitha-shashidhara: Hi @marcmo, I have posted my investigation findings in this [comment](https://github.com/eclipse-openbsw/openbsw/pull/68#issuecomment-2728417265).
I have created a separate PR for cache issue - https://github.com/eclipse-openbsw/openbsw/pull/78

marcmo:
@jiangyilism please rebase

platforms/s32k1xx/hardFaultHandler/src/hardFaultHandler.s
marcmo, line 1: according to @kitsnet  line 3 (.syntax unified) of platforms/s32k1xx/hardFaultHandler/src/hardFaultHandler.s should be removed as well. This line tells about the GNU extensions of ARM assembly currently being used, and seemingly not supported by clang.

jiangyilism: gcc needs ".syntax unified" otherwise it defaults to ".syntax divided" and fails to build.
clang has ".syntax unified" by default and ignores ".syntax unified".
https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/ARM/AsmParser/ARMAsmParser.cpp#L11737

.github/workflows/build.yml
```diff
@@ -7,62 +7,63 @@ jobs:
...
       - name: Cache CMake files
         id: cache-cmake
         uses: actions/cache@v4
         with:
-          path: |
-            cmake-build-posix
-            cmake-build-s32k148
-          key: ${{ runner.os }}-cmake-${{ matrix.platform }}-${{ matrix.cpp-standard }}-${{ hashFiles('**/*.cpp', '**/*.h',  '**/*.cmake', '**/*.txt', '**/*.c', '**/*.s', 'admin/cmake/ArmNoneEabi.cmake') }}
+          path: cmake-build-${{ matrix.platform }}-${{ matrix.compiler }}
+          key: ${{ runner.os }}-cmake-${{ matrix.platform }}-${{ matrix.compiler }}-${{ matrix.cpp-standard }}-${{ hashFiles('**/*.cpp', '**/*.h',  '**/*.cmake', '**/*.txt', '**/*.c', '**/*.s') }}
```
jiangyilism, line 22: Note that 'admin/cmake/ArmNoneEabi.cmake' is removed since it is already included by  '**/*.cmake'

admin/cmake/ArmNoneEabi.cmake
```diff
@@ -1,37 +1,31 @@
-# Toolchain for gcc-arm-none-eabi-10.3-2021.10
 set(CMAKE_SYSTEM_NAME Generic)
```
jiangyilism, line 1: ArmNoneEabi.cmake now contains gcc and clang commom stuffs.  The real cmake toolchain files are now ArmNoneEabi-gcc.cmake and ArmNoneEabi-clang.cmake (the latters include ArmNoneEabi.cmake).
Please review this change.
P.S. Can also rename ArmNoneEabi.cmake to ArmNoneEabi-common.cmake if needed. To avoid others still depending on the filename ArmNoneEabi.cmake

marcmo (approved):
LGTM
@shamitha-shashidhara will check the build script changes

shamitha-shashidhara (approved):
Hi @marcmo, the build scripts LGTM. 
@jiangyilism Please format your changes.

PR #69, opened 2025-03-04
Support zephyr build

frankdarcyacn:
These changes are to support builds in https://github.com/esrlabs/openbsw-zephyr


frankdarcyacn:
Just to add a little more ifo on this PR, it contains 2 changes to support building samples in https://github.com/esrlabs/openbsw-zephyr

1. Add the option `INCLUDE_OPENBSW_LIBS_BSP` which is `ON` by default, so there's no change to the build behaviour in `OpenBSW` (as built for `FreeRTOS`), but in the `demo_app` sample in the `openbsw-zephyr` repo `INCLUDE_OPENBSW_LIBS_BSP` is set to `OFF` and `bspZephyr` is included instead.

2. Change the aliased library name `os::freeRtos` to `osFreeRtos`. Without this change the build of samples in the `openbsw-zephyr` repo fails with this error...
```
CMake Error at /home/frank/github/esr/openbsw_zephyr_workspace/openbsw/libs/bsw/asyncFreeRtos/CMakeLists.txt:5 (target_link_libraries):
  The link interface of target "asyncFreeRtos" contains:

    os::freeRtos

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```
This is despite the build having nothing to do with `asyncFreeRtos`. 
Without this change, there are two ways to fix the build...
a. Include the following in each sample's `CMakeLists.txt` ...
```
add_library(dummyfreeRtos INTERFACE)
add_library(os::freeRtos ALIAS dummyfreeRtos)
```
so its not ideal to have to do this in build that has nothing to do with freeRtos 
or...
b. add this to each sample's `CMakeLists.txt` ...
```
cmake_policy(SET CMP0028 OLD)
```
See https://cmake.org/cmake/help/latest/policy/CMP0028.html
So names containing `::` are treated differently, simplest solution is to remove them.

marcmo (approved):
LGTM

PR #77, opened 2025-03-14
Fix build failures for target estdExamples

frankdarcyacn:
Fix build failures for target estdExamples due to '#pragma clang' with gcc compiler
Only seen when building executables/unitTest

### DIFF
--- a/libs/bsw/estd/examples/deque.cpp
+++ b/libs/bsw/estd/examples/deque.cpp
@@ -8,7 +8,9 @@
 #include <gtest/gtest.h>
 
 #pragma GCC diagnostic ignored "-Wunknown-pragmas"
+#ifdef __clang__
 #pragma clang diagnostic ignored "-Wunused-private-field"
+#endif
 
 namespace
 {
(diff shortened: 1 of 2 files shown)


frankdarcyacn:
@marcmo the code-coverage.yml workflow is failing with latest code on main branch. This PR fixes it.

libs/bsw/estd/examples/deque.cpp
```diff
@@ -8,7 +8,9 @@
 #include <gtest/gtest.h>
 
 #pragma GCC diagnostic ignored "-Wunknown-pragmas"
+#ifdef __clang__
```
jiangyilism: code-coverage.yml fails while other build workflows succeeds since code-coverage.yml is building with gcc-11.4.0 

gcc failed to handle `-Wunknown-pragmas` and was fixed in gcc-13.3
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89038

If we wrap `#pragma clang` with `#ifdef __clang__` then we can also try to remove `#pragma GCC diagnostic ignored "-Wunknown-pragmas"` since the latter's sole purpose is to suppress `-Wunused-private-field`

frankdarcyacn: Done

PR #81, opened 2025-03-21
Add dependency on the linker script in cmake

marcmo:
originates from CR #127432

### DIFF
--- a/doc/learning/setup/index.rst
+++ b/doc/learning/setup/index.rst
@@ -28,6 +28,8 @@ If you have access to a
 `S32K148 development board <https://www.nxp.com/design/design-center/development-boards/automotive-development-platforms/s32k-mcu-platforms/s32k148-q176-evaluation-board-for-automotive-general-purpose:S32K148EVB>`_
 then below is some important information to note...
 
+.. _known_issues_with_s32k148_board:
+
 Known Issues with the S32K148 board
 +++++++++++++++++++++++++++++++++++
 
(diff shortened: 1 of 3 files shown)


executables/referenceApp/platforms/s32k148evb/main/CMakeLists.txt
```diff
@@ -42,3 +42,7 @@ set_target_properties(
     startUp
     PROPERTIES PROP_LINKER_SCRIPT
                "${CMAKE_CURRENT_SOURCE_DIR}/linkerscript/application.dld")
+set_target_properties(
+    startUp
```
jiangyilism, line 46: Should it be `app.referenceApp` instead?

doc/learning/setup/index.rst
```diff
@@ -53,6 +55,12 @@ Known Issues with the S32K148 board
         but please note that many people have encountered difficulties with this, especially at the step to enable it
         using an activation code. If you have difficulties you may prefer to set up ``gdbserver`` as described in :doc:`setup_s32k148_gdbserver`
         with an IDE of your choice.
+    * - 4
+      - P&E Micro provide versions of gdb server for Windows, Linux and Mac that work with their OpenSDA USB drivers
+        to support flashing and debugging on the board via USB. See :ref:`s32k148_gdb_via_usb`.
+        However the USB drivers are only available for Windows and Linux on x86 based hosts, not for ARM,
+        and so debugging on the board from ARM hosts (eg. a Raspberry Pi or an ARM-based Mac) requires an external debugger,
```
ThoFrankACN, line 62: For me it's actually possible to use the intel-native binary on ARM-based Macs through Rosetta. And I didn't need to install any additional drivers. So I would propose to remove all mentions of Macs here. We also don't officially support macOS anyways.
@frankdarcyacn

executables/referenceApp/platforms/s32k148evb/main/CMakeLists.txt
ThoFrankACN: It's assigned to `app.referenceApp` in `executables/referenceApp/application/CMakeLists.txt`:
```
if (TARGET startUp)
    get_target_property(LINKER_SCRIPT startUp PROP_LINKER_SCRIPT)

    if (DEFINED LINKER_SCRIPT)
        target_link_options(app.referenceApp PRIVATE -T ${LINKER_SCRIPT})
    endif ()
endif ()
```

PR #82, opened 2025-03-21
Update I/O configuration to support S32K148 176 LQFP

marcmo:
originates from CR #127357


evengrinovich:
@alinemundt  you are right, removed it from the commit message

PR #84, opened 2025-03-25
Use #pragma once

rolandreichweinbmw:
Using #pragma once removes the burden of generating and maintaining include guards and removes the risk of associated problems on refactoring.

See also:

* https://en.wikipedia.org/wiki/Pragma_once


jiangyilism (approved):
CMake switched to `pragma once` years ago:
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5171

They also have a dedicate clang-tidy checker "cmake-use-pragma-once" to auto transform/fix header guard: 
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7933#291cf0aba9b42819466e7e30c7356ed4934a8233

We can borrow the idea too when we someday plan to integrate clang-tidy into openbsw.

marcmo:
looks good ;) but what about portability when using compilers like Diab?

rolandreichweinbmw:
As far as I can see, the Diab compiler since 7.x is LLVM based and GCC compatible. I.e. supports #pragma once.

rolandreichweinbmw:
General compiler support is listed here: https://en.wikipedia.org/wiki/Pragma_once#Portability

marcmo (approved):
LGTM

PR #85, opened 2025-03-26
Integrate sscache for faster unit build compilation

shamitha-shashidhara:
- Utilize sscache to speed up the compilation process for unit builds.
- Eliminate the need for Docker containers in the CI workflow to simplify the setup.
- Implement actions/cache to store cmake-build-unit-tests artifacts to ensure these cached artifacts can be reused across different workflows to improve efficiency.

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -1,85 +1,95 @@
 name: Code Coverage 
 
-on:
-  workflow_call:
-  push:
-    branches:
-      - main
-  pull_request:
-
-env:
-  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
-  BUILD_TYPE: Debug
+on: [workflow_call, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     permissions:
       contents: write
 
+    env:
+      BUILD_TYPE: Debug
+      SCCACHE_GHA_ENABLED: "true"
+
     steps:
-    - uses: actions/checkout@v4
+    - name: Checkout code
+      uses: actions/checkout@v4
 
-    - name: Pull and Run Docker with GCC 11.4.0 and LCOV
+    - name: Install GCC 11.4.0 and LCOV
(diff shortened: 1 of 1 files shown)


.github/workflows/code-coverage.yml
```diff
@@ -1,85 +1,95 @@
...
-          #Badges
-          LinePercentage=\$(lcov --summary cmake-build-unit-tests/coverage.info | grep 'lines' | awk '{print \$2}') &&
-          echo \"Line Percentage: \$LinePercentage\" &&
-          wget \"https://img.shields.io/badge/coverage-\${LinePercentage}25-brightgreen.svg\" -O line_coverage_badge.svg &&
-
-          FunctionPercentage=\$(lcov --summary cmake-build-unit-tests/coverage.info | grep 'functions' | awk '{print \$2}') &&
-          echo \"Function Percentage: \$FunctionPercentage\" &&
-          wget \"https://img.shields.io/badge/coverage-\${FunctionPercentage}25-brightgreen.svg\" -O function_coverage_badge.svg &&
+        sudo apt-get update
+        sudo apt-get install -y gcc-11 g++-11 lcov wget zip
```
jiangyilism, line 22: Why is it building with gcc-11 instead of the version used in `.github/workflows/build.yml` ?

shamitha-shashidhara: The versions of gcov and lcov we use for code coverage analysis are aligned with GCC 11. Using a different version of GCC and gcov will lead to mismatched coverage data, as seen with the GCC version used in .github/workflows/build.yml before, which caused the` geninfo: ERROR: mismatched end line` error. This issue was fixed in commit by changing GCC version to 11 [c31f0104b891f50eae2f75e4897c7f990a765ea0].

marcmo: using the same compiler version would be of course nicer...could we use the version defined in the build.yml and find a matching version for the gcov tools?

marcmo: I will create an issue for that and merge this PR since it will speed up our verification a lot!

PR #91, opened 2025-04-07
Fix: CI breaks build on removed files

shamitha-shashidhara:
- Removed restore keys option from actions/cache@v3 to prevent restoring caches that are close but not an exact match.
- Deleted cmake-build-unit-tests directory before running unit test build to ensure a clean build environment.
- Added --ignore-errors source to genhtml command to prevent HTML generation failures if files are not found.

This commit addresses the "cannot read" error caused by old cache contents.


marcmo (approved):
LGTM

PR #93, opened 2025-04-08
Fix: cache storage legacy service is shutting down error

shamitha-shashidhara:
Updated the sccache action to the latest version (v0.0.8). This change addresses the cache storage error caused by the decommissioning of the legacy cache service.


marcmo (approved):
LGTM

PR #110, opened 2025-05-27
Fix wrong byte in Flow Control Frame.

maximilianraucheckerbmw:
Sending a UDS request which doesnt fit into one single CAN frame (e.g. WDBI with larger payload) the ECU responds with the FC frame

vcan0   0F0   [8]   30 0F F0 FF FF FF FF FF

to the tester. Here, the third byte representing the STmin byte (0xF0) seems to be invalid according to ISO-TP. This seems to come from a parameter used in DoCANSystem in the reference executable. To ensure the STmin byte in the FC frame is always valid according to ISO-TP, we also modify the encoding function.


a-elkamel (approved):
LGTM!

matthiaskessler:
Thanks for the fix. I think, in a future commit, we should think about adding the unit of the constants to the name, which just makes it a little easier to read, e.g. MIN_SEPARATION_TIME_US, RX_TIMEOUT_MS.

PR #114, opened 2025-06-04
Show build status of Posix, S32k148

SuhashiniNaik:
-  This badge displays the build status of the latest push event on the main branch, including merges from pull requests.

image

Clicking on the badge redirects to the corresponding GitHub Actions workflow run:
image

Output can be viewed from here: [link](https://github.com/esrlabs/openbsw/tree/StatusBadge?tab=readme-ov-file#build-status-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,10 +3,7 @@
 
 ## Build Status 🚀
 
-| Platform       | Status                                                                 |
-|----------------|------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


SuhashiniNaik:
Hi @vladyslavmarkovaccenture

We currently run both the s32k148 and POSIX builds within the same workflow .yml file. This is intentional, as separating them into different workflows would lead to code duplication particularly in the compiler setup, build logic, and environment configurations that are shared between the two.
Regarding the status badges: they now link to the `build.yml` workflow run page [Build S32k and posix platform](https://github.com/eclipse-openbsw/openbsw/actions/workflows/build.yml) that includes both s32k148 and POSIX builds.

PR #118, opened 2025-06-05
Add additional services in UdsTool

marcmo:
originates from CR #131660

### DIFF
--- a/tools/UdsTool/app/services.py
+++ b/tools/UdsTool/app/services.py
@@ -9,10 +9,8 @@ def read(client, did):
 
     # Create request
     req = uds.ReadDataByIdentifier.make_request([did], {did: udsoncan.AsciiCodec(4)})
-
     # Send a request to the ECU
     response = client.send_request(req)
-
     # Print the response
     return helper.handleResponse(response)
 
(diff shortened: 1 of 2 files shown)


tools/UdsTool/app/services.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Print the response
+    return helper.handleResponse(response)
+
+
+def write(client, did, data):
+    did = int(did, 16)
+    data_len = len(data)
+
+    # Create request
```
hbragge-acn: I think it's unnecessary to repeat identical comments in the same source file

tools/UdsTool/udsTool.py
```diff
@@ -22,11 +27,172 @@ def start():
...
         # Handle Ethernet connection
-        client = application.createEthConnection(host, ecu, source)
-        application.read(client, did)
+        client = connection.createEthConnection(host, ecu, source)
+        services.read(client, did)
+
+
+# UDS service Write Data By Identifier (0x2e)
+@start.command(help="UDS service Write Data By Identifier (0x2e)")
+@click.option("--can", is_flag=True, help="Use CAN connection")
```
hbragge-acn: Please extract the common parameters into a single place, for example by using this pattern (from https://github.com/pallets/click/issues/108#issuecomment-280489786):
common_params

hbragge-acn (changes requested):
Looks pretty good and seems to work, I added two comments

tools/UdsTool/udsTool.py
KevShaju: Thank you for the suggestion. I have made the requested changes. Please check.

PR #125, opened 2025-06-23
Add ThreadX posix support

marcmo:
originates from CR #132457


executables/referenceApp/application/src/app/app-threadx.cpp
```diff
@@ -0,0 +1,228 @@
...
+void run()
+{
+    printf("hello\r\n");
+    staticInit();
+    tx_kernel_enter();
+}
+
+extern "C"
+{
+void tx_application_define(void* first_unused_memory)
```
jiangyilism, line 110: Can we merge app-threadx.cpp and app-freetos.cpp ?

The start-up can be like:

for threadx:
```
main() -> app_run_stage_0() /* threadx-specific */ -> 
tx_kernel_enter() (noreturn) -> tx_application_define() -> app_run_stage_1() /* shared by all rtos. openbsw api calls... */ -> AsyncAdapter::run()
```
for FreeRTOS:
```
main() -> app_run_stage_0() /* freertos-specific */ -> app_run_stage_1() /* shared by all rtos. openbsw api calls... */ -> AsyncAdapter::run() -> vTaskStartScheduler() (noreturn)
```

executables/referenceApp/platforms/s32k148evb/main/src/bsp/startUp-threadx.S
```diff
@@ -0,0 +1,501 @@
...
+    .long   DefaultISR                                      /* Bus Fault Handler*/
+    .long   DefaultISR                                      /* Usage Fault Handler*/
+    .long   0                                               /* Reserved*/
+    .long   0                                               /* Reserved*/
+    .long   0                                               /* Reserved*/
+    .long   0                                               /* Reserved*/
+    .long   DefaultISR                                      /* SVCall Handler*/
+    .long   DefaultISR                                      /* DebugMon_Handler                                /* Debug Monitor Handler*/
+    .long   0                                               /* Reserved*/
+    .long   __tx_PendSVHandler                              /* PendSV Handler*/
```
jiangyilism, line 26: Most of startUp-threadx.S and startUp-freertos.S are duplicated.
We can move threadx-specific code like __tx_SysTickHandler to another file.
And override PendSVHandler and SysTickHandler in cmake.
e.g.
```
.long   PendSVHandler
.long   SysTickHandler 
```
and in cmake:
```
if (threadx)
    target_add_definitions(... private
        PendSVHandler=__tx_PendSVHandler
        SysTickHandler=__tx_SysTickHandler)
elseif (freertos)
    target_add_definitions(... private
        PendSVHandler=PendSV_Handler
        SysTickHandler=SysTick_Handler)
endif ()
```

libs/bsw/asyncThreadX/include/async/EventManager.h
```diff
@@ -0,0 +1,78 @@
+// Copyright 2021 Accenture.
+
+/**
+ * \ingroup async
+ */
+#ifndef GUARD_9337F16A_502D_440E_A3D7_4516CD2AB6DF
+#define GUARD_9337F16A_502D_440E_A3D7_4516CD2AB6DF
```
jiangyilism: `#ifndef ...` -> `#pragma once`
Also, you might want to change all the "Copyright ..." to "Copyright 2025".

platforms/posix/bsp/bspInterruptsImpl/CMakeLists.txt
```diff
@@ -1,7 +1,13 @@
...
-target_include_directories(bspInterruptsImpl PUBLIC include)
+if (BUILD_TARGET_PLATFORM MATCHES "FREERTOS")
+    add_library(bspInterruptsImpl
+                freertos/src/interrupts/suspendResumeAllInterrupts.cpp)
+    target_include_directories(bspInterruptsImpl PUBLIC freertos/include)
+    target_link_libraries(bspInterruptsImpl PRIVATE freeRtos)
+elseif (BUILD_TARGET_PLATFORM MATCHES "THREADX")
+    add_library(bspInterruptsImpl
+                threadx/src/interrupts/suspendResumeAllInterrupts.cpp)
+    target_include_directories(bspInterruptsImpl PUBLIC threadx/include)
```
jiangyilism, line 9: Is creating a cmake variable `rtos` more scalable for adding other rtos types?
For example `set(rtos "threadx")` or `set(rtos "freertos")`. Then we can do 
`target_include_directories(bspInterruptsImpl PUBLIC "${rtos}/include")` without if-else branching and matching `BUILD_TARGET_PLATFORM` regex. Can also simplify many other cmake codes.

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
```diff
@@ -0,0 +1,283 @@
...
+        ContextType context,
+        RunnableType& runnable,
+        TimeoutType& timeout,
+        uint32_t delay,
+        TimeUnitType unit);
+
+    static void cancel(TimeoutType& timeout);
+
+private:
+    struct TaskInitializer : public StaticRunnable<TaskInitializer>
```
jiangyilism, line 99: Can threadx also provide a `libs/bsw/asyncThreadX/include/async/TaskInitializer.h` which provides `IdleTask` and `TimerTask` like freertos?

ermankomurcu: ThreadX includes a hidden system timer thread (similar to the TimerTask), which runs at the highest priority (0). However, it does not provide an idle task by default. To handle background activities such as logging and command processing, we can create a BackgroundTask thread with the lowest priority. In fact, this BackgroundTask already exists in the example application and is currently used for logging and command handling. 
I think there's no need to implement a separate system timer thread, as ThreadX creates it internally. Instead, we can provide the BackgroundTask as a low-priority thread to handle background operations.

platforms/posix/bsp/bspInterruptsImpl/CMakeLists.txt
ermankomurcu: I think the same way, and I have already added the "rtos" parameter to the CMake options. It will be included when the commit I prepared is pushed to the PR.

libs/bsw/asyncThreadX/include/async/EventManager.h
ermankomurcu: "#ifdef"s will be replaced with the "pragma" and copyright text will be updated.

executables/referenceApp/platforms/s32k148evb/main/src/bsp/startUp-threadx.S
ermankomurcu: I agree. We can do that since this is low level code, end users typically do not bother reading it.

executables/referenceApp/application/src/app/app-threadx.cpp
ermankomurcu: We can merge these codes. However, adding support for other RTOSes will increase complexity and reduce readability due to additional logic and #ifdef blocks. Despite the extra maintenance cost and code redundancy, I believe having a separate app file for each RTOS will make it easier for end users to follow and adopt the code.

mthiede-acn2: In general, the idea would be to hide the OS specific parts in the respective async adapters.

jiangyilism: However, app-freertos.cpp and app-threadx.cpp have many duplicate code.
In my above approach,
app_run_stage_1(), which contains most content of this file, is shared by both threadx and freertos.
app_run_stage_0() is per-rtos, either branch on rtos types by #ifdef , or put it in different files like run_stage_0_freertos.cpp/run_stage_0_threadx.cpp and branch in CMake layer.


freertos app_run_stage_0 just directly calls app_run_stage_1
threadx app_run_stage_0 calls tx_kernel_enter(). And app_run_stage_1 is called in the end of  tx_application_define()

platforms/posix/bsp/bspInterruptsImpl/CMakeLists.txt
ermankomurcu: cmake rtos parameter is added

libs/bsw/asyncThreadX/include/async/EventManager.h
ermankomurcu: done.

executables/referenceApp/platforms/s32k148evb/main/src/bsp/startUp-threadx.S
ermankomurcu: This will be addressed.

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
```diff
@@ -0,0 +1,292 @@
...
+    {
+        return CONTEXT_INVALID;
+    }
+}
+
+template<class Binding>
+void ThreadXAdapter<Binding>::initTask(TaskInitializer& initializer)
+{
+    ContextType const context    = initializer._context;
+    TaskContextType& taskContext = _taskContexts[static_cast<size_t>(context) - 1];
```
jiangyilism, line 169: In libs/bsw/asyncThreadX/include/async/Types.h ,  `CONTEXT_INVALID` is defined to be `0xFF` so `0` should be a valid context. But that will lead to invalid mem access `_taskContexts[-1]`

executables/referenceApp/application/src/logger/logger.cpp
```diff
@@ -65,8 +71,14 @@ ComponentConfigType loggerComponentConfig(loggerComponentMapping);
 LoggerComposition loggerComposition(loggerComponentMapping, "RefApp");
 
 DefaultLoggerCommand loggerCommand(loggerComponentConfig);
-::console::AsyncCommandWrapper
-    asyncCommandWrapper(loggerCommand.root(), ::async::AsyncBinding::AdapterType::TASK_IDLE);
+
+::console::AsyncCommandWrapper asyncCommandWrapper(
+    loggerCommand.root(),
+#ifdef SUPPORT_FREERTOS
```
jiangyilism, line 77: Can we abstract this in `AdapterType` ?
e.g.
`::async::AsyncBinding::AdapterType::TASK_CONSOLE` (= `TASK_IDLE` or `TASK_BACKGROUND`)
or 
`::async::AsyncBinding::AdapterType::TASK_DEFAULT` ?

platforms/posix/bsp/bspInterruptsImpl/threadx/src/interrupts/suspendResumeAllInterrupts.cpp
```diff
@@ -0,0 +1,34 @@
+// Copyright 2025 Accenture.
+
+#include "interrupts/suspendResumeAllInterrupts.h"
+
+#include <tx_api.h>
+#include <unistd.h>
+
+static bool threadXInitialized{false};
+
+void setThreadXInitialized() { threadXInitialized = true; }
```
jiangyilism, line 10: Is this a temp workaround or the intended implementation?

executables/referenceApp/asyncCoreConfiguration/threadx/include/async/Config.h
```diff
@@ -0,0 +1,43 @@
+// Copyright 2025 Accenture.
+
+#pragma once
+
+#define ASYNC_CONFIG_NESTED_INTERRUPTS (1)
```
jiangyilism, line 5: In freertos, this is defined in `libs/bsw/asyncFreeRtos/freeRtosConfiguration/FreeRTOSConfig.h` instead of the ref app.
To be consistent,
option 1. In threadx, moving it to tx_user.h (new file, a config file like FreeRTOSConfig.h)
option 2. In freertos, moving it to executables/referenceApp/asyncCoreConfiguration/freertos/include/async/Config.h

thomas-herrmann-1: In FreeRTOS the macro `ASYNC_CONFIG_NESTED_INTERRUPTS` _can_ be defined in `async/Config.h`. What you see in `FreeRTOSConfig.h` is only a default definition of the macro in case the application doesn't explicitly configure it. If we need such for ThreadX (i.e. it makes a difference in implementation and usage) then we should follow a similar approach: let the application configure whether it requires nested interrupts in `async/Config.h`. And then do whatever is appropriate to implement it for the specific OS. Having something like default values for such macros, typically defined in an async file that includes `async/Config.h` is still a way to go.

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
thomas-herrmann-1: We anyway need to discuss about idle task implementation for ThreadX. So yes, if it turns out that we've no special treatment of Idle task (unlike as for FreeRTOS) then 0 might be a valid task context. But - even if  `CONTEXT_INVALID` is defined to be `0xff` - to me this doesn't necessarily mean that it's crucial to have 0 available for ordinary task contexts.

executables/referenceApp/application/src/logger/logger.cpp
thomas-herrmann-1: `async` was never designed to support a certain applicational usage, therefore I wouldn't chose `TASK_CONSOLE` to be defined for the lowest prio task. And `TASK_DEFAULT` wouldn't express something very useful to me. We anyway found out that we need to discuss the role of Idle task in context of ThreadX and whether ThreadX and FreeRTOS (which itself uses `IDLE hook` to identify the special role) implementations of async should be completely aligned.

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
```diff
@@ -0,0 +1,292 @@
...
+ * `Binding` type to adapt and configure task-related components, such as `TaskContext`,
+ * `TaskConfig`, and backgroud task.
+ *
+ * \tparam Binding The binding type specifying application-specific configurations.
+ */
+template<class Binding>
+class ThreadXAdapter
+{
+public:
+    static size_t const TASK_COUNT           = Binding::TASK_COUNT;
```
jiangyilism, line 42: Suggest to add:
`static_assert(TASK_COUNT <= TX_MAX_PRIORITIES, "TASK_COUNT exceeds TX_MAX_PRIORITIES");`

ermankomurcu: I agree and this check will be added.

ermankomurcu: I agree and this check will be added.

PR #126, opened 2025-06-26
Add EEPROM driver for POSIX

marcmo:
originates from CR #132321


platforms/posix/bsp/bspEepromDriver/src/eeprom/EepromDriver.cpp
```diff
@@ -0,0 +1,93 @@
...
+        return ::bsp::BSP_ERROR;
+    }
+    return ::bsp::BSP_OK;
+}
+
+bsp::BspReturnCode
+EepromDriver::write(uint32_t const address, uint8_t const* const buffer, uint32_t const length)
+{
+    bool success
+        = ((-1 != eepromFd) && (address < EEPROM_SIZE) && ((address + length) < EEPROM_SIZE));
```
hbragge-acn: Just noticed that there seems to be an off-by-one error in this check: the last EEPROM byte can't be written. You can fix it like this (btw. this also avoids the potential overflow when adding integers):
`bool success = ((-1 != eepromFd) && (address < EEPROM_SIZE) && (length <= (EEPROM_SIZE - address)));`

Please also add a corresponding unit test, to make it easier to verify that the issue is fixed.

platforms/posix/bsp/bspEepromDriver/src/eeprom/EepromDriver.cpp
```diff
@@ -0,0 +1,93 @@
...
+        return ::bsp::BSP_ERROR;
+    }
+    return ::bsp::BSP_OK;
+}
+
+bsp::BspReturnCode
+EepromDriver::read(uint32_t const address, uint8_t* const buffer, uint32_t const length)
+{
+    bool success
+        = ((-1 != eepromFd) && (address < EEPROM_SIZE) && ((address + length) < EEPROM_SIZE));
```
hbragge-acn: Same as above

hbragge-acn (changes requested):

platforms/posix/bsp/bspEepromDriver/test/gtest/src/eeprom/EepromDriverTest.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+    EXPECT_EQ(::bsp::BSP_OK, _cut.init());
+
+    uint8_t dataToWrite[] = {0xAB};
+    uint32_t address      = EepromDriverTest::EEPROM_SIZE - 1;
+    EXPECT_EQ(::bsp::BSP_OK, _cut.write(address, dataToWrite, 1));
+
+    uint8_t readData[1] = {0};
+    EXPECT_EQ(::bsp::BSP_OK, _cut.read(address, readData, 1));
+    EXPECT_EQ(dataToWrite[0], readData[0]);
```
mthiede-acn2: The write() method which is being tested could have modified the dataToWrite[0], e.g. to be 0. in that case the test would be false positive

amit-warbhe: Done

PR #128, opened 2025-06-28
Switch from ESTD to ETL using closure

rolandreichweinbmw:
This is a variant of https://github.com/eclipse-openbsw/openbsw/pull/58 using etl::delegate instead of etl::inplace_function.

This is possible by using the newly introduced etl::closure as already merged in upstream ETL.


rolandreichweinbmw:
To prevent confusion, we can continue this in te original PR https://github.com/eclipse-openbsw/openbsw/pull/58 now. Generally now using etl::delegate and etl::closure in async Call+Function.

PR #137, opened 2025-07-04
Remove dependencies on std

rolandreichweinbmw:
By replacing std usage with etl counterparts, code will compile in more cases in ETL_NO_STL mode.

This is preparational work for the upcoming ETL release from which we want to include etl::closure and in turn etl::tuple. This requires decoupling etl from std if we want to keep ETL_NO_STL mode.


marcmo (approved):
LGTM

PR #142, opened 2025-07-11
Update ETL to version 20.42.2

rolandreichweinbmw:
New ETL contains etl::closure, so we can now rely on the interface as agreed upstream. This PR also removes the own provided closure interface from etl_addons.

Further, etl::tuple can now be used without some compile error cases.


hbragge-acn (approved):
Looks good

jiangyilism (approved):

hbragge-acn (approved):

PR #144, opened 2025-07-30
Add ethernet support

marcmo:
originates from CR #133582


executables/referenceApp/application/src/systems/DemoSystem.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Accenture.
+// Copyright 2025 Accenture. All rights reserved. Accenture proprietary and confidential material.
```
rolandreichweinbmw: Why? Isn't it covered by the Apache license?

ghost: Thank you, fixed

rolandreichweinbmw: Consider executables/referenceApp/application/src/systems/EthernetSystem.cpp also?

ghost: Missed that one, thank you.

PR #154, opened 2025-08-14
Added new Uart interface

cristiancsordas:
Creates a generic uart interface. Adapt the S32k Uart implementation to this new interface


johannes-esr:
@cristiancsordas, it seems like you have no valid ECA signed. This is mandatory to contribute to this project. Please check https://www.eclipse.org/projects/handbook/ for details.

rolandreichweinbmw:
> @cristiancsordas, it seems like you have no valid ECA signed. This is mandatory to contribute to this project. Please check https://www.eclipse.org/projects/handbook/ for details.

We just made sure the ECA for @cristiancsordas is signed, see the linked Eclipse page. I hope the Eclipse API picks this up soon.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,78 @@
...
+namespace bsp
+{
+/**
+ * This class implements the UART communication for S32K1xx platforms.
+ * It inherits from the base class UartApi and implements the methods for writing and reading data
+ * over the UART interface. It also provides a method for initializing the UART communication.
+ */
+class Uart
+: public bsp::UartApi
+, public bsp::UartConfig
```
rolandreichweinbmw: While the intention is to define a common interface bsp::UartApi and implement it in a concrete bsp::Uart, it is not technically enforced that bsp::Uart actually implements bsp::UartApi (as done with virtual interfaces using pure virtual functions and "override").

I don't propose virtual here, but we need to sort this issue out.

E.g. well documenting that the bsp::UartApi has technically only documentational purpose?

rolandreichweinbmw:
To resolve the function override issue, we have different options now:
1. Define an API header in libs/bsp/... and implement it in the BSP
2. Use CRTP for that
3. Use interface definition in libs/bsp with (pure) virtual interfaces and implement in a derived class in the platform's BSP

If we can assume that only one BSP is used at build time, the first two options are more memory and runtime efficient.

platforms/s32k1xx/bsp/bspUart/src/Uart.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    return counter;
+}
+
+bool Uart::writeByte(uint8_t data)
+{
+    if (!isTxActive())
+    {
+        _uartDevice.uart.DATA = (static_cast<uint32_t>(data) & 0xFFU);
+        while (isTxActive()) {}
```
maslovw: infinite blocking, always use a counter when waiting for hardware reaction with error handling, timeout should be configurable either statically or by argument passing

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
maslovw: Diamond Inheritance Risk: When you have class Uart : public UartApi, public UartConfig, you create potential for diamond inheritance if either base class shares a common ancestor. This leads to ambiguous method calls and memory layout issues. 
I'd try to minimize usage of multiple inheritance and leave it only when it's need and makes sense. 

Multiple inheritance creates an "is-a" relationship with both classes. But Uart isn't really "a kind of UartConfig" - it has configuration. This violates the Liskov Substitution Principle.

Changes to either base class can break the derived class in unexpected ways. For embedded systems where reliability is critical, this increases risk.

  Better Alternative - Composition:
  class Uart : public UartApi
  {
  private:
      UartConfig _config;  // HAS-A relationship instead of IS-A
  public:
      static Uart& getInstance() const;
  };

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,78 @@
...
+     * this method must be called before using the read/write methods
+     */
+    void init();
+
+    /**
+     * factory method which instantiates and configures an UART object.
+     * If the object exists it will returns only a reference to it.
+     * @param id: TERMINAL, ...
+     */
+    static Uart& get(Id id);
```
maslovw: nameless `get` might be not obvious, it's better to avoid simple name like `get` and `set` for big class

executables/unitTest/bsp/bspUart/include/Uart.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include "bsp/uart/UartApi.h"
+
+namespace bsp
+{
+
+class Uart : public bsp::UartApi
+{
+public:
+    size_t write(uint8_t const* data, size_t length);
```
maslovw: size_t as return type might be sufficient, but using dedicated data type with error code is more useful. 
Read and Write might fail because of different issues, and it's much easier to debug if code can return something useful 
smth like: 
```
 enum class UartResult
  {
      SUCCESS,
      TIMEOUT,
      HARDWARE_ERROR,
      BUFFER_FULL,
      INVALID_PARAMETER
  };

  struct UartWriteResult
  {
      UartResult result;
      size_t bytesWritten;
  };

```

or even more generic to use across whole bsw: 

 ```
// Common result type for all BSW drivers
  template<typename T>
  struct BswResult
  {
      BswErrorCode error;
      T value;

      bool isSuccess() const { return error == BswErrorCode::SUCCESS; }
      T& operator*() { return value; }
  };

  // All BSW APIs follow this pattern
  BswResult<size_t> write(uint8_t const* data, size_t length);
  BswResult<CanFrame> canReceive(uint32_t timeoutMs);
  BswResult<uint16_t> adcRead(AdcChannel channel);

```

Comment (rolandreichweinbmw) [reply executables/unitTest/bsp/bspUart/include/Uart.h]: Before anyone implements BswResult, please note that we already have etl::expected (backport from C++23 std::expected).

File: platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h (hunk 1 above)
Comment (cristiancsordas) [reply platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h]: -  The diamond problem would arise in a situation like this:
class Uart1: public bsp::UartApi
{
public:
size_t write(uint8_t const* data, size_t length);
};
class Uart2: public bsp::UartApi
{
public:
size_t write(uint8_t const* data, size_t length);
};
class UartDeriv: public Uart1, public Uart2
{};
This Use-case will not hapen in a real case scenario. 
Please detail with an example for the UartApi where the diamond inheritance will create a potential risk.

- UartConfig class defines the available Uarts for this specific platform. The enum class Id, defined in the UartConfig, could have been part of the Uart class. But, having the UartConfig class separed from the Uart creates a better separation and a more readable code.
For example, with this approach calling the init() method for the TERMINAL Uart will result in something like this:
Uart::get(Uart::TERMINAL).init();
Therefore, the intention here was to have an "is-a" relation, 
The UartDevice has a "has-a" relation and it provides the settings of a specific Uart.

**Comment by aschaal** [comment]
I have added some comments to #198, most of them apply here as well.
If it turns out that #154 is favored over #198, I will add specific comments here as well.

File: executables/unitTest/bsp/CMakeLists.txt [hunk 5]
```diff
@@ -2,3 +2,4 @@ add_subdirectory(bspConfiguration)
 add_subdirectory(bspIo)
 add_subdirectory(bspMcu)
 add_subdirectory(bspSci)
+add_subdirectory(bspUart)
```
aschaal: This should not be here. Unfortunately our guidelines are not explaining this very good yet, I am currently improving it.

The executable/unitTest folders shall contain "configuration" used across multiple unit tests.
What I can see here, you provide a mock. Please move that into the bspUart module: "bspUart/mock"

Thank you!

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,99 @@
+Low level interfaces
```
aschaal: Please use "Title Case".

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,99 @@
...
+====================
+
+This section describes the conventions to be followed when defining low level interfaces in the BSW.
```
aschaal: Do you mean "in the BSP"?

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,99 @@
...
+
+Naming Conventions
+------------------
+- Interface names should end with the suffix "Api" to clearly indicate their purpose as interfaces. For example, a UART interface should be named `UartApi`.
+- Interface methods should be prefixed with a verb that clearly indicates the action being performed, such as "get", "set", "init", "read", or "write". For example, a method to read data from a UART interface could be named `read()` or `readData()`.
+
+Design Conventions
+------------------
+- Interfaces should be designed to be as generic as possible, allowing for multiple implementations. This promotes code reusability and flexibility.
```
aschaal: In general, try to follow the column limit convention of 100 chars where possible.

doc/dev/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,99 @@
...
+- The static interfaces should be validated if the compiler supports C++20 or later using concept checks to ensure that the implementing classes adhere to the defined interface. This can be achieved using `static_assert` statements in combination with C++20 concepts.
+- Provide clear and concise documentation for each interface and its methods, including descriptions of parameters, return values.
+- The interface should be placed in a dedicated header file, typically named after the interface itself (e.g., `UartApi.h` for the `UartApi` interface).
+- Declaration of interfaces using pure virtual methods should be also possible, but for performance reasons, prefer static interfaces with concept checks where applicable.
+- The low level interfaces should be placed in the bsp namespace.
+- Isolate the configuration and platform-specific details from the interface definition to ensure portability across different platforms.
+- Configure the concrete implementations of the interfaces in separate configuration files or classes, allowing for easy adaptation to different hardware or software environments.
+
```
aschaal, line 28: Please explain in a separate section " The Three Approaches":
- Interface without virtual
- Interface with virtual
- Static Interface for thinks like SystemTimer which exists just once

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
```diff
@@ -0,0 +1,36 @@
+// Copyright 2024 Accenture.
+// Copyright 2025 BMW AG
+
+#include <bsp/UartPrivate.h>
```
aschaal: Why is that file called UartPrivate.h and not UartConfig.h? The cpp file is called UartConfig.
You also specify the UartConfig struct in this header...

doc/dev/guidelines/conventions/interfaces.rst
aschaal: Btw, a nice diagram (per approach) would be helpful, but you can add that later.

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,99 @@
...
+- The static interfaces should be validated if the compiler supports C++20 or later using concept checks to ensure that the implementing classes adhere to the defined interface. This can be achieved using `static_assert` statements in combination with C++20 concepts.
+- Provide clear and concise documentation for each interface and its methods, including descriptions of parameters, return values.
+- The interface should be placed in a dedicated header file, typically named after the interface itself (e.g., `UartApi.h` for the `UartApi` interface).
+- Declaration of interfaces using pure virtual methods should be also possible, but for performance reasons, prefer static interfaces with concept checks where applicable.
+- The low level interfaces should be placed in the bsp namespace.
+- Isolate the configuration and platform-specific details from the interface definition to ensure portability across different platforms.
+- Configure the concrete implementations of the interfaces in separate configuration files or classes, allowing for easy adaptation to different hardware or software environments.
+
+Example
```
aschaal: I am struggling a bit with UartApi as example (for the "first approach"). Maybe better "ExampleApi"?
Then you can simplify it a bit. 

You can still refer to Uart as "complete working blueprint".

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,99 @@
...
+-------
+.. code-block:: cpp
```
aschaal, line 44: Personally, I would prefer one code-block per file for better readability.

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,99 @@
...
+
+    // UartApi.h
```
aschaal: Add the "complete" path, e.g. "libs/bsw/bsp/include/bsp/example/ExampleApi.h" or
"platforms/<platform name>/bsp/bspExample/include/bsp/Example.h"

doc/dev/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,99 @@
...
+        static_assert(                       \
+            bsp::UartCheckInterface<_class>, \
+            "Class " #_class " does not implement UartApi interface correctly");
+
+    #else
+    #define BSP_UART_CONCEPT_CHECKER(_class)
+    #endif
+
+    // Uart.h
+    #pragma once
```
aschaal, line 121: I see that you tried to squeeze the code, but please try to use correctly formatted snippets (with some empty lines, etc.). In my eyes this would look more consistent.

executables/unitTest/bsp/CMakeLists.txt
cristiancsordas: bspUart removed from executables/unitTest/bsp. Created the UartMock class

doc/guidelines/conventions/interfaces.rst
cristiancsordas: changed to "Low Level Interfaces"

doc/guidelines/conventions/interfaces.rst
cristiancsordas: changed to "BSP"

doc/dev/guidelines/conventions/interfaces.rst
cristiancsordas: I added to the documentation the "Three approaches"

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
cristiancsordas: Renamed the file UartPrivate.h to UartConfig.h

doc/guidelines/conventions/interfaces.rst
cristiancsordas: Renamed Uart to Example. At the bottom of the file in the References is Uart mentioned as example

doc/guidelines/conventions/interfaces.rst
cristiancsordas: added one code-block per file

doc/guidelines/conventions/interfaces.rst
cristiancsordas: added the complete path of the file location

doc/dev/guidelines/conventions/interfaces.rst
cristiancsordas: added empty lines after "#pragma once" or after the file names

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,214 @@
...
+- Provide clear and concise documentation for each interface and its methods, including descriptions of parameters, return values.
+- The interface should be placed in a dedicated header file, typically named after the interface itself (e.g., `ExampleApi.h` for the `ExampleApi` interface).
+- Declaration of interfaces using pure virtual methods should be also possible, but for performance reasons, prefer static interfaces with concept checks where applicable.
+- The low level interfaces should be placed in the bsp namespace.
+- Isolate the configuration and platform-specific details from the interface definition to ensure portability across different platforms.
+- Configure the concrete implementations of the interfaces in separate configuration files or classes, allowing for easy adaptation to different hardware or software environments.
+
+Approaches to Interface Definition
+----------------------------------
+There are three possible approaches to defining interfaces:
```
aschaal: approaches to define interfaces

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,214 @@
...
+using pure virtual classes (abstract base classes), using static interfaces with concept checks, and using static interface methods.
```
aschaal: Again, we have a convention to stick to 100 chars column limit, please check all files and line break them if > 100 chars.

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,214 @@
...
+    public:
+        static size_t read(::etl::span<uint8_t> data);
+        static size_t write(::etl::span<uint8_t const> const& data);
+    };
+    } // namespace bsp
+
+Configuration of Concrete Implementations
+-----------------------------------------
+- Configuration details for concrete implementations should be separated from the interface definitions.
+- Use configuration structures or classes to encapsulate platform-specific settings.
```
aschaal: Maybe "platform- and application-specific settings" as the settings are projects dependent...

platforms/s32k1xx/bsp/bspUart/test/mock/include/UartMock.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
```
aschaal: Please add the mock to UT build and instantiate it to ensure it works properly.

platforms/s32k1xx/bsp/bspUart/test/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+add_executable(bspUartTest src/IncludeTest.cpp)
```
aschaal, line 1: "add_subdirectory" needs to be added to CMakeLists in OpenBSW root (in the case of BUILD_UNIT_TESTS and S32K1XX).

platforms/s32k1xx/bsp/bspUart/include/bsp/UartConfig.h
```diff
@@ -0,0 +1,26 @@
...
+    uint32_t baud;
+};
+
+struct Uart::UartConfig
+{
+    LPUART_Type& uart;
+    bios::Io::PinId txPin;
+    bios::Io::PinId rxPin;
+    uint8_t speedModes;
+    UartBaudRate const* baudRate;
```
christian-schilling: Why is this a pointer field? Needs an explaining comment.

platforms/s32k1xx/bsp/bspUart/src/Uart.cpp
```diff
@@ -0,0 +1,98 @@
...
+    _uartConfig.uart.STAT   = 0;
+    _uartConfig.uart.MODIR  = 0;
+
+    _uartConfig.uart.FIFO  = 0;
+    _uartConfig.uart.WATER = 0;
+    // Last
+    _uartConfig.uart.CTRL  = LPUART_CTRL_RE(1U) + LPUART_CTRL_TE(1U);
+}
+
+bool Uart::isRxReady() const
```
christian-schilling: Could this be a non member static function taking a reference to `_uartConfig.uart.STAT`? If yes it should.

platforms/s32k1xx/bsp/bspUart/src/Uart.cpp
```diff
@@ -0,0 +1,98 @@
...
+        _uartConfig.uart.STAT = _uartConfig.uart.STAT | LPUART_STAT_OR_MASK;
+    }
+    return ((_uartConfig.uart.STAT & LPUART_STAT_RDRF_MASK) != 0);
+}
+
+size_t Uart::read(::etl::span<uint8_t> data)
+{
+    size_t bytes_read = 0;
+
+    if (data.size() == 0U)
```
christian-schilling: Why this check? Isn't the case already covered by the loop condition below?

platforms/s32k1xx/bsp/bspUart/src/Uart.cpp
```diff
@@ -0,0 +1,98 @@
...
+    {
+        return 0;
+    }
+
+    while (isRxReady() && (bytes_read < data.size()))
+    {
+        data[bytes_read] = _uartConfig.uart.DATA & 0xFFU;
```
christian-schilling: This is 3 dereferences in a loop. Should be `auto const DATA& = _uartConfig.uart.DATA;` before the loop.

libs/bsw/bsp/include/bsp/uart/UartConcept.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
+#include <bsp/uart/UartApi.h>
+#include <etl/type_traits.h>
+
+#include <concepts>
```
rolandreichweinbmw: We should prevent including mixed std:: and etl:: headers at the same time since this led to compile errors in the past when using the resulting combined headers.

etl/concepts.h is expected to be available soon, from https://github.com/ETLCPP/etl/pull/1213

So etl::same_as<> and etl::derived_from<> can be used then.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,80 @@
...
+{
+public:
+    /**
+     * sends out an array of bytes over the UART interface.
+     * the method will block until the data is sent.
+     * @param data - pointer to the data to be sent
+     *        length - the number of bytes to be sent
+     * @return the number of bytes written to the uart interface
+     */
+    size_t write(::etl::span<uint8_t const> const& data);
```
christian-schilling: Why does `write` take `span` by `const&` but `read` by value?

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
```diff
@@ -0,0 +1,36 @@
...
+    {(LPUART_BAUD_OSR(15)) + LPUART_BAUD_SBR(26)}, // = 115200 48MHz FIRC
+    {(LPUART_BAUD_OSR(9)) + LPUART_BAUD_SBR(8)}    // = 2MBit 80MHz PLL
+};
+
+Uart::UartConfig const config_uart[] = {
+    {
+        *LPUART1,
+        bios::Io::UART1_TX,
+        bios::Io::UART1_RX,
+        static_cast<uint8_t>(sizeof(baudRateConfig) / sizeof(UartBaudRate)),
```
christian-schilling: This looks like pointer + length as separate fields -> why not use `span`?

platforms/s32k1xx/bsp/bspUart/mock/include/UartMock.h
```diff
@@ -0,0 +1,22 @@
...
+
+namespace bsp
+{
+
+class UartMock : public UartApi
+{
+public:
+    MOCK_METHOD1(write, size_t(::etl::span<uint8_t const> const& data));
+    MOCK_METHOD1(read, size_t(::etl::span<uint8_t> data));
+    MOCK_METHOD0(init, void());
```
christian-schilling: `init` is not part of `UartApi`. Why is it in the mock? Probably shouldn't be.

platforms/s32k1xx/bsp/bspUart/test/mock/include/UartMock.h
aschaal: When you do it, you will find out that you have 2x namespace bsp....

doc/guidelines/conventions/interfaces.rst
cristiancsordas: done

doc/guidelines/conventions/interfaces.rst
cristiancsordas: all lines have less than 100 chars

platforms/s32k1xx/bsp/bspUart/include/bsp/UartConfig.h
cristiancsordas: the pointer was used as a pointer to an array. The size of the array was specified in the speedModes variable.
They are replaced now by only one field: uint32_t baudRate

platforms/s32k1xx/bsp/bspUart/src/Uart.cpp
cristiancsordas: isRxReady and isTxActive are now static functions with _uartConfig.uart.STAT as reference

platforms/s32k1xx/bsp/bspUart/src/Uart.cpp
cristiancsordas: the check is removed

platforms/s32k1xx/bsp/bspUart/src/Uart.cpp
cristiancsordas: done

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: read and write take now span by value

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
cristiancsordas: the pointer was removed, there is only the baudRate variable

platforms/s32k1xx/bsp/bspUart/mock/include/UartMock.h
cristiancsordas: init method removed from the mock

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,230 @@
...
+.. code-block:: cpp
+
+    // libs/bsw/bsp/include/bsp/example/ExampleApi.h
+
+    #pragma once
+
+    namespace bsp {
+    class ExampleApi {
+    public:
+        virtual ~ExampleApi() = default;
```
matthiaskessler: In our coding conventions we have rule CPP-010, Interfaces should start with `I`. 
We also have rule CPP-024, destructors of interfaces should be protected, non-virtual to prevent code calling delete through a base class pointer from compiling.

I am open to discussing these rules, but I think, they make sense.

doc/dev/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,230 @@
...
+
+    // libs/bsw/bsp/include/bsp/example/ExampleApi.h
+
+    #pragma once
+
+    namespace bsp {
+    class ExampleApi {
+    public:
+        virtual ~ExampleApi() = default;
+        virtual size_t read(::etl::span<uint8_t> data) = 0;
```
matthiaskessler, line 53: Just as an idea, since we will be using the type `::etl::span<uint8_t>` and `::etl::span<uint8_t const> const` a lot...what do you think about a reusable typedef `bsp::bytes` and (maybe) `bsp::const_bytes` ?

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,230 @@
...
+    public:
+        size_t read(::etl::span<uint8_t> data);
+        size_t write(::etl::span<uint8_t const> const& data);
+    };
+    BSP_EXAMPLE_CONCEPT_CHECKER(Example);
+    } // namespace bsp
+
+3. Static Interface Methods:
+
+- Define a class with static methods representing the interface.
```
matthiaskessler: I don't get this approach. Derived classes cannot "implement" a static function of a base class. They can add their own declaration/implementation of a static function but that's it. You'll e.g. not be able to access the implementation of a derived implementation through a base class pointer.
BTW, I personally prefer to call static functions via the scope resolution operator `::`  instead of the dot operator `.`.

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,230 @@
...
+
+    #include <bsp/ExampleConfig.h> // application specific includes
+
+    namespace bsp {
+    Example::ExampleConfig const config_example[] = {
+        // Configuration for Example instances
+    };
+
+    static Example instances[] = {
+        Example(config_example[0]),
```
matthiaskessler: I'd prefer to follow our coding guidelines also in examples and currently we are to use camelCase. Again, I am open for discussions but a fan of consistency ;-)

executables/referenceApp/application/src/app/app.cpp
```diff
@@ -193,8 +196,15 @@ IdleHandler idleHandler;
...
 
+void staticInit()
+{
+    Uart::getInstance(Uart::Id::TERMINAL).init();
+    Uart::getInstance(Uart::Id::TERMINAL).waitForTxReady();
+}
+
 void run()
 {
+    staticInit();
```
matthiaskessler, line 206: I think, we should discuss when and which `init` functions are bein called. My guess is, that a UART for serial output should be initialized as soon as possible. We are calling `boardInit()` from the startup code, but it's in the platform specific part of the application. Maybe we should add a generic callout there, too.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/include/bsp/uart/UartId.h
```diff
@@ -0,0 +1,22 @@
+// Copyright 2025 BMW AG
+
+#pragma once
+
+#include <cstdint>
+
+namespace bsp
+{
+
+class UartId
```
matthiaskessler: Maybe an `enum class UartId` is enough. I don't see the point in having a class that just provides an enum.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/stdIo/stdIo.cpp
```diff
@@ -1,15 +1,24 @@
...
+    static bsp::Uart& uart = bsp::Uart::getInstance(bsp::Uart::Id::TERMINAL);
+    uart.write(etl::span<uint8_t const>(&byte, 1U));
 }
 
 extern "C" int32_t getByteFromStdin()
 {
-    uint8_t const byte = static_cast<uint8_t>(SerialLogger_getc());
-    return (byte != 0U) ? byte : -1;
+    static bsp::Uart& uart = bsp::Uart::getInstance(bsp::Uart::Id::TERMINAL);
+    uint8_t data_byte      = 0;
```
matthiaskessler: `dataByte`

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/stdIo/stdIo.cpp
```diff
@@ -1,15 +1,24 @@
...
+    etl::span<uint8_t> data(&data_byte, 1U);
+    uart.read(data);
```
matthiaskessler, line 18: Many UART APIs that I looked at, provide a `simple` API for reading a single byte. While your approach is possible, my fear is, that it might actually generate inefficient code.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace bsp
+{
+
+uint32_t const baudRateConfig[] = {
+    (LPUART_BAUD_OSR(15)) + LPUART_BAUD_SBR(26), // = 115200 48MHz FIRC
+    (LPUART_BAUD_OSR(9)) + LPUART_BAUD_SBR(8)    // = 2MBit 80MHz PLL
+};
+
+Uart::UartConfig const config_uart[] = {
```
matthiaskessler: `configUart` or, if it's really in ROM because it's `const` then `CONFIG_UART`. If this becomes a more frequent pattern, we should agree on a naming convention.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/stdIo/stdIo.cpp
```diff
@@ -1,15 +1,24 @@
...
 
+#include "bsp/Uart.h"
 #include "charInputOutput/charIoSerial.h"
 #include "platform/estdint.h"
 
 extern "C" void putByteToStdout(uint8_t const byte)
 {
-    (void)SerialLogger_putc(static_cast<int32_t>(byte));
+    static bsp::Uart& uart = bsp::Uart::getInstance(bsp::Uart::Id::TERMINAL);
+    uart.write(etl::span<uint8_t const>(&byte, 1U));
```
matthiaskessler, line 10: Maybe we should have an overload for writing characters, I am not sure how good a compiler is in optimizing taking the address of an argument that is passed by value. My guess is, that this code is slower.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
```diff
@@ -0,0 +1,35 @@
...
+};
+
+static Uart instances[] = {
+    Uart(config_uart[static_cast<uint8_t>(Uart::Id::TERMINAL)]),
+};
+
+bsp::Uart& Uart::getInstance(Id id)
+{
+    static_assert(
+        static_cast<uint8_t>(Uart::Id::INVALID) >= static_cast<uint8_t>(etl::size(instances)),
```
matthiaskessler: I think, you added a constant `NUMBER_OF_UARTS` in some header file. Using that would be less confusing.

BTW, this API would allow passing `Uart::Id::INVALID` as `id` and would lead to accessing the `instances` array out of bounds.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/include/bsp/uart/UartId.h
```diff
@@ -0,0 +1,22 @@
...
+{
+public:
+    enum class Id
+    {
+        TERMINAL,
+        INVALID,
+    };
+};
+
+static constexpr uint8_t NUMBER_OF_UARTS = static_cast<uint8_t>(UartId::Id::INVALID);
```
matthiaskessler: I'd prefer `size_t` as type for NUMBER_OF_... constants. This will likely lead to more efficient code because it's a type that uses typically a full register.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/include/bsp/uart/UartId.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include <cstdint>
+
+namespace bsp
+{
+
+class UartId
+{
+public:
+    enum class Id
```
matthiaskessler: If you `static_cast` those constants to `uint8_t` later, wouldn't it make sense to specify it here? 
But it really depends on how you are planning to uses the constants later. My experience is, that artificially limiting the type to sth. less than the register width of the architecture can lead to inefficient assembler code being generated by the compiler.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
```diff
@@ -0,0 +1,35 @@
...
+        "Not enough Uart instances defined");
```
matthiaskessler, line 35: Reading the error message lets me think that this

`static_assert(::etl::size(instances) == ::bsp::NUMBER_OF_UARTS, ...)`

would make it easier to read. Or maybe `::etl::size(instances) >=` if you want to allow more configs then Ids.

executables/referenceApp/platforms/s32k148evb/main/src/lifecycle/StaticBsp.cpp
```diff
@@ -37,7 +38,7 @@ StaticBsp::StaticBsp()
 , _requestUpdateGateRegisters(false)
 , _mode(_INIT_)
 {
-    sciInit();
+    Uart::getInstance(Uart::Id::TERMINAL).init();
```
matthiaskessler: Doesn't this overlap with the call from `staticInit()` in `app.cpp`?

libs/bsw/bsp/include/bsp/uart/IUartApi.h
```diff
@@ -0,0 +1,36 @@
...
+#include <etl/span.h>
+
+#include <cstddef>
+#include <cstdint>
+
+namespace bsp
+{
+
+/**
+ * Base class used to define the generic interface for the uart communication
```
matthiaskessler: Base class defining the generic interface for UART communication.

I think, we have the convention to write full sentences, including a full stop. Also UART is a well defined abbreviation, I personally would write it in all CAPS.

libs/bsw/bsp/include/bsp/uart/IUartApi.h
```diff
@@ -0,0 +1,36 @@
...
+ */
+class UartApi
+{
+public:
+    /**
+     * sends out an array of bytes
```
matthiaskessler: I'd prefer "Sends out a span of bytes." or even "Sends out a number of bytes.". Using array suggests a different type and many other types are convertible to `span`.

libs/bsw/bsp/include/bsp/uart/IUartApi.h
```diff
@@ -0,0 +1,36 @@
...
+     * @param data - pointer to the data to be send
```
matthiaskessler: In most cases, we are using `\param`. I am open to discussing that, but we should aim for a consistent style. In the past, we had issues with some static code analysis tools and the character `@`, that's why we switched to using `\`.

libs/bsw/bsp/include/bsp/uart/IUartApi.h
```diff
@@ -0,0 +1,36 @@
...
+     *        length - the number of bytes to be send
```
matthiaskessler: The params, are IMHO wrong. We are passing a span, where length is a member, that's the beaty and whole point in using `span` over pointer + length.

libs/bsw/bsp/include/bsp/uart/IUartApi.h
```diff
@@ -0,0 +1,36 @@
...
+     * @return the number of bytes written
+     */
+    size_t write(::etl::span<uint8_t const> const data);
+
+    /**
+     * reads an array of bytes
```
matthiaskessler: See above.

libs/bsw/bsp/include/bsp/uart/UartConcept.h
matthiaskessler: Will this include even compile, if using older compiler versions?

libs/bsw/bsp/include/bsp/uart/UartApi.h
```diff
@@ -0,0 +1,36 @@
...
+#include <cstddef>
+#include <cstdint>
+
+namespace bsp
+{
+
+/**
+ * Base class used to define the generic interface for the uart communication
+ */
+class UartApi
```
matthiaskessler: I am missing an API to write and read single bytes. I am afraid, that just providing the `span` API might lead to inefficient code. Also, many APIs that I looked at are providing sth. like `available()` so that you can check if it's worth calling `read()`.

If we are defining a new API, then we should also think about the API for a possible asynchronous implementation of writing a bigger span of data using DMA.

libs/bsw/bsp/include/bsp/uart/UartConcept.h
```diff
@@ -0,0 +1,45 @@
...
+        bsp::UartCheckInterface<_class>, \
+        "Class " #_class " does not implement UartApi interface correctly");
+
+#else
+#define BSP_UART_CONCEPT_CHECKER(_class)
+#endif
+
+} // namespace bsp
+
+/**
```
matthiaskessler: I think, this comment is misplaced.

executables/unitTest/bsp/bspUart/include/bsp/Uart.h
matthiaskessler, line 1: Since some other UART related files are in `bsp/uart/...` maybe we should add this folder here as well?

executables/unitTest/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * This class implements the UART communication for the unit tests.
+ */
+class Uart
+: public bsp::UartApi
+, public bsp::UartId
+{
+public:
+    /**
+     * sends out an array of bytes .
```
matthiaskessler: I'd suggest to use a consistent brief style throughout the comments. I know we are not very consistent in the entire code base yet, but we should aim for it ;-) 
My suggestion is to write a whole sentence, i.e. Start with a capital letter and end with a ., what do you think about that?

libs/bsw/bsp/include/bsp/uart/IUartApi.h
```diff
@@ -0,0 +1,36 @@
+// Copyright 2025 BMW AG
+
+#pragma once
+
+#include <etl/span.h>
+
+#include <cstddef>
+#include <cstdint>
+
+namespace bsp
```
matthiaskessler: What do you think about adding a nested namespace `uart`? It would replicate the file path and also structure the things we are planning to add to the `bsp` namespace.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
matthiaskessler, line 1: Maybe it's worth adding another layer of path and namespace here. The file `UartApi` is placed in a folder `bsp/uart`, which I like. I'd suggest to use this consistently and also add a namespace `uart` inside `bsp`.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,80 @@
...
+
+#include <bsp/uart/UartApi.h>
+#include <bsp/uart/UartConcept.h>
+#include <bsp/uart/UartId.h>
+
+namespace bsp
+{
+/**
+ * This class implements the UART communication for S32K1xx platforms.
+ * It inherits from the base class UartApi and implements the methods for writing and reading data
```
matthiaskessler: nit: I think, we should limit comment to meaningful information. It's obvious that it inherits from `UartApi` and implements that API.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,80 @@
...
+ * over the UART interface. It also provides a method for initializing the UART communication.
```
matthiaskessler: This looks like maybe Copilot created the last sentence ;-). It doesn't really add information but is actually incomplete because next to implementing UART communication it provides more that a method for initializing the UART communication.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,80 @@
...
+ */
+class Uart
+: public bsp::UartApi
+, public bsp::UartId
```
matthiaskessler: Can you explain, why we should inherit from `bsp::UartId`? I don't get the point of it. It just defines an enum, which we can also use by accessing it with fully qualified names...

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,80 @@
...
+{
+public:
```
matthiaskessler, line 13: Again, I suggest to use correct grammar and spelling. This comment has two sentences that end with a full stop. They should start with a capital letter.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * sends out an array of bytes over the UART interface.
+     * the method will block until the data is sent.
+     * @param data - pointer to the data to be sent
```
matthiaskessler: Again, using `span` is way better and we don't need to mention pointer and length.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * checks if the UART is initialized
+     * @return true if initialized, false otherwise
+     */
+    bool isInitialized() const;
+
+    /**
+     * waits until the UART is ready to transmit data
+     * @return true if ready within timeout, false otherwise
```
matthiaskessler: If you mention a timeout here, I suggest you let the caller specify it in a reasonable unit, maybe microseconds.

executables/unitTest/bsp/bspUart/doc/index.rst
matthiaskessler, line 1: Just a question: Why do we need the UT implementation at all, if we also have a Mock?

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,80 @@
...
+     */
+    bool waitForTxReady();
+
+    /**
+     * factory method which instantiates and configures an UART object.
```
matthiaskessler: AFAIK, this is not true. The instances are already created and `getInstance()` merely returns a reference to it.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
```diff
@@ -0,0 +1,35 @@
...
+    {
+        *LPUART1,
+        bios::Io::UART1_TX,
+        bios::Io::UART1_RX,
+        baudRateConfig[0],
+    },
+};
+
+static Uart instances[] = {
+    Uart(config_uart[static_cast<uint8_t>(Uart::Id::TERMINAL)]),
```
matthiaskessler: Depending on what the constructor of `Uart` does, we should discuss if it's a good idea to instantiate them here in the global scope. We don't have full control, when the constructor is called and in ECUs that support sleep and sleep RAM, this might lead to uninitialized objects after wakeup.

Maybe we should postpone the discussion about sleep/wakeup but we should discuss if we want full control, of when the `Uart` instances are constructed.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,80 @@
...
+    size_t read(::etl::span<uint8_t> data);
+
+    /**
+     * configures and starts the UART communication
+     * this method must be called before using the read/write methods
+     */
+    void init();
+
+    /**
+     * checks if the UART is initialized
```
matthiaskessler: IMHO, "Returns if this Uart instance is initialized or not." would be a perfectly valid one line brief. I am not a fan of having simple methods and then being forced to repeat the brief description in the `\return` comment because some rule requires me to have a `\return` comment. I think we should agree, that this is not always necessary, what do you think?

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
```diff
@@ -0,0 +1,80 @@
...
+
+    Uart(UartConfig const& uartConfig) : _uartConfig(uartConfig) {}
+
+private:
+    /**
+     * writes one byte of data to the UART interface.
+     * @param data - the byte to be written
+     * @return true if the byte was written successfully, false otherwise
+     */
+    bool writeByte(uint8_t data);
```
matthiaskessler, line 66: I think, a simple API to read and write bytes should even be part of the public API.

platforms/s32k1xx/bsp/bspUart/include/bsp/UartParams.h
```diff
@@ -0,0 +1,20 @@
...
+#include <io/Io.h>
+#include <mcu/mcu.h>
+
+namespace bsp
+{
+
+struct Uart::UartConfig
+{
+    LPUART_Type& uart;
+    bios::Io::PinId txPin;
```
matthiaskessler: Maybe these member can even be `const`? It's just an idea...

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,230 @@
...
+.. code-block:: cpp
+
+    // platforms/s32k1xx/bsp/bspExample/include/bsp/Example.h
+
+    #pragma once
+
+    #include <bsp/ExampleApi.h>
+    #include <bsp/ExampleConcept.h>
+    namespace bsp {
+    class Example : public ExampleApi {
```
matthiaskessler: I don't see how this is really an interface. If you inherit the declaration of non-virtual functions and redeclare/implement them in a derived class, you are hiding the names of the base class and you'll not be able to access the implementations through a pointer to the base class.
The concepts example seems only useful if you use templated code, and I am not sure, if all UART using code should be templated, or am I missing sth.?

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,230 @@
...
+    #include <bsp/ExampleConfig.h> // application specific includes
+
+    namespace bsp {
+    Example::ExampleConfig const config_example[] = {
+        // Configuration for Example instances
+    };
+
+    static Example instances[] = {
+        Example(config_example[0]),
+        Example(config_example[1]),
```
matthiaskessler: I commented on the issues with this form of instantiation further down.

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,230 @@
...
+  but for performance reasons, prefer static interfaces with concept checks where applicable.
+- The low level interfaces should be placed in the bsp namespace.
+- Isolate the configuration and platform-specific details from the interface definition
+  to ensure portability across different platforms.
+- Configure the concrete implementations of the interfaces in separate configuration
+  files or classes, allowing for easy adaptation to different hardware or software environments.
+
+Approaches to Interface Definition
+----------------------------------
+There are three possible approaches to define interfaces:
```
matthiaskessler: After reading your examples I tend to disagree. I will provide an still incomplete list of what I think the different approaches are as a basis for a further discussion:

1. The approach with using abstract base classes is the "classical" one and will work. Advantages are, that applications can be written against the interface without the need of having any implementation in the include path and it's easy to provide multiple implementations for different platforms. Disadvantage is the runtime overhead. Similar effects can be achieved by using function pointers or `etl::delegate` instead of virtual functions but that might have even worse runtime.
2. Another approach could be to put the driver interface header files with type and function declarations in a common location and the let the platforms provide the implementation in a separate compilation unit. This works only if the whole project ever contains only one possible implementation of the interface. Advantage: no virtual functions, still applications can be programmed against declaration without the need to include the implementation. Disadvantage: doesn't scale well.
3. Following a templated approach using CRTP or concepts. Disadvantage: All implementations will become templated, which means that applications using the interface will have the drivers in the include path. Advantage: non virtual functions.

doc/dev/guidelines/conventions/interfaces.rst
matthiaskessler: Please see my below comments and also the comment from @rolandreichweinbmw . Before spending time on documenting too much, I think, we need to discuss about the general feasibility of the approaches.

doc/guidelines/conventions/interfaces.rst
cristiancsordas: done

platforms/s32k1xx/bsp/bspUart/test/mock/include/UartMock.h
cristiancsordas: UartMock is instantiated inside of BspTest. Fixed the 2x namespace bsp

platforms/s32k1xx/bsp/bspUart/test/CMakeLists.txt
cristiancsordas: done

doc/guidelines/conventions/interfaces.rst
cristiancsordas: I changed the file name from UartApi.h into IUartApi.h. I changed also the class name from UartApi->IUartApi.
I adapted in the documentation the example for the interface with pure virtual functions: the destructor is protected and non-virtual

doc/guidelines/conventions/interfaces.rst
cristiancsordas: The idea behind this approach is to enforce (and also document) the interface. This is similar to the approach number 3, but for classes with static methods. The base class will never be available through a pointer. 
The enforcement can be done where available through the C++20 concepts.

doc/guidelines/conventions/interfaces.rst
cristiancsordas: I renamed the config_example into configExample. The same, the config_uart in the UartConfig.cpp was renamed into ConfigUart

executables/referenceApp/application/src/app/app.cpp
cristiancsordas: I think that boardInit() does a first initialization of the board, mostly related to clocks, before MPU is started. The most of the peripherals are initialized inside of StaticBsp::hwInit(). 
Do you want that I add the Uart initialization in the StaticBsp::hwInit() method? Or should be a new function which is not part of StaticBsp? If yes, where should be located this function?

A small remark. Before, the Uart had a sort of lazy initialization. The first call SerialLogger_putc initialized it and waited until the Tx transmission was ready. This had the advantage that it blocked (waiting that Tx is available) only the task from which SerialLogger_putc was called. But veryfing all the time if the Uart is initialized is not performant.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/include/bsp/uart/UartId.h
cristiancsordas: The Uart class inherits from the UartId class. The inheritance from an enum class is not supported in C++.
Having the UartId (which is located in the bspConfiguration) separated from Uart could allow different applications to use different configurations. 
And I think that here inheritance is better than composition because it allow this call: Uart::getInstance(Uart::Id::TERMINAL)

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/stdIo/stdIo.cpp
cristiancsordas: done

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/stdIo/stdIo.cpp
cristiancsordas: Some collegues requested an interface which can read and write in one call multiple bytes. 
For an Uart implementation based on interrupts it can be helpfull sometimes to read more than 1 byte to reduce the CPU load and to prevent the internal buffer overflow.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
cristiancsordas: done, renamed into CONFIG_UART

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
cristiancsordas: I replaced in the static_assert the Uart::Id::INVALID with NUMBER_OF_UARTS. 
I added a runtime assert for the case that Id is INVALID

executables/referenceApp/platforms/s32k148evb/bspConfiguration/include/bsp/uart/UartId.h
cristiancsordas: changed the uint8_t into size_t. Also for the cast operations used inside of Uart::getInstance

executables/referenceApp/platforms/s32k148evb/bspConfiguration/include/bsp/uart/UartId.h
cristiancsordas: I added a type specifier for the enum class : enum class Id : size_t. The casts are also done using size_t. 
The elements of this type should have now the size of the register of the chosen architecture.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
cristiancsordas: changed to static_assert(::etl::size(instances) == ::bsp::NUMBER_OF_UARTS, ...)

executables/referenceApp/platforms/s32k148evb/main/src/lifecycle/StaticBsp.cpp
cristiancsordas: I removed this call from StaticBsp.cpp

libs/bsw/bsp/include/bsp/uart/IUartApi.h
cristiancsordas: changed to "Base class defining the generic interface for UART communication."

libs/bsw/bsp/include/bsp/uart/IUartApi.h
cristiancsordas: changed to "sends out a number of bytes"

libs/bsw/bsp/include/bsp/uart/IUartApi.h
cristiancsordas: changed in all the added files @param into \param and @return into \return

libs/bsw/bsp/include/bsp/uart/IUartApi.h
cristiancsordas: removed the description regarding length

libs/bsw/bsp/include/bsp/uart/IUartApi.h
cristiancsordas: updated: "reads a number of bytes"

libs/bsw/bsp/include/bsp/uart/UartConcept.h
cristiancsordas: I moved the #include <concepts> inside of the block defined by the macro __cpp_concepts. 
I tested with gcc, the versions C++20 and C++14

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,231 @@
...
+
+    // libs/bsw/bsp/include/bsp/example/IExampleApi.h
+
+    #pragma once
+
+    namespace bsp {
+    class IExampleApi {
+    public:
+        virtual size_t read(::etl::span<uint8_t> data) = 0;
+        virtual size_t write(::etl::span<uint8_t const> const& data) = 0;
```
matthiaskessler: I saw, that in the "real" code, you already unified the code to passing both `span`s in `read` and `write` by value.

doc/guidelines/conventions/interfaces.rst
```diff
@@ -0,0 +1,231 @@
...
+
+.. code-block:: cpp
+
+    // libs/bsw/bsp/include/bsp/example/IExampleApi.h
+
+    #pragma once
+
+    namespace bsp {
+
+    class IExampleApi {
```
matthiaskessler: Maybe we need to discuss this again. Currently I don't consider this class an interface, which is, why I wouldn't prefix it with `I`. I do not understand how inheriting from this class (as done further down: `class Example : public IExampleApi` helps enforcing anything. I think, we shouldn't mix approaches. If we want to use concepts, fine, but in this case, we can just leave the inheriting from a base class which contains non-virtual function declarations away.

libs/bsw/bsp/include/bsp/uart/UartConcept.h
cristiancsordas: comment moved before the "concept UartConcept"

executables/unitTest/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: the interfaces are located under libs/bsw/bsp/include/bsp and they use all the same "rule", after bsp is the folder of the module with the corresponding header files: bsp/eeprom, bsp/memory, ...
The implementation for s32k is located under platforms/s32k1xx/bsp and they follow this "rule": after bsp is the folder of the module, but in the include and src there is a subfolder describing again the module. Something like this: 
- for io bsp/bspIo/include/io and bsp/bspIo/src/io
- for can bsp/bspFlexCan/include/can and bsp/bspFlexCan/src/can

If I follow this rule I should change in the posix, s32k and unit tests implementations the structure like this:
bsp/bspUart/include/uart, bsp/bspUart/src/uart.

Is this correct?

executables/referenceApp/application/src/app/app.cpp
matthiaskessler: We'll separate the discussion of startup-code and initialization from this PR. I created [an issue](https://github.com/eclipse-openbsw/openbsw/issues/305)

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/stdIo/stdIo.cpp
matthiaskessler: Can be done if necessary.

libs/bsw/bsp/include/bsp/uart/UartConcept.h
```diff
@@ -0,0 +1,46 @@
...
+#include <cstddef>
+#include <cstdint>
+
+namespace bsp
+{
+
+#if __cpp_concepts
+
+#include <etl/type_traits.h>
+#include <concepts>
```
rolandreichweinbmw, line 10: Please include outside namespace bsp.

cristiancsordas: done

executables/unitTest/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: I changed the descriptions so that they start with capital letter and end with a .

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: I agree with this point. But, I suggest to have a separate discussion with other people involved. If the decision is for having multiple namespaces (like uart) inside the bsp namespace then we can do it in a separate PR.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: I tried to remove some unnecessary comments.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: I removed some unnecessary comments

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: changed, instead of inheritance, the forward declaration of "enum class Id" is used now.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: done

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: done, span is mentioned instead of pointer and length

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: I simplified the comment. It is difficult to get the real time-out value in microseconds based on the  while/loop iterations

executables/unitTest/bsp/bspUart/doc/index.rst
cristiancsordas: I was not able to compile the unitTest/bspConfiguration without providing a Uart.h file

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: done, the description was updated.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
cristiancsordas: Agree, a discussion is necessary here. Maybe we should also define cleary the state of an "hardware" object (like Uart) : instantiated and initialized. And where these states are handled after startup and wakeup.

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: I agree, we whould aim for simplicity. In this case here, the comments are based on the Doxygen rules, which imply that /return is commented for a method which returns a value. 
But, if we agree that these rules (like Doxygen) are not necessary then we can simplify the comments.
I removed the /return comment

platforms/s32k1xx/bsp/bspUart/include/bsp/Uart.h
cristiancsordas: I suggest to postpone this and have a  general discussion regarding BSP synchron and asynchron interfaces. 
After, we can adapt this interface in a next PR.

platforms/s32k1xx/bsp/bspUart/include/bsp/UartParams.h
cristiancsordas: done, txPin, rxPin and baudRate are const

doc/guidelines/conventions/interfaces.rst
cristiancsordas: done. As discussed, the inheritance is used only for the interfaces with pure virtual methods

doc/guidelines/conventions/interfaces.rst
cristiancsordas: done. Documentation updated, we will have only the 2 cases: normal inheritance (based on virtual methods), no inheritance for BSP classes where performance is important (checks done with C++20 concepts)

executables/unitTest/bsp/bspUart/include/bsp/Uart.h
matthiaskessler: I think so. Let's discuss that outside of this PR.

doc/guidelines/conventions/interfaces.rst
cristiancsordas: done. The documentation was updated

doc/guidelines/conventions/interfaces.rst
cristiancsordas: I removed the IExampleApi from the documentation. Please check the latest update. There is only the IExample used as an interface with virtual methods

executables/referenceApp/platforms/posix/bspConfiguration/include/bsp/uart/UartConfig.h
```diff
@@ -0,0 +1,20 @@
...
+namespace bsp
+{
+
+enum class Uart::Id : size_t
+{
+    TERMINAL,
+    INVALID,
+};
+
+static constexpr uint8_t NUMBER_OF_UARTS = static_cast<uint8_t>(Uart::Id::INVALID);
```
rolandreichweinbmw: size_t?

executables/referenceApp/platforms/s32k148evb/bspConfiguration/include/bsp/uart/UartConfig.h
```diff
@@ -0,0 +1,17 @@
...
+namespace bsp
+{
+
+enum class Uart::Id : size_t
+{
+    TERMINAL,
+    INVALID,
+};
+
+static constexpr uint8_t NUMBER_OF_UARTS = static_cast<uint8_t>(Uart::Id::INVALID);
```
rolandreichweinbmw: size_t?

executables/unitTest/bsp/bspConfiguration/include/bsp/uart/UartConfig.h
```diff
@@ -0,0 +1,20 @@
...
+namespace bsp
+{
+
+enum class Uart::Id : size_t
+{
+    TERMINAL,
+    INVALID,
+};
+
+static constexpr uint8_t NUMBER_OF_UARTS = static_cast<uint8_t>(Uart::Id::INVALID);
```
rolandreichweinbmw: size_t?

PR #172, opened 2025-09-02
Add UDP multicast support

christian-schilling:
Adds missing CRC based filter bit calculation to ENET driver and multicast join to DemoSystem.


matthiaskessler (commented):
👍

PR #177, opened 2025-09-09
Add Valgrind support and integrate memory checks into CI

marcmo:
originates from CR #134850


shamitha-shashidhara:
I've put together a brief summary based on my findings about sanitizers and Valgrind. Here's a beginner’s perspective:

- **Valgrind**

**Memory leaks:** Excellent (Memcheck very reliable)
  
**Thread issues:** Limited detection
  
**Use-after-free:** Detected
  
**Uninitialized memory:** Detected reliably (Memcheck)
  
**Recompilation:** Not required — works with existing binaries
  
**Ease of use:** Can be installed via package manager and run directly with:
  
 ctest -T memcheck
  
**Performance:** Very slow (tests can run 10–15 minutes or more, depending on machine speed)
  
**Best for:**
  
 Deep memory analysis
  
 Debugging legacy binaries (no need to rebuild)
  
 Situations where performance isn’t critical


- **Sanitizers (ASan, TSan, MSan, UBSan)**

**Memory leaks:** Good (AddressSanitizer catches many, but not all leaks)

**Thread issues:** Very strong (ThreadSanitizer is excellent)

**Use-after-free:** Detected

**Uninitialized memory:** Partially detected (MemorySanitizer)

**Recompilation: Required** — must

**Ease of use:** add -fsanitize=address,undefined to Cmakelists.txt

**Performance:** Faster than Valgrind, but still take longer time.

**Best for:**

Faster feedback during development

Detecting multithreading issues

When you control the build and can recompile

**Sanitizer Type**

address (ASan) - Detects memory corruption like buffer overflows and use-after-free

leak - Detects memory leaks

undefined (UBSan) - Detects undefined behavior like integer overflows

thread (TSan) - Detects data races in multithreaded programs

 **MSan (MemorySanitizer) Not Supported in GCC**

MSan is not available in GCC. The -fsanitize=memory flag is not recognized by GCC.

MSan is only supported in Clang.


 What happens if we use both of them at same time?

If you build with Sanitizers enabled in CMake (e.g. -fsanitize=address): That binary is already instrumented. Running Valgrind memcheck on it usually causes conflicts (both tools hook into memory management). You’ll get crashes, “false positives,” or meaningless output like the one you pasted earlier. 
Workaround:
We should add an option in CMake to enable or disable sanitizers. This way, developers who want to use sanitizers can enable them, while those who prefer using Valgrind can disable sanitizers and use Valgrind explicitly.

PR #182, opened 2025-09-15
Add some Documentation about the internal Implementation of Async

christophruethingbmw:


rolandreichweinbmw:
I guess the build error for Puncover Report results from a broken build script, since the PR doesn't change any source code.

Anyway, I wonder why we need -specs=nano.specs when we have -specs=nosys.specs at the same time?

(See admin/cmake/ArmNoneEabi-gcc.cmake)

Issue #176, opened 2025-09-04
Consolidation of assert from ESTD to ETL


rolandreichweinbmw:
Hi!

One leftover of the switch from ESTD to ETL is estd_assert() which by default is doing:
* Stacktrace if available on __linux__ / __APPLE__, otherwise printing plain source location, and abort()
* abort() otherwise

While we can keep this default behaviour, I propose to switch the estd_assert interface to using ETL instead for consistency (ETL_ASSERT).

The interfaces are quite close to each other. At build time, we can choose the strategy via compiler symbols:

ETL: etl/error_handler.h: ETL_ASSERT()
- ETL_NO_CHECKS: No checks (default)
- ETL_USE_ASSERT_FUNCTION: Call handler function from macro
- ETL_USING_EXCEPTIONS: Throw
- ETL_USING_EXCEPTIONS+ETL_LOG_ERRORS: Call handler function and throw
- ETL_IS_DEBUG_BUILD: OS assert()

ESTD: util/estd/assert.h: estd_assert()
- ESTL_NO_ASSERT: No checks
- ESTL_HARD_ASSERT_FUNC: to be defined
- ESTL_ASSERT_MESSAGE_ALL: Call handler function via macro (default)
- ESTL_ASSERT_FUNCTION: Call handler function via function

Before providing a PR for this, I would like to clarify if there are any requirements to keep the:
* ESTL_HARD_ASSERT_FUNC case, i.e. is it a use case that can't be solved with configuring a classic handler function as in the ETL_USE_ASSERT_FUNCTION or ESTL_ASSERT_MESSAGE_ALL cases?
*  ESTL_ASSERT_FUNCTION case where the estd::estd_assert_no_macro is called unconditionally and needs to decide by itself how to handle both true and false cases?

If there are no real requirements for those cases, I would map all those ESTL cases to the respective ETL implementations.

What do you think?

PR #184, opened 2025-09-16
Remove ESTD headers and use ETL instead

rolandreichweinbmw:
With the ethernet feature in OpenBSW, the big_endian.h and memory.h headers from ESTD were re-introduced. This PR fixes this by removing them again and using ETL instead.

The remaining assert.h is kept intentionally to be discussed in the separate issue https://github.com/eclipse-openbsw/openbsw/issues/176


executables/referenceApp/application/src/systems/EthernetSystem.cpp
```diff
@@ -5,11 +5,12 @@
 #include <bsp/SystemTime.h>
 #include <ethernet/EthernetLogger.h>
 #include <etl/functional.h>
+#include <etl/memory.h>
```
ghost: Is this needed here?

executables/referenceApp/application/src/systems/EthernetSystem.cpp
```diff
@@ -5,11 +5,12 @@
...
 #include <lwip/init.h>
 #include <lwip/prot/ethernet.h>
 #include <lwip/timeouts.h>
 #include <lwipSocket/utils/LwipHelper.h>
-#include <util/estd/big_endian.h>
+#include <util/estd/assert.h>
```
ghost, line 11: Is this needed here?

executables/referenceApp/application/src/systems/EthernetSystem.cpp
rolandreichweinbmw: No. Thanks for the hint. Will adjust.

executables/referenceApp/application/src/systems/EthernetSystem.cpp
rolandreichweinbmw: Yes. estd_assert was previously provided by an indirectly included header, just by accident. This include is removed now.

lmgivelaccenture (approved):

PR #185, opened 2025-09-16
Fix building unit tests with clang

rolandreichweinbmw:
CodeCoverage.cmake and cmake seem to interfere by cmake (wrongly) setting CMAKE_ASM_COMPILER_VERSION to an empty string, and CodeCoverage.cmake evaluating it.

Since cmake is at fault here, we fix it by carefully setting CMAKE_ASM_COMPILER_VERSION to CMAKE_C_COMPILER_VERSION in this case.


johannes-esr:
@rolandreichweinbmw, should we postpone this change until #186 is merged? If so, can you change the status to draft?

rolandreichweinbmw:
Yes, of course!

vladyslavmarkovaccenture:
Unit tests are built with clang after #186

rolandreichweinbmw:
Thanks, works now!

PR #186, opened 2025-09-16
Make openbsw usable via add_subdirectory

marcmo:
originates from CR #133842


.ci/build_matrix.py
```diff
@@ -1,19 +1,317 @@
...
+            return env
+
+    if platform == Platform.x86:
+        if cxxid == CxxId.GNU:
+            env["CC"] = get_full_path("gcc-11")
+            env["CXX"] = get_full_path("g++-11")
+            return env
+        if cxxid == CxxId.Clang:
+            env["CC"] = get_full_path("clang")
+            env["CXX"] = get_full_path("clang++")
```
rolandreichweinbmw: Is there a reason behind the asymmetry between versioned GCC but unversioned clang? I guess because of the concrete GCC and Clang versions in Ubuntu 24.04? Doesn't the default GCC from there work?

.ci/build_matrix.py
```diff
@@ -1,19 +1,317 @@
...
+        "build": None,
+        "test": None,
+        "cxxid": CxxId.Both,
+        "platform": Platform.x86,
+        "build-config": Config.NoConfig,
+    },
+    # workflow for s32k148 built with GNU compiler
+    # cmake --workflow --preset ci-s32k148-gcc
+    {
+        "config": ["cmake", "--workflow", "--preset", "ci-s32k148-gcc"],
```
rolandreichweinbmw: Is it necessary to explicitly distinguish between CI and Non-CI versions of the commands? For reproducing locally, the CI versions will possibly need to be run locally anyway, and CI should also make sure that the locally executed commands work correctly?

CMakePresets.json
```diff
@@ -0,0 +1,277 @@
...
+            "configuration": "RelWithDebInfo"
+        },
+        {
+            "name": "s32k148-clang",
+            "displayName": "S32K148 build (Clang)",
+            "description": "Build reference application for S32K148 platform with Clang (Release by default; for Debug use --config Debug)",
+            "configurePreset": "s32k148-clang"
+        },
+        {
+            "name": "_s32k148-clang_deb",
```
rolandreichweinbmw: For consistency, maybe better call it *_dbg (already being used below) or *_debug

deb is also overloaded with Debian package.

CMakePresets.json
```diff
@@ -0,0 +1,277 @@
...
+            "inherits": "s32k148-clang",
+            "configuration": "Debug"
+        },
+        {
+            "name": "_s32k148-clang_rel",
```
rolandreichweinbmw: Maybe better call it *_release instead of *_rel since rel could  confusingly suggest PIC?

NOTICE.md
```diff
@@ -43,7 +43,7 @@ We recommend to read their licenses, as their terms may differ from the terms de
 | FreeRTOS Real Time Kernel | 10.6.2  | MIT       | ``libs/os/3rdparty/freeRtos/LICENSE.md``                             |
 | CMSIS                     | 6.1.0   | Apache v2 | ``platforms/s32k1xx/bsp/bspMcu/include/3rdparty/cmsis/LICENSE``      |
 | NXP S32K148 Headers       | 1.1a    | BSD-3     | ``platforms/s32k1xx/bsp/bspMcu/include/3rdparty/nxp``                |
-| CodeCoverage              |         | BSD-3     | ``admin/cmake/CodeCoverage.cmake``                                   |
+| CodeCoverage              |         | BSD-3     | ``cmake/toolchains/CodeCoverage.cmake``                                   |
```
rolandreichweinbmw, line 46: Table format?

libs/bsw/docan/include/docan/datalink/DoCanFrameCodec.h
```diff
@@ -35,6 +36,9 @@ enum class CodecResult : uint8_t
     INVALID_FRAME_TYPE
 };
 
+static constexpr uint8_t EXTENDED_SF_DL_EDGE_SIZE  = 8U;
+static constexpr uint16_t ESCAPED_SEQ_MESSAGE_SIZE = 4095U;
```
rolandreichweinbmw, line 40: Is it necessary to narrow this down to below size_t? uint8_t and uint16_t typically need more (.text) ROM and runtime cycles due to additional conversion operations. At least on common embedded architectures like ARM.

libs/bsw/uds/include/uds/connection/IncomingDiagConnection.h
```diff
@@ -10,7 +10,14 @@
 #include "uds/connection/ErrorCode.h"
 #include "uds/connection/PositiveResponse.h"
 
+// clang-format off
+
+// compilation on clang won't work without tuple_size
+#include <tuple>
```
rolandreichweinbmw: Let's solve this centrally in the etl_profile.h to make it consistent for all ETL usage.

I'm actually commonly doing it this way in my projects.

NOTICE.md
vladislavmarkov: I didn't get the question. But good thing you highlighted this, because CodeCoverage.cmake should be in cmake/modules, and not in toolchains.

libs/bsw/uds/include/uds/connection/IncomingDiagConnection.h
vladislavmarkov: I tried doing that with etl_profile first, but to no avail. How does your etl_profile.h looks like @rolandreichweinbmw ?

CMakePresets.json
vladislavmarkov: naming is not that important in this context, since those were supposed to be "private" presets not meant to be used directly. I might as well just rename them to "{...}_r" and "{...}_d" :slightly_smiling_face:

.ci/build_matrix.py
vladislavmarkov: Good question - I think it makes sense to set fixed version, since it shouldn't matter which linux distro is used as a host OS.

.ci/build_matrix.py
vladislavmarkov: No. I did this solely for the demo purposes. Demo is done, and I don't think I'll use only "ci" version on CI either. I'll rename it once I get to fixing CI. I'll rename it to "wf-...".

libs/bsw/docan/include/docan/datalink/DoCanFrameCodec.h
vladislavmarkov: Didn't know that. If it's true, then it should stay as size_t. I will double check that.

.ci/build_matrix.py
rolandreichweinbmw: OTOH most distros and their versions contain a default gcc and clang, but not all contain e.g. a gcc-11.

NOTICE.md
rolandreichweinbmw: I meant the whitespace alignment within the table cell (minor).

.ci/build_matrix.py
vladislavmarkov: If we leave it to default, then we need to make sure it works for all past and future distros. We do, however, test the build using multiple versions of compilers to make sure we don't have errors on some specific version (which was the case sometimes). But I'd vote for a fixed mainline compiler versions for now. That could rule out some misunderstandings.

rolandreichweinbmw: Then maybe do similarly for both gcc and clang and not fixed for one and implicit for the other?

libs/bsw/uds/include/uds/connection/IncomingDiagConnection.h
rolandreichweinbmw: In another project I'm successfully using:
```
#pragma once

// Define tuple_size from clang headers consistently since
// ETL uses them for deduction guide
#include <tuple>

#include "etl/profiles/cpp17_no_stl.h"

// ETL_COMPILER_* will be autodetected correctly only if
// ETH_COMPILER_GENERIC (from default profile above) is not defined
#undef ETL_COMPILER_GENERIC

// ETL's initializer_list.h would otherwise define one in conflict with STL's
#define ETL_NO_INITIALIZER_LIST

#define HUGE_VALL HUGE_VAL
#define ETL_NO_LIBC_WCHAR_H
```

What exactly went wrong in your case?  Theoretically, the above line in etl_profile.h should have the same effect as your direct inclusion of tuple. Assuming that OpenBSW's etl_profile.h is being found correctly during build.

You can also describe how I can reproduce the issue you encountered, for reference.

vladislavmarkov: I was fiddling with ETL_NO_STL, ETL_USING_STL, ETL_USING_CPP11, because the part of ETL includes STL's <tuple> looks like this (etl/tuple.h).
```
#include "platform.h"

#if ETL_NOT_USING_CPP11 && !defined(ETL_IN_UNIT_TEST)
  #error NOT SUPPORTED FOR C++03 OR BELOW
#endif

#if ETL_USING_CPP11

#if ETL_USING_STL
  #include <tuple>
#endif
```
The compilation error was
```
openbsw-devenv/openbsw/libs/3rdparty/etl/include/etl/tuple.h:1293:10: error: no template named 'tuple_size'; did you mean 'etl::tuple_size'?
 1293 |   struct tuple_size<etl::tuple<Types...>> : etl::integral_constant<size_t, sizeof...(Types)>
      |          ^~~~~~~~~~
      |          etl::tuple_size
```

rolandreichweinbmw: With this change:

```
diff --git a/libs/3rdparty/etl_configuration/etl_profile.h b/libs/3rdparty/etl_configuration/etl_profile.h
index 3605050a8..17ec95317 100644
--- a/libs/3rdparty/etl_configuration/etl_profile.h
+++ b/libs/3rdparty/etl_configuration/etl_profile.h
@@ -31,6 +31,8 @@ SOFTWARE.
 #ifndef ETL_PROFILE_H
 #define ETL_PROFILE_H

+#include <tuple>
+
 #define ETL_TARGET_DEVICE_GENERIC
 #define ETL_TARGET_OS_NONE

diff --git a/libs/bsw/uds/include/uds/connection/IncomingDiagConnection.h b/libs/bsw/uds/include/uds/connection/IncomingDiagConnection.h
index 31ccaf3fe..5c1298703 100644
--- a/libs/bsw/uds/include/uds/connection/IncomingDiagConnection.h
+++ b/libs/bsw/uds/include/uds/connection/IncomingDiagConnection.h
@@ -12,8 +12,6 @@

 // clang-format off

-// compilation on clang won't work without tuple_size
-#include <tuple>
 #include <etl/closure.h>

 // clang-format on
```

It works for me in your branch. Reproduced with:

```
cmake --preset tests-dbg
cmake --build --preset tests-dbg
ctest --preset tests-dbg --parallel

cmake --preset s32k148-gcc
cmake --build --preset s32k148-gcc
```

How would you reproduce the issue?

.ci/build_matrix.py
rolandreichweinbmw: In general I would try to use default compilers for host compile, since actual issues we expect would be compiler bugs (rare!), if you don't try on too old distributions.

Maintenance of explicit compiler versions is an additional effort, and using default compiler version make OpenBSW work with all modern Ubuntu and Debian versions.

By defining a default environment (currently Ubuntu 24.04), we already implicitly define reference versions of gcc and clang since the distributions have constant default compiler versions each.

vladyslavmarkovaccenture: Noted. I'm okay with either, since we still verify compilation for multiple compilers and compiler versions anyways - fixing the compiler version is probably not the most important thing on a project either way. I will change it to defaults then.

libs/bsw/docan/include/docan/datalink/DoCanFrameCodec.h
vladislavmarkov: You probably mean the extension from values below "native" register width when doing LDRB (extended by e.g. UXTB). That would be a concern if those were **not** `static constexpr` but `.data` or `.rodata` **variables**. After it's compiled there will be no extensions and it shouldn't impact `.text` section, unless it's _ODR_-ed - _materialized_ by taking it's address - that will force a compiler to put it into `.rodata` for example and load it as such. Static constexpr fundamentals are promoted to native registry width by the compiler.
TL;DR Changing it to uint8_t / uint16_t makes it neither worse nor better. My motivation was: "I see a small number value - I narrow down it's type".

libs/bsw/uds/include/uds/connection/IncomingDiagConnection.h
vladislavmarkov: I will try that tomorrow. I still don't understand though, why do we need to include `<tuple>` in profile, if it's included in `etl/tuple.h`? My expectation would be - configure ETL with ETL_* options and ETL itself figures out which version of a tuple to take. Looks like design flaw to me, but I might be missing something.

rolandreichweinbmw: It is an odd combination of circumstances. See also https://github.com/ETLCPP/etl/blob/master/include/etl/tuple.h at the end: Clang ships some header files in a way that it sometimes indirectly includes its own tuple header even though we don't do it directly. Since the ETL authors decided to use std::tuple_size for deduction guides, we need to consistently define the base template or not. In this case, we do, by including tuple unconditionally since depending on including some .h headers from the build environment it is only done sometimes and sometimes not.

Consider it a workaround what should be encapsulated somewhere. In this case, etl_profile.h

vladyslavmarkovaccenture: @rolandreichweinbmw we have etl_configuration in libs/3rdparty. I assume you added it. If it's in a 3rdparty, where can I find a repo which has it? Also if I understand correctly, `etl_profile.h` is project specific, which makes `executables/referenceApp` and `executables/unitTest` better candidates to store this file as not 3rdparty and have it customized for their needs. I'd assume, 3rdparty part can be either included **unmodified** (hence 3rdparty) from a project-specific `etl_profile.h`, or dropped completely. Please comment.

rolandreichweinbmw: That's correct. In this regard, etl_profile.h is OpenBSW specific and not part of upstream, and in fact not 3rdparty itself. However, we put it close to 3rdparty/etl since they relate to each other.

If you find a better home for it, that's fine for me.

vladyslavmarkovaccenture:
> Hi! Everytime you push, I get notification mails. If the purpose is to test the CI jobs, maybe you can reproduce those locally instead?

Hi @rolandreichweinbmw . I get twice as much emails (because I use both private and company account) on this PR. Everything that I could test locally I tested before creating this PR. All the changes I make now are related to github workflow..

PR #193, opened 2025-09-23
Extend and improve safety modules

marcmo:
originates from CR #135547


doc/features/functional_safety.rst
```diff
@@ -0,0 +1,162 @@
...
+     - A collection of templated classes which formalize a condition, its check and the reporting
+       event if the condition isn't met.
+   * - :ref:`safetyUtils`
+     - Provides helper classes like the declaration for the logger component.
+
+Watchdog Test
+-------------
+
+.. note::
+    This test resets the system twice, So if you notice two watchdog resets after the control
```
rolandreichweinbmw: typo? So -> so

executables/referenceApp/application/src/app/app.cpp
```diff
@@ -170,7 +182,7 @@ class IdleHandler : private ::async::RunnableType
 
     void shutdown()
     {
-        Logger::info(LIFECYCLE, "Lifecycle shutdown complete");
+        Logger::info(LIFECYCLE, "Lifecycle shutdown complete\n");
```
rolandreichweinbmw: Is the logged line break necessary here? (Not done anywhere else with Logger::info())

executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+            safetyUtils
+            safeMemory
+            util
+            bsp
+            bspInterrupts
+            startUp)
+
+if (BUILD_UNIT_TESTS)
+    target_compile_definitions(safeLifecycle PUBLIC ESTD_HAS_EXCEPTIONS
+                                                    ESTL_ASSERT_MESSAGE_ALL)
```
rolandreichweinbmw: I propose adding ETL_THROW_EXCEPTIONS for consistency reasons even though the tests don't need that currently. But we needed to fix it in other cases later on. So it can probably save us from trouble later on.

executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/include/safeLifecycle/IsrHooks.h
```diff
@@ -0,0 +1,11 @@
+// Copyright 2025 Accenture.
+
+#ifndef GUARD_ACC658E4_1322_4D4A_96C5_55FC25ECF550
```
rolandreichweinbmw: #pragma once

like everywhere else?

executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/include/safeLifecycle/IsrWrapper.h
```diff
@@ -0,0 +1,32 @@
+// Copyright 2025 Accenture.
+
+#ifndef GUARD_AA36739B_A60B_45A7_8232_AB4C9DBC63F3
```
rolandreichweinbmw: #pragma once

?

executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/include/safeLifecycle/IsrWrapper.h
```diff
@@ -0,0 +1,32 @@
...
+#define GUARD_AA36739B_A60B_45A7_8232_AB4C9DBC63F3
+
+// clang-format off
+#define SAFE_ISR(func) \
+void Safe_##func(void) \
+{                      \
+    SafePreIsrHook();  \
```
rolandreichweinbmw, line 9: It might be possible to implement this with a C++ style function wrapper with a RAII pattern for Pre and Post, instead of doing it in C-style, to prevent macro use.

Depends on your safety approach I guess?

executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/include/safeLifecycle/SafetyManager.h
```diff
@@ -0,0 +1,39 @@
...
+     * because only safety relevant software can write to the safety RAM region. Services the
+     * watchdog every 80ms by calling cyclic method of safeWatchdog. Calls the cyclic function of
+     * SafeMemory.
+     */
+    void cyclic();
+    // [PUBLIC_API_END]
+
+private:
+    uint16_t _counter;
+    uint8_t const WATCHDOG_CYCLIC_COUNTER = 8;
```
rolandreichweinbmw: static?

executables/referenceApp/platforms/s32k148evb/safety/safeMemory/src/safeMemory/MemoryProtection.cpp
```diff
@@ -0,0 +1,253 @@
...
+        return false;
+    }
+    return true;
+}
+
+bool MemoryProtection::areRegionsConfiguredCorrectly(uint8_t& failed_region)
+{
+    constexpr uint8_t MAX_MPU_REGIONS = 16U;
+
+    for (uint8_t i = 0U; i < numOfMemoryProtectionConfigurationRamRegions; ++i)
```
rolandreichweinbmw, line 225: Have you checked if using uint8_t instead of common practice size_t (as above) is really better here?

executables/referenceApp/platforms/s32k148evb/safety/safeMemory/src/safeMemory/RomCheck.cpp
```diff
@@ -0,0 +1,89 @@
...
+        // Write seed value
+        crcRegisters.CTRL       = crcRegisters.CTRL | CRC_CTRL_WAS(1);
+        crcRegisters.DATAu.DATA = seedValue;
+        crcRegisters.CTRL
+            = crcRegisters.CTRL
+              & ~(CRC_CTRL_WAS(1)); // Clear CRC_CTRL[WAS] to start writing data values to the CRC.
+
+        uint32_t const numOfAreas = __checksum_list_start[0];
+
+        for (uint32_t i = 0U; i < numOfAreas; ++i)
```
rolandreichweinbmw: size_t ?

executables/referenceApp/platforms/s32k148evb/safety/safeMemory/src/safeMemory/RomCheck.cpp
```diff
@@ -0,0 +1,89 @@
...
+        {
+            auto const startAddress
+                = reinterpret_cast<uintptr_t const*>(__checksum_list_start[i * 2 + 1]);
```
rolandreichweinbmw, line 55: startAddress is a pointer to a pointer (stored as unsigned integer type)?

It technically works, but maybe it is unnecessarily confusing? Finally, below, the static_cast rather hints for "uint32_t const*"?

libs/safety/safetyUtils/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+add_library(safetyUtils INTERFACE)
+
+target_include_directories(safetyUtils INTERFACE include)
+
+target_link_libraries(safetyUtils INTERFACE util)
+
+if (BUILD_UNIT_TESTS)
+    target_compile_definitions(safetyUtils INTERFACE ESTD_HAS_EXCEPTIONS
+                                                     ESTL_ASSERT_MESSAGE_ALL)
```
rolandreichweinbmw: Add ETL_THROW_EXCEPTIONS

executables/referenceApp/platforms/s32k148evb/safety/safeMemory/src/safeMemory/RomCheck.cpp
aschaal: Linker symbols are always a bit tricky.

I use uintptr_t, so I can unit test the code even on 64 bit pointer platforms (otherwise I would have to compile it in mixed mode). The content itself is uint32_t, because it's for a 32 bit platform (S32K1). In unit test, this needs to be considered.

On the target platform, the static cast would not be needed, but it's for unit test compatibility.

executables/referenceApp/platforms/s32k148evb/safety/safeMemory/src/safeMemory/RomCheck.cpp
aschaal: Yes, here it makes sense, as this is only for array access.

doc/features/functional_safety.rst
aschaal: Fixed.

executables/referenceApp/application/src/app/app.cpp
aschaal: Fixed.

executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/include/safeLifecycle/IsrHooks.h
aschaal: Fixed.

executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/CMakeLists.txt
aschaal: Fixed.

executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/include/safeLifecycle/IsrWrapper.h
aschaal: Fixed.

executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/include/safeLifecycle/SafetyManager.h
aschaal: Fixed, also made it constexpr.

executables/referenceApp/platforms/s32k148evb/safety/safeMemory/src/safeMemory/MemoryProtection.cpp
aschaal: I would like to stick with uint8_t, as it is not just a loop counter.
Later on, this is used as number of something, which cannot be more than 256.

Personally I prefer using this smaller type.

libs/safety/safetyUtils/CMakeLists.txt
aschaal: Fixed.

executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/include/safeLifecycle/IsrWrapper.h
aschaal: The IVT is in startUp.S, e.g.:

`.long   Safe_CAN0_ORed_0_15_MB_IRQHandler`

We can use RAII, but the macro would be similar:

```
#define SAFE_ISR(func) \
void Safe_##func(void) \
{                      \
    IsrHookRaii hook;  \
    func();            \
}
```

So, we use this pattern since quite a while, so I vote to keep it for now, to get this large commit in.
Let us discuss in a separate commit if it is worth to change it and another solution is more modern.

PR #198, opened 2025-09-25
Added new Uart interface using reverse inheritance

cristiancsordas:


platforms/s32k1xx/bsp/bspUart/module.spec
```diff
@@ -0,0 +1,2 @@
+maturity: raw
```
aschaal: "maturity" still exists in some module.specs, but we don't use it anymore since a long time.
Please remove this value.

libs/bsw/bsp/include/bsp/uart/Uart.h
```diff
@@ -0,0 +1,46 @@
...
+class Uart : public UartImpl
+{
+public:
+    /**
+     * sends out an array of bytes
+     * @param data - pointer to the data to be send
+     *        length - the number of bytes to be send
+     * @return the number of bytes written
+     */
+    size_t write(uint8_t const* data, size_t length);
```
aschaal: Use ETL data structure for pointer+length.
(for write and read)

libs/bsw/bsp/include/bsp/uart/Uart.h
```diff
@@ -0,0 +1,46 @@
...
+#include <cstddef>
+#include <cstdint>
+
+namespace bsp
+{
+
+/**
+ * Base class used to define the generic interface for the uart communication
+ */
+class Uart : public UartImpl
```
aschaal: I don't like the naming here, as "Impl" is usually the specialized class.

platforms/s32k1xx/bsp/bspUart/include/bsp/UartImpl.h
```diff
@@ -0,0 +1,53 @@
...
+struct Uart;
+
+/**
+ * This class implements the UART communication for S32K1xx platforms.
+ * It is used as a base class for the Uart class.
+ * The Uart class implements the methods for writing and reading data
+ * over the UART interface.
+ * The UartImpl provides methods specifics to the S32K1xx platform.
+ */
+class UartImpl : public bsp::UartConfig
```
aschaal: Please avoid this inheritance chain.

Uart -> UartImpl - >UartConfig

You may think about "Composition Over Inheritance".
The config as base-base class is btw super strange.

The config should be a (template) parameter.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
```diff
@@ -0,0 +1,58 @@
...
+    },
+};
+
+bsp::Uart& Uart::getInstance(Id id)
+{
+    switch (id)
+    {
+        case Id::TERMINAL:
+        {
+            static Uart instance{
```
aschaal: Avoid function local static variables.
There is a hidden state variable generated by the compiler and always an "if" which checks if the instance is already created. Futhermore, it's not defined in C++ standard where this state variable is placed in memory.

Please find another way for creating those instances.
If you want lazy initialization, in most cases it's better to call an init() method later, not the constructor.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
```diff
@@ -0,0 +1,58 @@
...
+    {
+        *LPUART1,
+        bios::Io::PORT_UNAVAILABLE,
+        bios::Io::PORT_UNAVAILABLE,
+        0,
+        nullptr, // Dummy UART does not have a valid configuration
+    },
+};
+
+bsp::Uart& Uart::getInstance(Id id)
```
aschaal: If you really want to have that getInstance() dynamically, think about "ETL optional".

Beside from that, where is that needed? If another class needs an instance of Uart, you may hand it over in the constructor or init() of a lifecycle system...

platforms/s32k1xx/bsp/bspUart/include/bsp/UartImpl.h
andreychekasinbmw: Using inheritance for configuration is intended to replace includes, which are used, for example, [here](https://github.com/eclipse-openbsw/openbsw/blob/08d69a204ddc6c2d2dfc5d4c37d67760f6ba7bec/platforms/s32k1xx/bsp/bspIo/include/io/Io.h#L126) and [here](https://github.com/eclipse-openbsw/openbsw/blob/08d69a204ddc6c2d2dfc5d4c37d67760f6ba7bec/platforms/s32k1xx/bsp/bspIo/src/io/Io.cpp#L14). Direct including in the middle of a file is dangerous and confusing, but the idea of class scope extension looks pretty convenient. In this way, we can extend a class scope with application-specific constants like IDs, so that it's possible to write something like `Uart::getInstance(Uart::TERMINAL)`

aschaal: Fully agree, the current solution with the includes in the middle of a file is not good.

But I am also not sure about the getInstance approach. We have lifecycle systems, so if an objects needs Uart, this can be handed over there (even without inheritance or interfaces).

You can still use your idea of "reverse dependency" if Uart "has" a UartImpl as member.

platforms/s32k1xx/bsp/bspUart/module.spec
cristiancsordas: done. I removed the maturity:raw

libs/bsw/bsp/include/bsp/uart/Uart.h
cristiancsordas: I added etl::span for the read and write methods

libs/bsw/bsp/include/bsp/uart/Uart.h
cristiancsordas: I renamed the UartImpl into UartSpecific

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
cristiancsordas: the instance object is declared now outside of the getInstance() method. It is global (renamed into instance_terminal) . The switch/case returns a reference to this object (please check the UartConfig.cpp).
The creation of the object outside of the class forces the Uart class to have the constructor public (please check Uart.h).
I decided to move the getInstance() to UartSpecific.h. A class which returns a Singleton through getInstance() and has a public constructor is confusing.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
cristiancsordas: getInstance() should always return a reference to an object. ETL optional would imply that the object could not exist, like for the DUMMY_UART. But, trying to use such an object should result in an error (assert).

I agree that from the object oriented perspective it is good to pass the object as a reference to a class so that that class can handle different instances from the same type. 
But since for Uart in most of the projects will exist only one instance which handle the inputs and outputs of the terminal I think that is very practical to be able to call it from everywhere with just Uart::getInstance(Id::TERMINAL).

aschaal: > But since for Uart in most of the projects will exist only one instance

You have the id as argument, so obviously projects can have several instances.

In general, asserts in productive code should be avoided due to several reason. It's better to have a clear return code.
(as return value of the function, out parameter, etc. - or with ETL optional you have that already)

If you don't want that, I suggest thinking about another approach without "getInstance".

cristiancsordas: The Uart instances are instantiated now using a global etl::tuple object. I removed the Uart_Dummy configuration.
The UartConfig::Id is an enum class, therefore getInstance(Id id) can be called only with the values of Id (TERMINAL and INVALID).
I think that if getInstance() is called with the INVALID parameter a platform specific function for error handling should be called. 
For ex: for bare metal platform saving the error in a no init ram, do a reset and handle the error on startup. For Posix, log the error on terminal and kill the process.
For now an assert should be sufficent.

platforms/s32k1xx/bsp/bspUart/include/bsp/UartImpl.h
mthiede-acn2: With the inheritance approach, the UART interface can be extended by a platform specific implementation (e.g. a variant of the "write" function). I think we agreed that we want to build an extensible system since we can't foresee all the platform specific details.

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
mthiede-acn2: In your proposal, Uart instances are created in the UartConfig.cpp file. I'd like to point out that this config file plays a very similar role to the StaticBsp: it instanciates the driver object (here uart) as a global variable and provides a global getter function to it. At the moment, with static BSP, one could do `getStaticBsp().getAdc()` for example to get hold of a driver object from global scope. If we follow your proposal, we have to decide what to do with "static BSP", in my understanding it would disappear and its contents would be spread over several config files (one for each driver).

cristiancsordas: Having these components independent, and not part of one big object(like staticBSP) is one of the main scope of this approach. Uart, like most of the other hw peripherals, are, at least in my opinion, stand alone components. 
The configuration of a such component is done per application in its corresponding *_config.cpp file.
This is also why the instance object is instantiated in the _config.cpp files. 
The separation is very clear, somebody interested in configuring or reading the configuration of a HW component will just have to modify or read the *_config.cpp file and not going through a staticBSP which contains every HW components.

mthiede-acn2: I see. One thing to keep in mind though is that there are peripherals where the configuration is dependent on each other. That's where an integrated approach could have advantages. However, such dependencies can probably also be taken care of across several _config.cpp files. In the end, that's an implementation detail of the _config.cpp files. Essentially, one would split the "static bsp" into several _config.cpp files.

mthiede-acn2: The next question would be how we integrate the proposed structure into the lifecycle. Today we have a BspSystem which calls startup and shutdown on BSP components as well as it does the cyclic scheduling in a task for components which need that. So one way I could imagine is to keep the BspSystem as it is and call the startup/shutdown/cyclic APIs of the bsp components from there. Instead of passing a reference to StaticBsp into the BspSystem, we would directly access the globally available API implemented by the components and the config files.

cristiancsordas: > I see. One thing to keep in mind though is that there are peripherals where the configuration is dependent on each other. That's where an integrated approach could have advantages. However, such dependencies can probably also be taken care of across several _config.cpp files. In the end, that's an implementation detail of the _config.cpp files. Essentially, one would split the "static bsp" into several _config.cpp files.

I think that would be good to split "static bsp" between different config files. 
But, this is not a must, I think that this approach is very flexible. If somebody wants to keep a big staticBsp he still can do it, but with some differences. For ex in the case of Uart, the constructor will just need the Id of the Uart and not the configuration. You are still able to instantiate Uart objects. 
Dependencies can be split between different _config.cpp files. But this is also flexible and is up to the application where they instantiate the objects. In the _config.cpp could be used only references to the dependent objects.

cristiancsordas: > The next question would be how we integrate the proposed structure into the lifecycle. Today we have a BspSystem which calls startup and shutdown on BSP components as well as it does the cyclic scheduling in a task for components which need that. So one way I could imagine is to keep the BspSystem as it is and call the startup/shutdown/cyclic APIs of the bsp components from there. Instead of passing a reference to StaticBsp into the BspSystem, we would directly access the globally available API implemented by the components and the config files.

I would say that this is flexible and is up to the user how he creates his application. He could use getInstace() calls inside BspSystem, he could pass references to these objects in the constructor or creates a staticBsp and use it inside BspSystem

aschaal: With the global variables approach, is it possible to gmock them for unit tests?
This might be important for components with interdependencies.

Difficult to show this in this PR... Maybe add another (temporary BSP) component which uses an UartMock... What do you think?

platforms/s32k1xx/bsp/bspUart/src/UartSpecific.cpp
```diff
@@ -0,0 +1,106 @@
...
+        }
+        counter++;
+    }
+
+    return counter;
+}
+
+UartSpecific::UartSpecific(Id id) : _uartDevice(config_uart[static_cast<uint8_t>(id)]) {}
+
+void UartSpecific::init()
```
aschaal, line 56: Who calls this method in this PR? Maybe I overlook that...

executables/unitTest/bsp/bspUart/module.spec
```diff
@@ -0,0 +1,3 @@
+maturity: raw
```
aschaal: Please remove maturity variable here as well.

cristiancsordas: done

platforms/s32k1xx/bsp/bspUart/src/UartSpecific.cpp
cristiancsordas: the init() method is not yet called. This Uart implementation is not yet used.
I thought that is better to do it in 2 steps:
- first add this implementation as a new concept for creating the bsp interfaces.
- If this PR will be merged then replace the SciDevice with this new implementation. Instead of having sciInit() then call 
Uart::getInstance(Terminal).init().

executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp
```diff
@@ -0,0 +1,39 @@
...
+    },
+};
+
+etl::tuple<bsp::Uart> uart_instances{Uart(Uart::Id::TERMINAL)};
+
+bsp::Uart& UartSpecific::getInstance(Id id)
+{
+    switch (id)
+    {
+        case Id::TERMINAL: return etl::get<0>(uart_instances); break;
```
aschaal, line 33: This code RELIES on the fact, that TERMINAL is 0.

Maybe better:

```
if (id >=  Id::INVALID)
  assert();
return etl::get<id>(uart_instances);
```

aschaal: Or even better, check size of uart_instances. Then this method is 100% generic and might be moved to another place, e.g. to Uart.h, then the user does not need to implement that.

platforms/s32k1xx/bsp/bspUart/src/UartSpecific.cpp
aschaal: The first (simple) integration is needed to prove that the concept is really working. Maybe we overlook something which blocks the integration...

cristiancsordas: I will add a simple integration

platforms/s32k1xx/bsp/bspUart/src/UartSpecific.cpp
```diff
@@ -0,0 +1,106 @@
...
+        {
+            break;
+        }
+        counter++;
+    }
+
+    return counter;
+}
+
+UartSpecific::UartSpecific(Id id) : _uartDevice(config_uart[static_cast<uint8_t>(id)]) {}
```
aschaal, line 54: This comment is more general, as we are also discussing if that approach is suitable as concept for all BSP components:

This UART device / config (applies to any other component as well) has member variables, which can be overwritten before the MPU protection is active --> possible FFI violation...

So the init() method MUST (re)set all member variables, including the ID.
This means the init() method also needs to get the ID again. You may leave _uartDevice uninitialized in the constructor (also not setting to zeros due to performance reasons).

You may think about that.

Alternatives (maybe not so good):
- a placement-new to recreate Uart during lifecycle after MPU is active (performance?)
- logic to compare the intended config / member variables with actual values (public access to private members?)

PR #199, opened 2025-09-25
Discuss about Abstraction of Application from Drivers

christophruethingbmw:


christian-schilling:
I agree there is quite some confusion and room for improvement here.
Unfortunately I don't understand the diagram you made as the handwriting is pixelated and not legible.
I also agree that something like the current EthernetSystem could/should be reusable for different applications.
Regarding naming: Passing a reference to `IEthernetSystem` to `EthernetSystem` looks confusing, thats why the interface was named `IEthernetDriverSystem`.
I could definitely imagine to have interfaces for the drivers instead of the systems and then a system that can be configured to use different drivers.

PR #207, opened 2025-09-30
Fix: Add timeout to prevent pytest hanging on CAN/Ethernet tests

marcmo:
originates from CR #135689

### DIFF
--- a/.ci/pytest.py
+++ b/.ci/pytest.py
@@ -17,8 +17,7 @@ def build_reference_app():
 
 def run_pytest():
     os.chdir("./test/pyTest")
-    subprocess.run(["pytest", "--target=posix"], check=True)
-
+    subprocess.run(['pytest', '--target=posix', '--timeout=120'], check=True)
 
 if __name__ == "__main__":
     try:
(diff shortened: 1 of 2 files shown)


.ci/pytest.py
```diff
@@ -17,8 +17,7 @@ def build_reference_app():
 
 def run_pytest():
     os.chdir("./test/pyTest")
-    subprocess.run(["pytest", "--target=posix"], check=True)
-
+    subprocess.run(['pytest', '--target=posix', '--timeout=120'], check=True)
```
christian-schilling, line 20: Introducing timeouts to tests, especially those that run in CI checks, seems like a *very* bad idea.
Why do those tests hang in the first place? How can the cause be addressed?
What should happen if a timeout occurs? Should the job be retried?

shamitha-shashidhara: Hi @christian-schilling 
Test was hanging mainly because of a software change that led to a can communication issue (there were two such scenarios noted where the software change caused an issue with can and there were no can messages). Since this scenario is not because of the test design and the test hangs due to a software change,  pytest timeout were introduced where the test is interrupted after 120sec of inactivity.

nemadhu: @christian-schilling @shamitha-shashidhara After merging #264, the test environment itself would take care of the scenarios and this PR can be closed.

PR #219, opened 2025-10-06
Remove support for outgoing diag connections

christian-schilling:
We don't need outgoing connection in most cases and removing them will make further
refactoring of UDS simpler.
We can add a "UDS client" as separate module later if needed.


rolandreichweinbmw:
LGTM - my colleague will check if this is currently in use or can easily be missed temporarily.

PR #221, opened 2025-10-07
Combine IDiagDispatcher and IResumableDiagDispatcher

christian-schilling:
The base interface was never used directly so we don't need two distinct interfaces.

Change-Id: Ib5fbb30c0dd1f05ae85d0dbe80af40014650bd4c


libs/bsw/uds/include/uds/resume/ResumableResetDriver.h
```diff
@@ -93,7 +93,7 @@ class ResumableResetDriver
     void dispatchMessage();
 
     IUdsLifecycleConnector& fUdsLifecycleConnector;
-    IResumableDiagDispatcher* fResumableDiagDispatcher;
+    IDiagDispatcher* fDiagDispatcher;
     ::async::ContextType fContext;
```
matthiaskessler, line 97: We might reconsider the order of the members to avoid padding in the class layout.

libs/bsw/uds/mock/gmock/include/uds/resume/ResumableDiagDispatcherMock.h
matthiaskessler, line 1: Isn't this the same as `DiagDispatcherMock` now?

libs/bsw/uds/include/uds/resume/ResumableResetDriver.h
christian-schilling: We might. It's not really related to this change though, so I would do this as a separate change.

matthiaskessler: 👍

PR #222, opened 2025-10-07
Add asyncThreadX module

marcmo:
originates from CR #136009


johannes-esr:
@ermankomurcu, for the other third-party modules we (mostly) used Rim - would this also be possible for the ThreadX modules?

ermankomurcu:
@johannes-esr We can use Rim to integrate the ThreadX module. In this PR, we followed a similar approach to the one used for FreeRTOS.

platforms/posix/3rdparty/threadx_linux/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+    # {{END_TARGET_SOURCES}}
+)
+
+target_include_directories(${PROJECT_NAME}
+    PUBLIC
+    ${CMAKE_CURRENT_LIST_DIR}/inc
+)
+
+target_compile_definitions(${PROJECT_NAME} PUBLIC "-D_GNU_SOURCE -DTX_LINUX_DEBUG_ENABLE")
```
jiangyilism, line 22: `target_compile_definitions(${PROJECT_NAME} PUBLIC "_GNU_SOURCE" "TX_LINUX_DEBUG_ENABLE")`
To be consistent with the repo.
optionally remove the quotes.

.github/workflows/build.yml
```diff
@@ -39,21 +40,35 @@ jobs:
...
-          - preset: "s32k148-gcc"
+
+          - preset: "s32k148-freertos-gcc"
             platform: "linux"
-          - preset: "s32k148-clang"
+          - preset: "s32k148-threadx-gcc"
             platform: "linux"
-          - preset: "s32k148-gcc"
+          - preset: "s32k148-freertos-gcc"
             cxxid: "clang"
```
jiangyilism, line 61: `cxxid: "gcc"` ?

.github/workflows/build.yml
```diff
@@ -39,21 +40,35 @@ jobs:
...
+            cxxid: "clang"
+
+          - preset: "s32k148-freertos-clang"
+            platform: "linux"
+          - preset: "s32k148-threadx-clang"
+            platform: "linux"
+          - preset: "s32k148-freertos-clang"
+            cxxid: "gcc"
+          - preset: "s32k148-threadx-clang"
             cxxid: "gcc"
```
jiangyilism, line 72: `cxxid: "clang"` ?

CMakePresets.json
```diff
@@ -142,17 +241,53 @@
...
+            "configurePreset": "s32k148-freertos-gcc"
+        },
+        {
+            "name": "s32k148-threadx-gcc",
+            "displayName": "S32K148 ThreadX build (GCC)",
+            "description": "Build reference application for S32K148 platform with Clang (Release by default; for Debug use --config Debug)",
+            "configurePreset": "s32k148-threadx-gcc"
+        },
         {
             "name": "s32k148-clang",
```
jiangyilism, line 275: Is it still needed? Or completed replaced by s32k148-freertos-clang and s32k148-threadx-clang ?

CMakePresets.json
```diff
@@ -142,17 +241,53 @@
...
+            "configurePreset": "posix-freertos"
+        },
+        {
+            "name": "posix-threadx",
+            "displayName": "build POSIX-THREADX",
+            "description": "Build reference application for POSIX-compliant environment (Release by default; for Debug use --config Debug)",
+            "configurePreset": "posix-threadx"
+        },
         {
             "name": "s32k148-gcc",
```
jiangyilism, line 257: Is it still needed? Or completed replaced by s32k148-freertos-gcc and s32k148-threadx-gcc ?

.github/workflows/build.yml
rolandreichweinbmw: This is the exclude list. So typically lists the unused combinations.

.github/workflows/build.yml
rolandreichweinbmw: This is the exclude list. So typically lists the unused combinations.

executables/referenceApp/application/CMakeLists.txt
```diff
@@ -79,6 +74,17 @@ target_link_libraries(
             safeWatchdog
             socBsp)
 
+if (BUILD_TARGET_RTOS STREQUAL "FREERTOS")
+    target_link_libraries(app.referenceApp PRIVATE osFreeRtos asyncFreeRtosImpl)
+elseif (BUILD_TARGET_RTOS STREQUAL "THREADX")
+    target_link_libraries(app.referenceApp PRIVATE osThreadX asyncThreadXImpl)
```
jiangyilism: Is abstracting rtos here a good idea?
e.g.
`target_link_libraries(app.referenceApp PRIVATE osRtos asyncRtosImpl)`

osRtos corresponds to osFreeRtos or osThreadX . Can have a better name.

platforms/posix/3rdparty/threadx_linux/CMakeLists.txt
rolandreichweinbmw: Make sure to remove the "D" also to prevent defining the wrong symbol (with added D). See also

https://cmake.org/cmake/help/latest/command/target_compile_definitions.html

Thanks @jiangyilism for the hint!

platforms/posix/3rdparty/threadx_linux/CMakeLists.txt
```diff
@@ -1,22 +1,28 @@
...
-target_compile_definitions(${PROJECT_NAME} PUBLIC "-D_GNU_SOURCE -DTX_LINUX_DEBUG_ENABLE")
\ No newline at end of file
+target_include_directories(threadXLinux PUBLIC ${CMAKE_CURRENT_LIST_DIR}/inc)
+
+target_link_libraries(
+    threadXLinux PUBLIC bspInterrupts threadX threadXCoreConfiguration
+                        Threads::Threads)
+
+target_compile_options(threadXLinuxPort INTERFACE -DTX_INCLUDE_USER_DEFINE_FILE)
+target_compile_options(threadXLinux PRIVATE -DTX_INCLUDE_USER_DEFINE_FILE)
```
jiangyilism: Suggest to remove `-D` for consistency within the repo.
Another question, should `TX_INCLUDE_USER_DEFINE_FILE` be public instead?
We do not want any files accessing threadx headers without `TX_INCLUDE_USER_DEFINE_FILE` being defined.

platforms/posix/3rdparty/threadx_linux/CMakeLists.txt
ermankomurcu: Acknowledged, I’ll adjust it in the next revision.

CMakePresets.json
ermankomurcu: I have kept it for now to maintain compatibility with the current documentation. It implicitly refers to the s32k148-freertos-clang preset, since FreeRTOS is the project’s default RTOS. I’ll remove it once the documentation is updated.

CMakePresets.json
ermankomurcu: I have kept it for now to maintain compatibility with the current documentation. It implicitly refers to the s32k148-freertos-gcc preset, since FreeRTOS is the project’s default RTOS. I’ll remove it once the documentation is updated.

platforms/posix/3rdparty/threadx_linux/CMakeLists.txt
ermankomurcu: That makes sense. Thanks for the suggestion! I’ll apply the change accordingly.

NOTICE.md
```diff
@@ -41,6 +41,9 @@ We recommend to read their licenses, as their terms may differ from the terms de
 | FreeRTOS Real Time Kernel | 10.6.2  | MIT       | ``platforms/posix/3rdparty/freeRtosPosix/LICENSE.md``                |
 | FreeRTOS Real Time Kernel | 10.6.2  | MIT       | ``platforms/s32k1xx/3rdparty/freertos_cm4_sysTick/LICENSE.md``       |
 | FreeRTOS Real Time Kernel | 10.6.2  | MIT       | ``libs/os/3rdparty/freeRtos/LICENSE.md``                             |
+| ThreadX Kernel            | 6.4.1   | MIT       | ``libs/3rdparty/threadx``                                            |
+| ThreadX Cortex M4 Port    | 6.4.1   | MIT       | ``platforms/s32k1xx/3rdparty/threadx_cortex_m4``                     |
+| ThreadX Linux Port        | 6.4.1   | MIT       | ``platforms/posix/3rdparty/threadx_linux``                           |
 | CMSIS                     | 6.1.0   | Apache v2 | ``platforms/s32k1xx/bsp/bspMcu/include/3rdparty/cmsis/LICENSE``      |
```
rolandreichweinbmw, line 47: Can ThreadX be included with support of RIM? See also:

https://github.com/eclipse-openbsw/openbsw/blob/main/doc/guidelines/3rdparty.rst

This holds for the case we include it unchanged from a public git repo.

If not, it should be documented, what local modifications we include.

platforms/posix/3rdparty/threadx_linux/CMakeLists.txt
ermankomurcu: Outdated. I have removed this part from the CMake file but kept the same definitions in the original platforms/posix/3rdparty/threadx/ports/linux/gnu/CMakeLists.txt file without any changes.

executables/referenceApp/application/CMakeLists.txt
ermankomurcu: We need these two definitions to differentiate selected platform(posix or s32k1xx) target and unittest mock target.

platforms/posix/3rdparty/threadx_linux/CMakeLists.txt
ermankomurcu: Done.

NOTICE.md
ermankomurcu: With the last commit, I use RIM to integrate the threadx modules. I have also updated the documentation for the rim usage.

executables/referenceApp/application/CMakeLists.txt
jiangyilism: I mean we can make osRtos an alias to osFreeRtos or osThreadX. And asyncRtosImpl an alias to asyncFreeRtosImpl or asyncThreadXImpl.   So many target can just link osRtos/asyncRtosImpl without branching on BUILD_TARGET_RTOS (And avoid BUILD_TARGET_RTOS string compare in many cases).

ermankomurcu: I have implemented the change you suggested. The selected targets no longer depend on osFreeRtos or osThreadX. Instead, they now only use osRtos and asyncRtosImpl.

jiangyilism (approved):
Looks good to me.

libs/bsw/asyncThreadX/include/async/TaskContext.h
```diff
@@ -0,0 +1,306 @@
...
+#include "async/EventPolicy.h"
+#include "async/RunnableExecutor.h"
+#include "async/Types.h"
+#include "tx_api.h"
+
+#include <bsp/timer/SystemTimer.h>
+#include <etl/delegate.h>
+#include <etl/span.h>
+#include <timer/Timer.h>
+#include <util/estd/assert.h>
```
rolandreichweinbmw: Needs to be rebased / adjusted to new ETL_ASSERT() interface.

ermankomurcu: I just rebased.

.github/workflows/build.yml
```diff
@@ -79,24 +94,24 @@ jobs:
           docker run --rm -v "$PWD:/home/jenkins" -w /home/jenkins --user $(id -u):4996 dev python3 .ci/build.py --preset "${{ matrix.preset }}" --platform "${{ matrix.platform }}" --cxxid "${{ matrix.cxxid }}" --cxxstd "${{ matrix.cxxstd }}" --config "${{ matrix.config }}"
 
       - name: Run custom CMake build for elf-analyzer
-        if: steps.cache-cmake.outputs.cache-hit != 'true' && matrix.preset == 's32k148-gcc' && matrix.platform == 'arm' && matrix.cxxid == 'gcc' && matrix.cxxstd == '14' && matrix.config != ''
+        if: steps.cache-cmake.outputs.cache-hit != 'true' && matrix.platform == 'arm' && matrix.cxxid == 'gcc' && matrix.cxxstd == '14' && matrix.config != ''
```
SuhashiniNaik, line 97: Hi,
Could you help me understand why the `s32k148-gcc` preset was removed?
We still need the generated ELF artifacts for `Debug`, `RelWithDebInfo`, and `Release` builds for analysis in the ELF visualization project, so this preset seems necessary.

example we used to have artifacts generated like this: [Artifact generation example](https://github.com/eclipse-openbsw/openbsw/actions/runs/18902162054)

rolandreichweinbmw: Isn't it rather the case that the check for s32k148-gcc is rather redundant? I.e. already defined by the following combination of "gcc" and "arm". So the run is not prevented.

ermankomurcu: In this case, we need to produce two different elf files for presets s32k148-freertos-gcc and s32k148-threadx-gcc. Since the conditions matrix.platform == 'arm' and matrix.cxxid == 'gcc' are already used in the if clause, there is no need to explicitly include these presets in the condition. matrix.platform == 'arm' includes that only s32k148-x-x presets are selected and then matrix.cxxid == 'gcc' further filters them down to s32k148-x-gcc. Therefore, no need to add the specific presets ( s32k148-freertos-gcc and s32k148-threadx-gcc) to the condition

SuhashiniNaik: Thank you for the clarification.  I checked the recent runs and there’s a change in the artifact naming which may affect our analysis workflow:

Previously, the artifacts were named elf-s32k148-gcc-14-Debug…
[link](https://github.com/eclipse-openbsw/openbsw/actions/runs/18902162054)

Now they are named elf-s32k148-freertos-gcc-14-Debug…
[link](https://github.com/eclipse-openbsw/openbsw/actions/runs/18878372193)

Just for the confirmation, Is the freertos qualifier now always part of the preset (or matrix) for the S32K148 build?
also will the plain s32k148-gcc variant still run, or has it been replaced by s32k148-freertos-gcc
so the builds still generate the same ELF artifacts for Debug, RelWithDebInfo and Release

ermankomurcu: Since we added a new rtos option (**ThreadX**), I have updated the preset names to reflect the selected rtos. Previously, there was only one rtos option (freertos), so it was not necessary to include the rtos name in the preset name.

I have kept the existing presets (s32k148-gcc and s32k148-clang) for now, and they still refer to the default freertos option to maintain compatibility with the current documentation. However, these presets will be removed once the documentation is updated.

At this point, there is no need to test these old presets.

SuhashiniNaik: Thanks @ermankomurcu  that makes sense.

PR #226, opened 2025-10-09
Remove LOGSYNCHRON completely

aschaal:
- Replace the remaining usages with "printf"
- Remove commonDebug.h, which defines this MACRO
- Adapt documentation


platforms/s32k1xx/bsp/bspTja1101/src/phy/Tja1101Tester.cpp
```diff
@@ -2,10 +2,11 @@
 
 #include "phy/Tja1101Tester.h"
 
-#include "commonDebug.h"
 #include "phy/Tja1101.h"
 #include "util/format/SharedStringWriter.h"
 
+#include <cstdio>
```
matthiaskessler: Do we need this at all? It seems as if `LOGSYNCHRON` was not used in this file anyways and `printf` is also not used but `writer.printf`.

executables/referenceApp/safety/safeSupervisor/src/safeSupervisor/SafeSupervisor.cpp
```diff
@@ -70,7 +70,7 @@ void SafeSupervisor::handle(Event const& event)
         }
         case Event::WATCHDOG_STARTUP_CHECK_FAILURE:
         {
-            LOGSYNCHRON("Event: WATCHDOG_STARTUP_CHECK_FAILURE\n");
```
matthiaskessler, line 73: If you replace `LOGSYNCHRON` by `:Logger::error`, do you need to include `<cstdio>` at all?

executables/referenceApp/platforms/s32k148evb/main/src/os/isr/isr_wdg.cpp
```diff
@@ -1,6 +1,6 @@
 // Copyright 2024 Accenture.
 
-#include "commonDebug.h"
+#include <cstdio>
```
matthiaskessler: Is the include needed at all?

aschaal: No. Now removed.

platforms/s32k1xx/bsp/bspTja1101/src/phy/Tja1101Tester.cpp
aschaal: You are right, I have seen `writer.printf`, which is not `printf`.
Now removed.

executables/referenceApp/safety/safeSupervisor/src/safeSupervisor/SafeSupervisor.cpp
aschaal: No, you are right. Now removed.

Issue #86, opened 2025-04-02
Middleware / RPC support


rolandreichweinbmw:
We are planning to contribute a Middleware implementation in terms of RPC interface for events and method calls across cores when OpenBSW is instantiated on multiple cores.

Communication will be done via shared memory.

vladyslavmarkovaccenture:
@rolandreichweinbmw since we're planning to switch to ETL, you can use ETL.

rolandreichweinbmw:
I could only open the same kind of issue again. The only option I currently have is to call it either "Bug Report", "Feature Request", "Blank Issue", "Security Vulnerability". But I don't have the credentials to add the "enhancement" label.

(Unfortunately, I also can't re-open this issue by myself, so please decide if you re-open the issue or we set up a new one.)

mthiede-acn2:
Reopening this issue to represent the middleware enhancement.

mthiede-acn2:
It would be great if you could briefly describe the planned feature here.

rolandreichweinbmw:
(Edited the initial description above.)

PR #227, opened 2025-10-09
Create Message and Queue objects

filipe-cuim:
## Description

This PR is the first of the contribution of the Middleware/RPC framework that was mentioned by @rolandreichweinbmw on issue https://github.com/eclipse-openbsw/openbsw/issues/86.
This framework is based on a message passing mechanism, and as such I created a simple PR with just the Message and Queue objects that are being used.

### MiddlewareMessage

This is the message entity, which can be of 5 possible types:
- Request
- FireAndForgetRequest (a request that does not expect a response)
- Response
- Event
- Error

The payload of this message entity will be filled later by an entity called MessageAllocator

### Queue

This object is intended to be a multi-producer single-consumer queue, based on a circular buffer.
It provides two internal nested classes for reading and writing operations.


rolandreichweinbmw (commented):
Also, please note that in the new main branch, the additional gtest/ sub directories under test/ are not included anymore.

libs/bsw/middleware/include/middleware/core/types.h
```diff
@@ -0,0 +1,77 @@
...
+#include <etl/platform.h>
+
+#include <cstdint>
+
+namespace middleware
+{
+namespace core
+{
+
+using ServiceId = uint16_t;
```
christian-schilling: Why those type aliases?
They provide no type safety while obfuscating the readability of the code.

libs/bsw/middleware/include/middleware/core/middleware_message.h
```diff
@@ -0,0 +1,472 @@
...
+    struct ExternalHandle
+    {
+        constexpr ExternalHandle() = default;
+
+        uint32_t offset{0U}; ///< The offset, from the beginning of the memory region dedicated for
+                             ///< storing middleware data, where the message's payload is stored.
+        uint32_t size{0U};   ///< The size of the payload that is stored.
+        bool isPayloadShared{false}; ///< Tells if the payload is shared by several messages.
+
+        friend bool operator==(ExternalHandle const& lhs, ExternalHandle const& rhs)
```
christian-schilling: Why does this need to be a `friend`? There is no private...

libs/bsw/middleware/include/middleware/core/Message.h
```diff
@@ -0,0 +1,472 @@
...
+     *
+     */
+    struct ExternalHandle
+    {
+        constexpr ExternalHandle() = default;
+
+        uint32_t offset{0U}; ///< The offset, from the beginning of the memory region dedicated for
+                             ///< storing middleware data, where the message's payload is stored.
+        uint32_t size{0U};   ///< The size of the payload that is stored.
+        bool isPayloadShared{false}; ///< Tells if the payload is shared by several messages.
```
christian-schilling: Using for bytes to store a single bit seems wasteful. Could one bit of the other fields be used for this?

libs/bsw/middleware/include/middleware/core/middleware_message.h
```diff
@@ -0,0 +1,472 @@
...
+    {
+        flags_ &= ~static_cast<uint8_t>(flag);
+    }
+
+    Header
+        header_; ///< The message header containing the service, method, request and instance ids.
+    ClusterId srcClusterId_; ///< The source cluster id.
+    ClusterId tgtClusterId_; ///< The target cluster id.
+    AddressId addressId_;    ///< The proxy unique id.
+    mutable uint8_t flags_;  ///< Flags bitmask which contains a combination of MessageFlags values.
```
christian-schilling: Why is `mutable` needed? I doubt that it can't be avoided. But if it is really necessary it should be documented why.

libs/bsw/middleware/include/middleware/core/middleware_message.h
```diff
@@ -0,0 +1,472 @@
...
+ * \details This object has 32 bytes in total which are distributed between its private members.
+ * These consist of the following members:
+ * - header, which is similar to the SOME/IP's message ID and request ID members;
+ * - some dispatching information like the source cluster from where the message originates, the
+ * target cluster to which the message needs to go and an address ID that is unique to each proxy
+ * instance;
+ * - a payload which is a union of a buffer, and external handle that contains information to where
+ * the actual payload might be stored or an error value.
+ */
+class MiddlewareMessage
```
christian-schilling: Why is this not a plain struct with some free helper functions?
Looks like a lot of boilerplate with no real benefit to me. (See also recent discussion about trivial setters, the purpose of class/private and invariants.

libs/bsw/middleware/include/middleware/core/middleware_message.h
```diff
@@ -0,0 +1,472 @@
...
+    AddressId addressId_;    ///< The proxy unique id.
+    mutable uint8_t flags_;  ///< Flags bitmask which contains a combination of MessageFlags values.
+    mutable PayloadType payload_; ///< The message payload which can either store some data
+                                  ///< constructed in place, and external handle pointing to the
+                                  ///< place where the actual data is stored or an error value.
+};
+
+static_assert(
+    MiddlewareMessage::MAX_MESSAGE_SIZE == sizeof(MiddlewareMessage),
+    "size of middleware::core::MiddlewareMessage must be the same as cache line alignment!");
```
christian-schilling: How are messages related to cache lines? Why do they need to be the same "alignment"? How do we know what cache lines are like or if there even is a cache on the platform we are on?

libs/bsw/middleware/include/middleware/core/middleware_message.h
```diff
@@ -0,0 +1,472 @@
...
+     * \param flag
+     */
+    constexpr void setFlag_(MessageFlags const flag) { flags_ |= static_cast<uint8_t>(flag); }
+
+    /**
+     * \brief Unsets the \param flag to active in the flags bitmask.
+     *
+     * \param flag
+     */
+    constexpr void unsetFlag_(MessageFlags const& flag) const
```
christian-schilling: Why is a member modified even if the method is declared `const`. How is `mutable` preferable over making the method non const or using `const_cast`?
If there is a good reason, it needs to be documented.

libs/bsw/middleware/include/middleware/core/middleware_message.h
```diff
@@ -0,0 +1,472 @@
...
+        RequestId requestId;
+        InstanceId serviceInstanceId;
+    };
+
+    static constexpr size_t MAX_MESSAGE_SIZE = 32U;
+    static constexpr size_t MAX_PAYLOAD_SIZE = MAX_MESSAGE_SIZE - sizeof(Header)
+                                               - (2 * sizeof(ClusterId)) - sizeof(AddressId)
+                                               - sizeof(uint8_t);
+
+    using InternalBufferType = etl::array<uint8_t, MAX_PAYLOAD_SIZE>;
```
christian-schilling: Why obfuscate the use of array by aliasing the type?

libs/bsw/middleware/include/middleware/core/Message.h
```diff
@@ -0,0 +1,472 @@
...
+     */
+    ExternalHandle const& getExternalHandle_() const { return payload_.externalHandle; }
+
+    /**
+     * \brief Set the payload as an ExternalHandle type which will contain information to where the
+     * actual message's payload is stored.
+     *
+     * \param handle
+     */
+    void setExternalHandle_(ExternalHandle const& handle)
```
christian-schilling: This looks like there is actually an invariant to this class: The unions content must be in agreement with the "external handle" flag.
However: Since there is a way to independently set and unset flags (methods below) the invariant is not really enforced.

libs/bsw/middleware/include/middleware/core/middleware_message.h
christian-schilling: An `unit8_t` in the middle of some other fields. Does this cause padding or not? Would be easy to tell if those types where not obfuscated.

libs/bsw/middleware/test/src/core/middleware_message_unittest.cpp
```diff
@@ -0,0 +1,183 @@
...
+{
+
+class MiddlewareMessageTest : public ::testing::Test
+{
+public:
+    static bool checkMsgHeader(MiddlewareMessage const& lhs, MiddlewareMessage const& rhs)
+    {
+        MiddlewareMessage::Header const& lhsHeader = lhs.getHeader();
+        MiddlewareMessage::Header const& rhsHeader = rhs.getHeader();
+        return lhs.getSourceClusterId() == rhs.getSourceClusterId()
```
christian-schilling: `memcmp`?
It could *maybe* make some sense to create a custom matcher instead of implementing `==`. That would give more useful output when tests fail.

libs/bsw/middleware/include/middleware/core/types.h
filipe-cuim: It makes it easier to know what we're dealing with in the rest of the codebase. You don't need to remember if ServiceId is uint16_t or uint32_t, since it's already defined here.
About the readability of the code, I could agree, but with most editors you can get this information in a pretty obvious way. Also the information is in this types header, whose purpose is just to store these general purpose types.
Also, is there any reason you would want to know what ServiceId really is? Isn't it enough to just know you are dealing with a service identifier? I think having just an integer type actually obfuscates the readibility of the code more than using an alias. With an alias you provide intent and provide some semantics. Why would you want to know if it's a uint16_t or not?

I can remove these, but just wanted to leave my thoughts first.

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: Yes you are right. This can be moved to the global scope. Initially I just did it like this so that I could have the overloading of the global operator== next to the struct definition. What do you think? Should I move it to the outside?

libs/bsw/middleware/include/middleware/core/Message.h
filipe-cuim: Why you think it's wasteful? This type is part of the union, so it doesn't increase the size of it. Or are you referring to the runtime cost of +4 bytes when creating an instance of this type?

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: The mutable here and in the payload_ was because a message can be first constructed as a constant value, but the payload and the flags will always be updated later. I can check if I can remove these mutables, but I need to check in the entire codebase, because it will probably fail somewhere, so I'll take a bit more time.

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: It could be, but I wanted to limit the ways this object is constructed/manipulated. As such, I decided to write it as a class with "factory" methods, that provide more intent to what kind of message is being created.
Every time I see something where you want to use user-defined constructors, I think to use classes instead. Like I mentioned in the discussion, I generally am of the agreement that structs should be used for plain PODs, and as such should not have anything user defined.

But maybe I could place the target cluster, source cluster id and address id inside the header struct. The only reason I didn't do it, is because the current Header should never change after construction of the message. While the rest can change, because you can use a message with the same header, to send it to other destinations.

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: This was an old thing on this code. From what I understood at the time, I think the idea was that the queues, containing these messages will processed very frequently, and as such we wanted to avoid any cache misses.
About platforms that don't have cache, we never thought about that. Is openbsw also thought to platforms like that?

I also thought on an idea where we could have a configuration file (like ETL), where we specify these kind of "runtime" information. There we could specify if there is a cache, and what's the size of the cache. Do you think this would be a useful thing to have?

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: You're right here. I think this was because of some const object in the rest of the api, that was having its flag changed. And thus the method needed to be const (in order to be called for const objects). I'll also review this one.

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: I'm not obfuscating the use of the array, I'm providing information to the outside of the class, of what is its internal buffer type. Maybe it isn't needed. If it isn't I'll remove it.

libs/bsw/middleware/include/middleware/core/Message.h
filipe-cuim: The set and unset flags are private. They are only changed inside the message object.

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: The static assert that we have is currently checking this. So no, there is no padding in the middle.

libs/bsw/middleware/test/src/core/middleware_message_unittest.cpp
filipe-cuim: changed it to a matcher

libs/bsw/middleware/include/middleware/core/types.h
christian-schilling: It's all about context: Sometimes it is important so think about the semantics and values of a piece of data, like "The meaning if this number is a service ID" while in other contexts it is more important to think about it's representation like "these are two bytes I need to store / transmit / whatever". Ideally we can express both of these aspects to both kinds of readers of the code. And usually we can:
Variable *name* -> semantics
Variable *type* -> representation
It is obvious that usually nobody would name a variable of type "int" just "integer", because that is redundant and a missed opportunity to provide extra information by using both the name and the type to convey information to the reader. The same applies in reverse: Don't create a type by the same name as the variable name.

libs/bsw/middleware/include/middleware/core/middleware_message.h
christian-schilling: Couldn't the same be achieved with free factory functions?
I would say placing cluster id and address id into the header does make sense. That would also make the size calculation easier...

libs/bsw/middleware/include/middleware/core/middleware_message.h
christian-schilling: Once again: knowing the representation of the data (in this case array) is an important piece of information depending on context.

libs/bsw/middleware/include/middleware/core/Message.h
christian-schilling: Ah. True. This being in a union makes it less relevant.
Kind of unrelated: If this needs to be a tagged union, why not use `::etl::variant`?

libs/bsw/middleware/include/middleware/core/middleware_message.h
christian-schilling: Yes, I would move it outside.

libs/bsw/middleware/include/middleware/core/middleware_message.h
christian-schilling: About the construction "factory functions":
There is also the approach if using something like a "builder type" that is quite simple and convenient to use. Maybe we can get on a call and show what I mean...

libs/bsw/middleware/include/middleware/core/Message.h
christian-schilling: Ah, right. Didn't see that.

libs/bsw/middleware/include/middleware/core/middleware_message.h
christian-schilling: I know there isn't. I'm just saying it would be nice if that was more obvious.

libs/bsw/middleware/include/middleware/core/middleware_message.h
christian-schilling: I'm not sure that "aligning" the messages with cache lines does really help much with avoiding misses 🤔
I don't think it's that important in most cases to optimise for very specific cache (line) sizes but rather just focusing on data (and code) locality by default and optimising for such target specific parameters only if really needed.
About supporting targets that don't have cache: Yes, currently the only officially supported target is the s32k1, which does not have a (data) cache.
So the question would rather be the other way around: Will we also support targets that do have one?
Yes, we will ;)

libs/bsw/middleware/include/middleware/core/types.h
filipe-cuim: Ok, I can agree with that reasoning actually. Will change this. In a way it's also less words on types.h. I'll also see what the rest of the guys working on this think about this.

libs/bsw/middleware/include/middleware/core/Message.h
filipe-cuim: The problem with etl::variant is that it actually occupies more data than just its types. And this would break the static_assert we have to enforce this object to be of 32 bytes.
Using non-trivial types like etl::variant would also make the hardware crash for these types that will be placed/copied into shared memory, because internal pointers would become invalid if we are not careful during the copy or construction operations.

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: I think I get what you mean. Also we can indeed get on a call. That would be easier to synchronize. Do you use teams?

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: ok thanks for the info. In that case, then the static assert could be removed, and we would remove this constraints about the size of this object. Maybe we could also get in a call to better talk about this.

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: It wasn't needed by any module outside of message, so I deleted this alias.

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: I could successfully remove this const

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: Removed the static assert

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: I could successfully remove the mutables. But to remove the mutable of payload, I had to change memory.h in etl. How to proceed here @rolandreichweinbmw? Can etl inside openbsw deviate from official etl (temporarily of course), while I create the issue on official etl?

rolandreichweinbmw: We forked ETL temporarily in the past for similar reasons, and it looks like it would be very temporarily since we can assume that the const returning get_object_at() will be accepted directly. However, I propose this easier approach, since it prevents the added moving of the ETL source location (via RIM):

* Implement and document a temporary inline get_object_at() in the middleware (header). As a temporary solution.
* In parallel, open a PR with the missing get_object_at() at ETL
* As soon as the const version is available in main ETL (released), switch OpenBSW/Middleware to this new API

This way, we don't need to maintain the separate ETL fork (even though proven to work, and temporarily), which approach would need at least as many steps involved. So keep it more contained, less repos involved.

libs/bsw/middleware/include/middleware/core/Message.h
filipe-cuim: Moved this boolean to be a flag that will be stored inside the message header.

libs/bsw/middleware/include/middleware/core/middleware_message.h
filipe-cuim: I tried out the builder type pattern here, and separating the payload to its own class, but these changes would require some effort on the rest of our code base, and as such I would propose to do this refactor when we have more modules from the code base, committed here, so that we could have a broader view of the whole architectural design.

Also I just wanted to say that I felt that the builder pattern wasn't really tackling our needs here. Because we are not really concerned about a "lazy" initialization of the Message's members. What I was concerned was to create proper and valid instances of this Message object. And these are only the following: a Request or a Response (which need all members to be specified), a Fire and Forget Request or an Event (which don't need a requestId and a addressId) or an Error (which needs all members and a error state).
So technically, whenever a message is created, it should already contain every member initialized. The only exception to this is the Event which can actually initialize it's target cluster id or even change it later.

To conclude, I think maybe the builder pattern can be explored again when we have an idea of the impact in the rest of the code base. What do you think @christian-schilling ?

christian-schilling: I think coming back to this later and with more context makes a lot of sense. 👍🏻

libs/bsw/middleware/include/middleware/queue/QueueBase.h
```diff
@@ -0,0 +1,230 @@
...
+            if (0U == stats_.processedMessages)
+            {
+                ++stats_.startupLoad;
+            }
+        }
+
+        return writableIndex;
+    }
+
+private:
```
christian-schilling, line 222: According to our coding guidelines the convention for members is an `_underscorePrefix` not `suffix_`.

libs/bsw/middleware/include/middleware/core/Message.h
```diff
@@ -0,0 +1,423 @@
...
+            &payload_.externalHandle, ExternalHandle{offset, size});
+    }
+
+    /**
+     * \brief Checks if a flag is currently active in the flags bitmask.
+     *
+     * \param flag
+     * \return true if \param flag is active, otherwise false.
+     */
+    constexpr bool hasActiveFlag_(Flags const flag) const
```
christian-schilling: We don't put underscore pre or suffix on methods.

PR #228, opened 2025-10-09
Remove lots of trivial getters/setters

christian-schilling:
Data members that have a trivial setter have no invariant and thus can be public.

Change-Id: I66fb2ba1428ce2279910c6b219ade68f6c4feb60


libs/bsw/uds/include/uds/IDiagDispatcher.h
```diff
@@ -66,8 +58,9 @@ class IDiagDispatcher
 
 protected:
     DiagJobRoot& fDiagJobRoot;
-    IDiagSessionManager& fSessionManager;
 
+public:
+    IDiagSessionManager& fSessionManager;
     bool fEnabled;
```
matthiaskessler, line 64: I think, that we should rename the public fields because IMHO, now the code reads as if maybe a `private` was forgotten.
I would suggest to have `private` and `protected` fields still prefixed put public fields should not be prefixed, e.g. `enabled` instead of `fEnabled` here.

christian-schilling: I agree. I have a stack of 7 more changes already prepared on top of this and would prefer to fix this on top of the stack rather than going though all the rebase conflicts that will happen if I change it in this PR...

christian-schilling: I already started working on a new change to fix all the names. To be submitted separately.

matthiaskessler: OK for me.

PR #229, opened 2025-10-09
Add S32K148 build with clang in the CI

rolandreichweinbmw:
Also, add missing C++ versions

### DIFF
--- a/.ci/buildoperations.py
+++ b/.ci/buildoperations.py
@@ -129,6 +129,8 @@ class BuildOpTpl:
 
                 build_cmd_str += f" --config {str(config)}"
 
+        build_cmd_str += " --verbose"
+
         return build_cmd_str
 
     def get_full_test_command(self):


.github/workflows/build.yml
```diff
@@ -7,11 +7,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        preset: [ "tests-debug", "tests-release", "posix", "posix-with-tracing", "s32k148-gcc" ]
+        preset: [ "tests-debug", "tests-release", "posix", "posix-with-tracing", "s32k148-gcc", "s32k148-clang" ]
         platform: [ "arm", "linux" ]
         config: [ "Debug", "Release", "RelWithDebInfo" ]
         cxxid: ["gcc", "clang"]
-        cxxstd: [ 14, 20 ]
+        cxxstd: [ 14, 17, 20, 23 ]
```
johannes-esr, line 14: This was explicitly removed in #200, therefore I think we have to discuss this again.

rolandreichweinbmw: There was no further explanation attached to the reduced number of tested C++ versions.

If the reasoning was to just reduce the number of tests, I would oppose this decision.

You also wouldn't remove unit test, just because their number is already high, wouldn't you? ;-)

If it was because of economic reasons: The tests are relatively fast and cheap: The majority of time is still spent in unit tests and coverage.

If we support C++ up to 23 (at least forward compatible), we should test it to prevent surprises later on.

I think it is good to test all those combinations.

Finally, C++23 is the current standard and I'm also interested in C++17 because it is the standard supported by MISRA C++23.

aschaal: The problem is that we compile the stuff now with again almost 100 builds (and the number will grow in future, that is clear).
Building all possible combinations will not add any value in practice, it's just a waste of energy.

We can agree that our code must compile with also 17 and 23, which makes totally sense.
But we should not build every possible combination.

We could add some (nightly, weekly) build with all combinations and with even not yet supported compilers etc. to get a full overview of the current state, but not for every "git push"...

rolandreichweinbmw: Basically, the most important combinations are {14, 17, 20, 23} x {gcc, clang} x {s32K148, posix) + Unit Tests / Coverage for one of them

This makes only 17. And most of them are really cheap in terms of time and energy, compared to the Coverage one.

As I understand, most of the other combinations are redundant - maybe we can reduce those.

aschaal: Okay, indeed.
Most time is wasted in building the same docker container multiple times. Needs to be fixed in a another PR.

So I can accept this change.

aschaal (approved):

PR #230, opened 2025-10-10
Fix s32 mpu global vars

jiangyilism:
*** Avoid initializing global vars of libsafe*.a before mpu is enabled

linker script already asserts that .data section of libsafe*.a is empty.
However, .rodata might be merged into .text as a compiler optimization.
So those .rodata vars are initialized before mpu being enabled.

This PR extends the assertion to check .rodata . And move the global
vars initialization to MemoryProtection::init() .

### DIFF
--- a/doc/features/functional_safety.rst
+++ b/doc/features/functional_safety.rst
@@ -130,8 +130,13 @@ Memory Protection Unit
 ++++++++++++++++++++++
 
 The memory regions are project specific. In ReferenceApp, the initialized safety relevant data is
-mapped to .mpu_data section and uninitialized data relevant to safety is mapped to .mpu_bss section
-in the linkerscript. The memory is divided into several sections, which are configured by the MPU.
+mapped to .mpu_data/.mpu_rodata section and uninitialized data relevant to safety is mapped
+to .mpu_bss section in the linkerscript.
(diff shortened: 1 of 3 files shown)


cmake/toolchains/ArmNoneEabi-gcc.cmake
```diff
@@ -12,8 +12,18 @@ set(_EXE_LINKER_FLAGS "-specs=nano.specs \
 
 include("${CMAKE_CURRENT_LIST_DIR}/ArmNoneEabi.cmake")
 
-set(CMAKE_C_COMPILER arm-none-eabi-gcc)
-set(CMAKE_CXX_COMPILER arm-none-eabi-g++)
-set(CMAKE_ASM_COMPILER arm-none-eabi-gcc)
-set(CMAKE_LINKER arm-none-eabi-g++)
-set(CMAKE_AR arm-none-eabi-ar)
+if (NOT DEFINED CMAKE_C_COMPILER)
```
aschaal: This change is unrelated to MPU. Can you please move it out to an extra PR?

jiangyilism: Removed. Thanks.

PR #231, opened 2025-10-13
Split unit tests into generic and platform specific parts

aschaal:
Also remove the unused workflow presets.

### DIFF
--- a/platforms/s32k1xx/bsp/canflex2Transceiver/test/CMakeLists.txt
+++ b/platforms/s32k1xx/bsp/canflex2Transceiver/test/CMakeLists.txt
@@ -1,8 +1,5 @@
 add_executable(CanFlex2TransceiverTest src/can/CanFlex2TransceiverTest.cpp)
 
-target_include_directories(CanFlex2TransceiverTest PRIVATE ../include
-                                                           mock/include)
-
 target_link_libraries(
     CanFlex2TransceiverTest
     PRIVATE asyncMockImpl


doc/learning/unit_tests/index.rst
```diff
@@ -46,4 +58,10 @@ Run the tests:
 
     ctest --preset tests-debug --parallel
 
+If you modified some CMakeLists.txt files in the project don't forget to run:
+
+.. code-block:: bash
+
+   cmake-format -i $(find . -name CMakeLists.txt | sed '/3rdparty\/.*\/CMakeLists\.txt/d')
```
rolandreichweinbmw, line 61: Maybe we need to run that on the repo brefore, since this currently changes ~30 files as checked in in main?

aschaal: Yes, indeed! Many files are not formatted correctly. We should do that in another PR to avoid confusion between formatting and content changes, especially when changing CMake and CI content.

Will do after this PR is merged.

CMakePresets.json
```diff
@@ -11,26 +11,80 @@
...
             }
         },
         {
             "name": "tests-release",
-            "displayName": "Configure for testing",
-            "description": "Configure for testing",
+            "displayName": "Configure for testing generic modules (release)",
+            "description": "Configure for testing generic modules (release)",
             "inherits": "_config-base",
             "binaryDir": "${sourceDir}/build/tests/Release",
```
vladyslavmarkovaccenture: `build/tests/generic/Release`?

CMakePresets.json
```diff
@@ -11,26 +11,80 @@
     "configurePresets": [
         {
             "name": "tests-debug",
-            "displayName": "Configure for testing",
-            "description": "Configure for testing",
+            "displayName": "Configure for testing generic modules (debug)",
+            "description": "Configure for testing generic modules (debug)",
             "inherits": "_config-base",
             "binaryDir": "${sourceDir}/build/tests/Debug",
```
vladyslavmarkovaccenture: `build/tests/generic/Debug`?

doc/learning/unit_tests/index.rst
rolandreichweinbmw: treefmt, which we are already recommending for formatting source code before merge, is already configured for CMakeFiles.txt

However, the exclusion of *.txt is preventing it.

From a first look, we could just remove this "*.txt" exclusion from .treefmt.toml? Then, we don't need the explicit prescription above for cmake-format anymore?

CMakePresets.json
aschaal: Yes, changed that and harmonized also the other UT targets, so the names match the build folders now.

CMakePresets.json
aschaal: Yes, changed that and harmonized also the other UT targets, so the names match the build folders now.

doc/learning/unit_tests/index.rst
aschaal: Yes! That makes sense.
I just keep it here in this commit as I only moved the UT description.

Will remove it after this has been merged in a separate PR just for this.

.github/workflows/build.yml
```diff
@@ -87,7 +74,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: build
-          key: ${{ runner.os }}-cmake-${{ matrix.platform }}-${{ matrix.cxxid }}-${{ matrix.cxxstd }}-${{ hashFiles('**/*.cpp', '**/*.h',  '**/*.cmake', '**/*.txt', '**/*.c', '**/*.s') }}
+          key: ${{ runner.os }}-cmake-${{ matrix.preset }}-${{ matrix.config }}-${{ matrix.platform }}-${{ matrix.cxxid }}-${{ matrix.cxxstd }}-${{ hashFiles('**/*.cpp', '**/*.h',  '**/*.cmake', '**/*.txt', '**/*.c', '**/*.s') }}
           restore-keys: |
             ${{ runner.os }}-cmake-${{ matrix.platform }}-${{ matrix.cxxid }}-${{ matrix.cxxstd }}-
```
vladyslavmarkovaccenture: just curious, don't we need config and preset here too?

aschaal: Done

PR #240, opened 2025-10-16
Fix usage of CMAKE_CXX_STANDARD

rolandreichweinbmw:
When building unit tests, the specified CMAKE_CXX_STANDARD is not used as expected via a -std= command line argument when specifying e.g. -DCMAKE_CXX_STANDARD=17 on the cmake configure run.


rolandreichweinbmw:
I also updated the ARM GCC toolchain from 10.3-2021.10 to 14.3.rel1 since this issue uncovered the fact that the old toolchain didn't support C++23.

PR #242, opened 2025-10-16
Replace est_assert by ETL_ASSERT

### LINKED ISSUE #176: Consolidation of assert from ESTD to ETL

christophruethingbmw:
In the context of migrating to ETL we introduced a new `ETL_ERROR_GENERIC` macro to ETL to be able to easily place asserts without the need of specifying a separate exception class for each assert. The `ETL_ERROR_GENERIC` simply constructs a generic exception with the given text. Depending on the ETL configuration, this exception is thrown, passed to a custom assert handler or ignored at all.

For unit tests we configure ETL now to throw exceptions. For the target we implemented a custom assert handler which is more or less a copy of the existing handler in estd.

fixes #176

> [!NOTE]
> I also had to do a change in the timer implementation which otherwise always runs into an assert now (since we enabled asserts unit tests now using exceptions). The issue was in test `the_same_timeout_is_added_multiple_times` which tries to add the same timeout object a second time. The implementation then calls `_timeoutList.insert_after` with an already inserted object which then asserts. I don't think we want this, therefore I added an additional check to not insert.

### DIFF
--- a/libs/bsw/timer/test/src/TimerTest.cpp
+++ b/libs/bsw/timer/test/src/TimerTest.cpp
@@ -306,16 +306,19 @@ TEST_F(
 
 TEST_F(TimerTest, the_same_timeout_is_added_multiple_times)
 {
-    // forward_list handles this issue with is_in_use()
-    // it doesn't allow to add same node twice
+    // intrusive_forward_list will assert if the same noe is added twice.
+    // The implementation of Timer::set() has ta handle this and prevent
+    // a Timer instance from being added twice.
 
     setAndExpectTrigger(fTimer, fTimeoutMock1, 100);
     EXPECT_CALL(fAlarmMock, setAlarm(100));
     update(fTimer, fNow);
     fTimer.set(fTimeoutMock1, 100, fNow);
-    fTimer.set(fTimeoutMock1, 100, fNow);
+    // Pass different values, to make sure they are not used.
+    fTimer.set(fTimeoutMock1, 200, fNow);
 
     EXPECT_CALL(fTimeoutMock1, expired());
+    // Expire first passed value and make sure it's still used.
     fNow = 100U;
     update(fTimer, fNow);
 }


rolandreichweinbmw (approved):

executables/referenceApp/application/src/main.cpp
```diff
@@ -2,4 +2,33 @@
...
+#include <execinfo.h>
+#endif
+
+void etl_assert_function(etl::exception const& exception)
+{
+#if __has_include("execinfo.h")
+    void* callstack[128];
+    backtrace_symbols_fd(callstack, backtrace(callstack, 128), 2 /* = stderr */);
+#endif
+    std::cout << "Assertion at " << exception.file_name() << ':' << exception.line_number() << " ("
```
matthiaskessler: I think, we should use `printf` instead. We are also working on a change to replace the runtime's `printf` by the one from our 3rdparty lib. This is because the behavior of `printf` and I assume `std::cout` is platform and toolchain dependent. We have seen in the past implementation that use `malloc` internally and that's actually a no-go. So long story short, I think with using `std::cout` we would be adding one more thing that would potentially use dynamic memory allocation. And that might fail, so you would have an exception within the exception.

executables/referenceApp/application/src/main.cpp
```diff
@@ -2,4 +2,33 @@
...
+              << exception.what() << ") failed" << std::endl;
+    std::abort();
```
matthiaskessler: If we call `std::abort()` we need to make sure, that we do a system reset for the embedded targets at the end.
I think, `abort` will call `exit()` so we would need to register an exit handler using `atexit()`

executables/referenceApp/application/src/main.cpp
```diff
@@ -2,4 +2,33 @@
...
+}
+#else
+#include <cstdlib>
+
+void etl_assert_function(etl::exception const&) { std::abort(); }
+#endif
+
+void app_main()
+{
+    etl::set_assert_function(etl_assert_function);
```
matthiaskessler, line 32: That might be too late. What about asserts that happen during execution of constructors of static objects?

executables/referenceApp/application/src/systems/EthernetSystem.cpp
```diff
@@ -8,19 +8,22 @@
...
-#include <util/estd/assert.h>
+
+#include <etl/error_handler.h>
 
 extern "C"
 {
 int32_t vlanForNetif(void const* const vlwipNi)
 {
-    estd_assert(vlwipNi != nullptr);
+    ETL_ASSERT(vlwipNi != nullptr, ETL_ERROR_GENERIC("netif must not be null"));
```
matthiaskessler, line 18: As much as I like the text which tells you more about the error, as much I am concerned about the possible additional ROM usage for the strings. We have faced this issue in the past in many projects and came up with an implementation, that didn't store the strings at all. AFAIK, this is currently not possible in ETL. It has `ETL_VERBOSE_ERRORS` which lets you control if you want `__FILE__` to be expanded or not, but the error text is, I think, always there.

libs/bsw/cpp2can/include/can/canframes/CANFrame.h
```diff
@@ -178,7 +178,9 @@ inline uint8_t const& CANFrame::operator[](uint8_t const pos) const { return _pa
 
 inline void CANFrame::setPayload(uint8_t const* const payload, uint8_t const length)
 {
-    estd_assert(length <= MAX_FRAME_LENGTH);
+    ETL_ASSERT(
+        length <= MAX_FRAME_LENGTH,
+        ETL_ERROR_GENERIC("can frame length must be smaller than maximum length"));
```
matthiaskessler: nit: I'd prefer "CAN" instead of "can" in this to avoid confusion with the verb.

executables/referenceApp/application/src/main.cpp
```diff
@@ -2,4 +2,33 @@
...
+    backtrace_symbols_fd(callstack, backtrace(callstack, 128), 2 /* = stderr */);
+#endif
+    std::cout << "Assertion at " << exception.file_name() << ':' << exception.line_number() << " ("
+              << exception.what() << ") failed" << std::endl;
+    std::abort();
+}
+#else
+#include <cstdlib>
+
+void etl_assert_function(etl::exception const&) { std::abort(); }
```
matthiaskessler, line 27: I think, we should implement a referenceApp specific handler in a later commit that doesn't use the default of `std::abort()`.

executables/referenceApp/application/src/main.cpp
matthiaskessler: We can also do this in a later commit. We might need an extra hook in our startup code before `__exec_ctors` is called.

libs/bsw/util/CMakeLists.txt
```diff
@@ -43,9 +42,7 @@ target_include_directories(util PUBLIC include)
 target_link_libraries(util PUBLIC etl platform)
 
 if (BUILD_UNIT_TESTS)
-    target_compile_definitions(
-        util PUBLIC ESTD_HAS_EXCEPTIONS ESTL_ASSERT_MESSAGE_ALL
-                    ETL_THROW_EXCEPTIONS)
+    target_compile_definitions(util PUBLIC ETL_THROW_EXCEPTIONS)
```
matthiaskessler: I think, that the change in `etl_profile.h` should cover this already.

christophruethingbmw: I agree, removed it

executables/referenceApp/application/src/main.cpp
christophruethingbmw: I agree, as discussed, let's think about how we want to do this in a follow up.

executables/referenceApp/application/src/main.cpp
christophruethingbmw: Fully agree, I also thought about this but thought that for now this is the easies place. I think we should rework this together with the discussion raised from below (i.e. how we generally handle asserts and other kind of errors).

executables/referenceApp/application/src/systems/EthernetSystem.cpp
christophruethingbmw: I agree, maybe we should remove the text in case `ETL_VERBOSE_ERRORS` is not set, like we also do it for `__FILE__`. I will prepare a change in ETL to accomplish this.

matthiaskessler (approved):

Issue #195, opened 2025-09-24
Document usage of `rim` and integrate into CI


johannes-esr:
For know, we want to use `rim` to handle third party dependencies in OpenBSW. This has to be documented in the documentation. In addition, a check in the CI shall be enabled to check if third party dependencies are modified.

PR #245, opened 2025-10-17
Document usage of rim and integrate into CI

rolandreichweinbmw:
This fixes https://github.com/eclipse-openbsw/openbsw/issues/195

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -13,6 +13,7 @@ RUN for i in 1 2 3; do \
             ninja-build \
             python3-pip \
             python3.10 \
+            ruby-rubygems \
             tar \
             tmux \
             wget \
@@ -58,8 +59,10 @@ RUN wget https://github.com/mozilla/sccache/releases/download/v0.10.0/sccache-v0
     && rm sccache-v0.10.0-x86_64-unknown-linux-musl.tar.gz \
     && rm -rf sccache-v0.10.0-x86_64-unknown-linux-musl
 
+RUN gem install esr-rim
+
 RUN git config --global --add safe.directory /home/jenkins
-ENV PATH="/usr/bin/gcc-arm-none-eabi/bin:/usr/bin/llvm/bin:/usr/bin/llvm-arm/bin:usr/bin/clang-format-17:usr/bin/treefmt:/usr/bin/gcc:/usr/local/bin/sccache${PATH}"
+ENV PATH="/usr/bin/gcc-arm-none-eabi/bin:/usr/bin/llvm/bin:/usr/bin/llvm-arm/bin:${PATH}"
 
 RUN echo 'export HISTFILE=/root/.bash_history/history' >> /root/.bashrc
 
--- a/.github/workflows/rim.yml
+++ b/.github/workflows/rim.yml
@@ -1,0 +1,18 @@
+name: Third party libraries check
+
+on: [ merge_group, push, pull_request ]
+
+jobs:
+  rim:
+    runs-on: ubuntu-latest
+    steps:
+     - name: Checkout code
+       uses: actions/checkout@v4
+
+     - name: Build the docker image
+       run: |
+          docker build -f docker/Dockerfile.dev -t dev .
+
+     - name: Run the third party libraries check via RIM inside the docker container
+       run: |
+          docker run --rm -v "$PWD:/home/jenkins" -w /home/jenkins --user $(id -u):4996 dev rim status -d --verify-clean


doc/guidelines/3rdparty.rst
```diff
@@ -0,0 +1,43 @@
...
+
+* Making the source code of the included libraries available self-contained in
+  OpenBSW, to make OpenBSW independent of external download locations
+* Referencing the remote location of the original source
+* Keeping the maintained libraries in sync with the original location, with
+  support for updates and checking integrity
+
+As an introduction, some important features are demonstrated here:
+
+Set up new external library
```
aschaal: Please use title case

doc/guidelines/3rdparty.rst
```diff
@@ -0,0 +1,43 @@
...
+    rim sync -u https://github.com/ETLCPP/etl.git -r 20.44.0 -m "Update to ETL 20.44.0 in OpenBSW" libs/3rdparty/etl
+
+Check status of third libraries
+-------------------------------
+
+.. code-block:: bash
+
+    rim status -d --verify-clean
+
+Checks all third party libraries for deviations from original source and returns 1 if dirty
```
aschaal: End sentence with ".".

aschaal: Maybe shortly explain what "dirty" means.

doc/guidelines/3rdparty.rst
```diff
@@ -0,0 +1,43 @@
+Third Party Libraries
+=====================
+
+Third party libraries under ``libs/3rdparty`` are maintained by
```
aschaal: It is recommended if possible.
Some 3rdparty code may not be available via (public) git repo or we want just a few files.

-->  Explain also the fallback.

doc/guidelines/3rdparty.rst
```diff
@@ -0,0 +1,43 @@
+Third Party Libraries
```
aschaal, line 1: In general there is much more to do, e.g.:

- check license
- add "SBOM" data (we need to discuss what exactly, but there are somethings which are useful

Maybe briefly mention these points with a "to be clarified later" comment?

docker/Dockerfile.dev
```diff
@@ -58,8 +59,10 @@ RUN wget https://github.com/mozilla/sccache/releases/download/v0.10.0/sccache-v0
     && rm sccache-v0.10.0-x86_64-unknown-linux-musl.tar.gz \
     && rm -rf sccache-v0.10.0-x86_64-unknown-linux-musl
 
+RUN gem install esr-rim
+
 RUN git config --global --add safe.directory /home/jenkins
-ENV PATH="/usr/bin/gcc-arm-none-eabi/bin:/usr/bin/llvm/bin:/usr/bin/llvm-arm/bin:usr/bin/clang-format-17:usr/bin/treefmt:/usr/bin/gcc:/usr/local/bin/sccache${PATH}"
```
aschaal, line 62: I cannot judge this deletion by just looking at this line.
Why is that not needed anymore?

doc/guidelines/3rdparty.rst
aschaal: We have several 3rdparty folders, e.g. for platform specific libs.

The rule must be, that every third party FILE must have "3rdparty" in the path.

doc/guidelines/3rdparty.rst
rolandreichweinbmw: Right. I added a description how to do that in the project's root NOTICE.md.

docker/Dockerfile.dev
rolandreichweinbmw: * usr/bin/clang-format-17 : missing leading slash, and also PATH should contain directories only, not full paths to executable files, i.e., was broken. Worked anyway, because /usr/bin is included in PATH by default
* usr/bin/treefmt: Same as above
* /usr/bin/gcc: Same as above, regarding the full file path
* /usr/local/bin/sccache${PATH}: Not properly separated with ":" from the initial ${PATH} value. Worked anyway, since we have /usr/local/bin in the PATH anyway, and the first member of the initial PATH value was /usr/bin/sbin which was unused until now

So all those entries can safely be removed.

doc/guidelines/3rdparty.rst
rolandreichweinbmw: I documented that now, further down.

rolandreichweinbmw:
Thank you @aschaal for reviewing! I addressed all the points and force pushed an update to this branch.

PR #246, opened 2025-10-20
Adjust CMAKE_MODULE_PATH in root CMakeLists.txt

filipe-cuim:
## Description

When using the vscode cmake extension, I noticed that there was a problem with the inclusion of the CodeCoverage module. From what I understood, this because of the order that vscode handles paths to the cmake extension, when using cmake presets.
My idea here was to change the CMAKE_MODULE_PATH to the base cmake preset.


vladyslavmarkovaccenture:
In fact, I'd vote for creating a public repo that mimics the client and uses openbsw as a thirdparty. Such repo could help to both understand how to use openbsw as a thirdparty (which must be kept intuitive), and if changes in CMake can break client's builds.

filipe-cuim:
> I don't think it should be merged, because CMakePresets is technically optional. Build can work without it. CMakePresets is a layer above the usual way of building made for convenience. It also made for cases when we build openbsw and not a client superproject that uses openbsw as a thirdparty. If this change is merged (hypothetically), client will have to change his own way of building his project to mimic the module path addition, which doesn't sound good in any way. If VS Code has problems with building the project the way it is, then there should be other ways to fix it. CMakePresets were not added because of VS Code. They were added to make it easier to build project from a command line. I will bring up the topic of creating additional step for CI to test how well openbsw integrates into client projects as a thirdparty repository - that should help to recognize the issue we can potentially add for the client when modifying CMake files.

Ok, that's a good point. I actually thought that the presets was the official way of configuring and building the project.
But in that case, should we not use something in the top level CMake to make all paths relative to the source dir? Either using CMAKE_SOURCE_DIR, CMAKE_CURRENT_SOURCE_DIR or CMAKE_CURRENT_LIST_DIR? I tried this locally, and it also worked.

CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ project(
         "BSW workspace with reference application for multiple platforms"
     LANGUAGES CXX C ASM)
 
-list(APPEND CMAKE_MODULE_PATH "cmake/modules")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/modules")
```
vladyslavmarkovaccenture: Please, change to `CMAKE_CURRENT_SOURCE_DIR`

filipe-cuim: From your explanation makes sense. Do you think I should also change the other CMAKE_CURRENT_LIST_DIR in this file? Or in those cases it is correct?

vladislavmarkov: yes, in this file it makes sense to use `CMAKE_CURRENT_SOURCE_DIR` instead of `_LIST_DIR`.

PR #248, opened 2025-10-21
Make DiagDispatcher::dispatchIncomingRequest a static

christian-schilling:
This is an intermediate refactoring step indended to make it clear what data is accessed by what operations.

Change-Id: I6741834f19d43e2a7b3c7782606521e6f885ba01

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,8 @@ add_subdirectory(libs EXCLUDE_FROM_ALL)
 if (BUILD_UNIT_TESTS)
 
     add_subdirectory(executables/unitTest EXCLUDE_FROM_ALL)
-    target_include_directories(etl BEFORE INTERFACE executables/unitTest/etl_profile)
+    target_include_directories(etl BEFORE
+                               INTERFACE executables/unitTest/etl_profile)
 
     if (OPENBSW_PLATFORM STREQUAL "posix")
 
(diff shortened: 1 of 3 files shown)


libs/bsw/uds/include/uds/DiagDispatcher.h
```diff
@@ -107,6 +109,9 @@ class DiagDispatcher
     uint8_t dispatchTriggerEventRequest(transport::TransportMessage& tmsg) override;
 
 private:
+    using SendBusyResponseCallback
```
christian-schilling, line 112: Note that this callback is a temporary measure that will definitely be removed again.

CMakeLists.txt
```diff
@@ -97,7 +97,8 @@ add_subdirectory(libs EXCLUDE_FROM_ALL)
 if (BUILD_UNIT_TESTS)
 
     add_subdirectory(executables/unitTest EXCLUDE_FROM_ALL)
-    target_include_directories(etl BEFORE INTERFACE executables/unitTest/etl_profile)
+    target_include_directories(etl BEFORE
+                               INTERFACE executables/unitTest/etl_profile)
```
rolandreichweinbmw, line 102: Maybe we need to include CMakeLists.txt (in the root folder) explicitly to .treefmt.toml, since currently only the ones in subdirs are operated on via treefmt?

PR #249, opened 2025-10-23
Configure clang-tidy

filipe-cuim:
## Description

The goal of this PR is to configure a .clang-tidy that respects openbsw coding conventions, and a ci job that triggers clang-tidy checks on files changed.

### DIFF
--- a/.ci/clang-tidy.py
+++ b/.ci/clang-tidy.py
@@ -3,8 +3,8 @@ Clang-tidy Checker
 ==================
 
 This script runs the llvm helper tool clang-tidy-diff on all commits
-from the current branch until origin/main. It receives as argument
-the build directory where compile_commands.json is located.
+from the current branch's HEAD to the common commit between this branch and origin/main.
+It receives as argument the build directory where compile_commands.json is located.
 Before running the clang-tidy-diff tool, it creates a temporary cleaned
 version of compile_commands.json without flags that clang does not
 recognize. If any clang-tidy warning is found in the changed lines,
@@ -55,8 +55,17 @@ def check_clang_tidy(build_dir: str) -> str:
     Runs clang-tidy-diff from llvm, on the git diff and returns stderr output.
     """
 
+    # First, get the merge base
+    merge_base = subprocess.run(
+        ["git", "merge-base", "HEAD", "origin/main"],
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    base_commit = merge_base.stdout.strip()
+
     git_diff = subprocess.run(
-        ["git", "diff", "-U0", "origin/main..HEAD"],
+        ["git", "diff", "-U0", f"{base_commit}...HEAD"],
         capture_output=True,
         text=True,
         check=True,
@@ -90,7 +99,7 @@ if __name__ == "__main__":
 
         clang_tidy_stderr = check_clang_tidy(TMP_BUILD_DIR.as_posix())
         shutil.rmtree(TMP_BUILD_DIR)
-        if re.search(r"warning treated as error", clang_tidy_stderr, re.MULTILINE):
+        if re.search(r"warnings? treated as errors?", clang_tidy_stderr, re.MULTILINE):
             print("\nClang-Tidy Warnings Detected", file=sys.stderr)
             sys.exit(1)
         else:


.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,24 @@
...
+     - name: Checkout code
+       uses: actions/checkout@v4
+       with:
+         fetch-depth: 0
+
+     - name: Build the docker image
+       run: |
+          docker build -f docker/Dockerfile.dev -t dev .
+
+     - name: Run the posix debug build inside the docker container
```
simon-d-bmw: need another run for s32k1.

Also: check if clang-tidy build for reference application is feasible (if it takes too long - could it be done in post commit? or is there a "pareto" option?)

.ci/clang-tidy.py
```diff
@@ -0,0 +1,73 @@
...
+        sys.exit(1)
+    try:
+        BUILD_DIR = Path(sys.argv[1])
+        TMP_BUILD_DIR = BUILD_DIR / "tmp"
+        TMP_BUILD_DIR.mkdir(exist_ok=True)
+        clean_compile_commands(BUILD_DIR, TMP_BUILD_DIR)
+        clang_tidy_stderr = check_clang_tidy(TMP_BUILD_DIR.as_posix())
+        shutil.rmtree(TMP_BUILD_DIR)
+        if re.search(r"warning treated as error", clang_tidy_stderr, re.MULTILINE):
+            print("\nClang-Tidy Warnings Detected!!!", file=sys.stderr)
```
simon-d-bmw:
```suggestion
            print("\nclang-tidy warnings detected", file=sys.stderr)
```

.ci/clang-tidy.py
simon-d-bmw, line 1: please add brief description what the script does.

.github/workflows/clang-tidy.yml
filipe-cuim: Why do a run for s32k1?

simon-d-bmw: I see that there already is a build for s32k1, maybe we overlooked it previously.
But still, now only the **tests** are checked with clang-tidy. The reference app should be checked, too. 
That would be presets `"posix-freertos", "posix-threadx", "s32k148-freertos-clang", "s32k148-threadx-clang" `

simon-d-bmw:
> If I get the change right, only clang-tidy warnings are kept, which belong to lines changed in a PR. What happens to warnings which are occuring on different lines, but are introduced due to the PR?

@johannes-esr the approach chosen here would not detect new violations outside of the diff. 
So this method is not 100% ironclad, i.e. allows such warnings to be merged undetected. However, I expect this to be a rather rare case, and it can still be detected later on in nightly builds.

hbragge-acn:
@filipe-cuim Hi, I like this idea but I don't think it's realistic to enforce zero findings with this list of checks, many of which are rather informational and some prone to false positives too. Problem is that the diff can contain unrelated findings that happen to be located on the same line, maybe preventing you from doing a bugfix (eg. change the value of a variable) without a bunch of unrelated changes as well (eg. changing the name of that variable everywhere). Or being forced to change your code due to an informational finding (say, "performance-unnecessary-value-param").

Perhaps we should limit this to a smaller number of checks, and make sure that all existing findings are fixed before enforcing it?

simon-d-bmw:
> @simon-d-bmw @johannes-esr would you be fine if we improve this in a later PR, and do the full checking of the repo later, and for now keep just the checking on the diff? Or would you prefer to do the checking of the full PR now?

I would prefer to merge it as is.

filipe-cuim:
> > @hbragge-acn Maybe you are right, would you have a suggestion on what I should have enabled? Also, the readability clang-tidys I think are important, because they enforce the coding conventions of openbsw. I think I have aligned most of them with what is written in the documentation (I may have missed them though), but I think we should keep them all. If you have an important thing to merge, which clang-tidy is blocking you can always add NOLINTBEGIN and NOLINTEND to the start and end of the regions where you don't want to clang-tidy to run. I think it's preferable to do that (along with a good justification) then not having the checks enabled. What do you think?
> 
> I’m not insisting on excluding any specific check, my general point is just that it would be annoying to have a PR blocked by a clang-tidy finding that the PR didn't even introduce. That’s why I’m suggesting we ensure that there are only very few or ideally zero existing findings before merging this PR.
> 
> However the number of findings is probably too large right now to fix or suppress them all up front, so a more realistic option would be to start by enforcing only a small number of checks, and later add more as we fix the existing findings for each new check.

I'm afraid that even doing that will probably take a long time, and would require a continuous work by someone to follow all of those checks, and fix the warnings for each clang-tidy check. Maybe this could be discussed in the next openbsw round. Because some of the guys here, are more of the opinion to merge this current approach, and then try to improve on it (moving towards a solution that provides a full report of the repository and just checks if any new warnings were added on a PR).

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,60 @@
+name: Clang-tidy check
+
+on: [ merge_group, push, pull_request ]
+
+jobs:
+  clang-tidy:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        preset: ["tests-posix-debug", "tests-posix-release", "tests-s32k1xx-debug", "tests-s32k1xx-release"]
```
simon-d-bmw, line 10: Include the reference application? (will need some more pararmeters)
```suggestion
        preset: ["tests-posix-debug", "tests-posix-release", "tests-s32k1xx-debug", "tests-s32k1xx-release", "posix-freertos", "s32k148-freertos-clang"]
```

doc/dev/guidelines/practices.rst
```diff
@@ -7,6 +7,29 @@ secure coding practices.
...
+
+.. code-block:: bash
+
+  git diff -U0 upstream/main...HEAD | clang-tidy-diff -p1 -use-color -path Directory_to_cleaned_compile_commands
+
+or, in case you want to run clang-tidy for all lines and not only the changed ones:
+
+.. code-block:: bash
+
+  clang-tidy --use-color -p Directory_to_cleaned_compile_commands Path_to_file_or_directory
```
simon-d-bmw, line 30: Only this works in the Docker container
```suggestion
  clang-tidy-17 --use-color -p Directory_to_cleaned_compile_commands Path_to_file_or_directory
```

.ci/clang-tidy.py
```diff
@@ -0,0 +1,110 @@
...
+            json.dump(cleaned_data, cf, indent=2)
+
+
+def check_clang_tidy(build_dir: str) -> str:
+    """
+    Runs clang-tidy-diff from llvm, on the git diff and returns stderr output.
+    """
+
+    git_diff = subprocess.run(
+        ["git", "diff", "-U0", "origin/main..HEAD"],
```
hbragge-acn: How about the case when origin/main has already moved on and the PR branch is no longer based on it?

.ci/clang-tidy.py
```diff
@@ -0,0 +1,110 @@
...
+        BUILD_DIR = Path(sys.argv[1])
+
+        # Create a temporary directory for the cleaned compile commands
+        TMP_BUILD_DIR = BUILD_DIR / "tmp"
+        TMP_BUILD_DIR.mkdir(exist_ok=True)
+        clean_compile_commands(BUILD_DIR, TMP_BUILD_DIR)
+
+        clang_tidy_stderr = check_clang_tidy(TMP_BUILD_DIR.as_posix())
+        shutil.rmtree(TMP_BUILD_DIR)
+        if re.search(r"warning treated as error", clang_tidy_stderr, re.MULTILINE):
```
hbragge-acn: This only works when there's just one error, but not with multiple errors (eg. "2 warnings treated as errors")

.ci/clang-tidy.py
```diff
@@ -0,0 +1,110 @@
+"""
+Clang-tidy Checker
+==================
+
+This script runs the llvm helper tool clang-tidy-diff on all commits
```
christian-schilling, line 5: "commits until origin/main" reads like "ancestors of origin/main".
Maybe "based on origin/main" or "since origin/main"?

Apart from that: I can't find any logic in the code that selects what commits to run on 🤔

filipe-cuim: You're right. Will adapt accordingly

.ci/clang-tidy.py
filipe-cuim: You are right. I need to double check this. My original idea here was to check for at least one finding. If there is at least one finding, we exist the script with 1, thus failing the job. The overall report from clang-tidy-diff is printed to stdout and stderr anyway, so we would see all the reports in the CI. But I forgot that if there is more warnings clang-tidy writes the string in plural.

.ci/clang-tidy.py
filipe-cuim: I suppose that what matters for the CI is PR branches that want to be merged in main? So this shouldn't be a problem I think?
The double dot operator here, means "all commits that are accessible from HEAD but not from origin/main". So what it actually does it to list all commits from HEAD to the common commit between HEAD and origin/main, but not listing any commits from the tip of origin/main to that common commit.

filipe-cuim: Hmm, maybe I'm wrong here, and I may have misinterpreted this. Let me double check.

doc/dev/guidelines/practices.rst
simon-d-bmw: by the way, we could also mention how to run clang-tidy on all files:

```
run-clang-tidy-17 -p Directory_to_cleaned_compile_commands
```

.clang-tidy
simon-d-bmw, line 1: We need to exclude files under 3rdparty from clang-tidy checking.
Suggestion:
Add .clang-tidy in the 3rdparty that disables all checks:
```
Checks: '-*'
```

PR #252, opened 2025-10-27
Add guidelines about unit test and documentation

marcmo:
originates from CR #136631


doc/guidelines/unittests.rst
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Test if writing to a full MemoryQueue fails.
+     *
+     * The queue is filled completely with data. Afterwards, an attempt to write additional
+     * data is made. This should fail and the queue content should remain unchanged.
+     * Also, the write pointer should not have moved. The return value of the write operation
+     * should indicate the failure with the value MEMORY_QUEUE_FULL.
+     */
+    TEST_F(MemoryQueueTest, WriteToFullQueue)
```
christian-schilling, line 56: I'd suggest snake_case for test case names.

aschaal: I have intentionally used CamelCase, as this is used by Google in their examples, e.g. here: https://google.github.io/googletest/primer.html

doc/guidelines/module.rst
```diff
@@ -0,0 +1,142 @@
...
+``module.spec`` is a yaml-file which will be read by our build tools.
+
+.. code-block:: yaml
+
+    # VALUE + EXAMPLE           DESCRIPTION                       POSSIBLE VALUES                DEFAULT
+    # ---------------           -----------                       ---------------                -------
+
+    std_c: c                    # Minimum required c standard.    [c, c99]                       c
+
+    std_cxx: c++03              # Minimum required c++ standard.  [c++03, c++11, c++14, c++17]   c++14
```
rolandreichweinbmw: Shouldn't we limit to OpenBSW supported standards, namely C++14, 17, 20 and 23?

doc/guidelines/documentation/styleguide.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Whitespaces
+-----------
+
+Trailing whitespaces are not allowed.
+
+Column Width
+------------
+
+Limit the column width in all ``.rst`` files to **100 characters** whenever possible.
```
simon-d-bmw: I assume column width means line length. While I agree with limiting line length for _source code_, for markup languages[ semantic line breaks](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) are a much better option.
What do you think?

doc/guidelines/module.rst
aschaal: Yes, thank you, good catch. I have updated the commit.

doc/guidelines/diagrams/component.rst
```diff
@@ -0,0 +1,60 @@
...
+
+.. list-table::
+   :header-rows: 1
+   :width: 100%
+   :widths: 15 50 50
+
+   * - Name
+     - Graphical Representation
+     - Description
+   * - Module
```
simon-d-bmw: Why not "component"?

doc/guidelines/diagrams/deployment.rst
```diff
@@ -0,0 +1,52 @@
...
+       devices. Hierarchical devices can be modeled using composition or by defining an internal
+       structure. Examples are: µCs, µPs, cores, switches, crystals, etc.
+   * - Physical Component
+     - .. image:: _static/deployment/deployment-physical.png
+          :align: center
+          :scale: 100%
+     - A physical component represents physical parts, which are used by the device to
+       provide the processing capabilities, e.g. RAM, ROM, flash, cache, etc. Physical components do
+       not provide processing capabilities.
+   * - Module
```
simon-d-bmw: Component? Or Software Component maybe?

doc/guidelines/documentation/styleguide.rst
aschaal: Good idea. Both make sense.

The "hard" limit should still be 100 chars, as this makes using compare-tools way easier.
But breaking earlier based on semantics is a good thing. I also do it sometimes.

doc/guidelines/diagrams/component.rst
aschaal: We use module as synonym for component. But I see the point, this is not common.
I will change that to "Component / Module".

doc/guidelines/diagrams/deployment.rst
aschaal: Same as in "Component Diagram". Will improve this heading.

doc/guidelines/documentation/content.rst
```diff
@@ -0,0 +1,110 @@
...
+usually not important to the user, except for how often and in which context it has to be called.
+
+Source code comments are very welcome, but they are additional information and do not replace this
+documentation.
+
+Introduction
+------------
+
+This introduction is for every interested project member who wants to get in touch with the module.
+It should describe the purpose of the module and list the most important features.
```
simon-d-bmw: So the introduction describes only the most important features. Where shall the other features be described?

There could be a "features" chapter which could provide a short description of **each** of the module's features. Those feature titles could also be referred to in later chapters, and in the tests.

doc/guidelines/documentation/structure.rst
```diff
@@ -0,0 +1,80 @@
+Structure
+=========
+
+Files
+-----
+
+Documentation always starts in ``<module root>/doc/index.rst``.
```
simon-d-bmw, line 7: at the moment there is also a directory "doc/features". 
Looking at the content:
* doc/features/storage looks like it is the documentation of the storage module. However, I guess the feature _persistent storage_ needs more than just this module, i.e. a driver.
* Safety looks like it is an overarching documentation of all safety modules. 

The question is: Do we need such a feature level, next to the module documentation?

aschaal: This guideline is about module documentation only, meaning the documentation IN and FOR a SPECIFIC module.

The feature section in the OpenBSW documentation shall provide an overview of OpenBSW.
Usually a feature is covered by several modules. For storage this is the central storage module plus all available drivers.
Maybe some parts should be moved to the storage module documentation.
Additionally, on these feature pages, a demo integration in e.g. ReferenceApp can be mentioned.

As of today, this feature section is far away from complete! Let's discuss that separately how to proceed here.

doc/guidelines/documentation/content.rst
aschaal: Yes, this makes, especially this extra chapter. I will add it.
Also a note in the unit test guideline.

aschaal: Just pushed a new commit covering your comment.

doc/guidelines/documentation/structure.rst
simon-d-bmw: ok

PR #261, opened 2025-11-03
Switch to pre-built Docker Image

christophruethingbmw:


docker/README.md
```diff
@@ -1,5 +1,10 @@
 # Docker Information
 
+> [!WARNING]
+> The docker setup is currently in re-work. We are moving the image to a separate repository and
```
christian-schilling, line 4: While I agree that re-building the same image over and over should be avoided, I don't see how avoiding this requires moving the Dockerfile to a different repository. We can just as well keep the Dockerfile here and still provide pre-built images.

However, moving the Dockerfile means making changes to the docker and trying them out becomes much more difficult. Essentially discouraging contributions to it and changing the perception of the docker and by extension CI from a collaborative effort as part of OpenBSW to being an external artifact for consumption.

Not to mention what this means for trying to experiment with changes to docker/ci on branches, forks, etc...

And if we introduce more different purpose built Dockerfiles (like for different tools/toolchains) this will get even more complex without any upside.

If there are reasons for moving the Dockerfile to a different repo, please explain, but "to avoid re-building the image" is certainly not.

ghost: I agree with @christian-schilling, it is important to keep the Dockerfile in sync with the repository which uses it. Having it in the same repository allows for atomic changes to it and the code that it helps build.

christophruethingbmw: I see you point of having the `Dockerfile` inside of the repository can have advantages, however, I would like to add some more thoughts on why we propose it like here:

- In my opinion the docker image should not change too frequently, only in case we for example decide to add a new compiler, update an existing one, add specific tooling, ... That's not happening in daily work.
- I think it is important to have a 100% reproducible build environment which is defined in the repository. Therefore, we release fixed version of our docker image and reference this fixed version from our `docker-compose.yaml`. In case we check out a specific commit (even long in the past), the `docker-compose.yaml` references to the proper docker image that works with the (probably old) state of the repo.
- Keep in mind that even if the `Dockerfile` does not change there is no guarantee that the resulting docker image is the same. When installing distribution packages they might be updated, external files might have changed on servers out of our control, maybe build fails because files are no longer available, ...
- It is more convenient for a new developer to simply download a pre-built image instead of needing to build the image first. Especially since this might lead to an undefined docker image (see above).

Since we IMHO need these two requirements a) have a pre-built image and b) refer to a fixed version from the repo itself, the only (not indefinately complex) way I see is to go with fixed releases. In case we would push an image from a workflow run (which could be different for each workflow run, even if the `Dockerfile` stays the same), how would somebody know which image to use? We won't update the `docker-compose.yaml` on merge.

Of course we could keep the `Dockerfile` in this repo and push fixed releases/tags to this repo which then builds the docker image, but I think that would not help in any way for your points. It just makes things more complicated in my opinion.

alinemundt: I also agree with @christian-schilling.
I think we should try to keep the repository as self-contained and transparent as possible, without splitting off parts that are tightly coupled. Keeping the Dockerfile here makes it clear and easy to understand where and how the build environment is defined.

@christophruethingbmw I question the first requirement you mention. Do we really need a pre-built image?
The reason there doesn't seem very strong to me, the difference for a developer between downloading the image and building it themselves is marginal.

I also think this might not be a good solution in regards to long-term flexibility. While I like that we have currently have just a single development container, for easy setup for the developers, over time it might make sense to split this up, into separate containers for different toolchains or components (e.g., Rust, auxiliary tools, etc.).
Keeping everything in just one container can become unmaintainable.

I agree that reproducibility is important. But I think for that we should take different steps.
We should be more careful to pin versions, for example, instead of basing the docker on "ubuntu:22.04", we should base it on a concrete version like "jammy-20251001".

We could still consider keeping the pre-build images as an additional backup. But in my opinion our CI should not rely on those.

PR #263, opened 2025-11-04
Move DiagJobRoot reference from interface to implementation

christian-schilling:
This is not accessed through the base class anymore so it can be moved down.

Change-Id: Ibb9064d5dc61709e482421b595b2c2725386d097

### DIFF
--- a/libs/bsw/uds/include/uds/IDiagDispatcher.h
+++ b/libs/bsw/uds/include/uds/IDiagDispatcher.h
@@ -27,8 +27,8 @@ public:
      * Constructor
      * \param   sessionManager  IDiagSessionManager
      */
-    explicit IDiagDispatcher(IDiagSessionManager& sessionManager, DiagJobRoot& jobRoot)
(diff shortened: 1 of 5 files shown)


rolandreichweinbmw (approved):

libs/bsw/uds/include/uds/IDiagDispatcher.h
```diff
@@ -27,8 +27,8 @@ class IDiagDispatcher
      * Constructor
      * \param   sessionManager  IDiagSessionManager
      */
-    explicit IDiagDispatcher(IDiagSessionManager& sessionManager, DiagJobRoot& jobRoot)
-    : fDiagJobRoot(jobRoot), fSessionManager(sessionManager), fEnabled(true)
+    explicit IDiagDispatcher(IDiagSessionManager& sessionManager)
+    : fSessionManager(sessionManager), fEnabled(true)
```
matthiaskessler, line 31: That's actually a good start of refactoring the interface. This interface is not a real interface but an abstract base class, so it was misleading anyways.

PR #266, opened 2025-11-04
Update elf path in target toml files inside test/pyTest to match the latest build_dir path in build.py

marcmo:
originates from CR #136871

### DIFF
--- a/test/pyTest/conftest.py
+++ b/test/pyTest/conftest.py
@@ -38,7 +38,7 @@ class TargetSession:
         """
         self.capserial().clear()
         start_target_process(
-            self.target_name, True if TargetSession.counter == 1 else False
+            self.target_name, TargetSession.counter == 1
         )
 
     def stop(self):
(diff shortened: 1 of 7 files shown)


test/pyTest/conftest.py
```diff
@@ -28,7 +37,9 @@ def start(self):
         it is not intended to be called from a test.
         """
         self.capserial().clear()
-        start_target_process(self.target_name, True if TargetSession.counter == 1 else False)
+        start_target_process(
+            self.target_name, True if TargetSession.counter == 1 else False
```
christian-schilling: Why the if/else? Isn't `TargetSession.counter == 1` already a boolean?

test/pyTest/target_info.py
```diff
@@ -174,5 +176,5 @@ def load(targets, cmd_line_no_restart):
     TargetInfo.load(args.target, args.no_restart)
 
     for name in TargetInfo.by_name:
-        print("target \""+name+"\" read from target_"+name+".toml")
+        print('target "' + name + '" read from target_' + name + ".toml")
```
christian-schilling: Consider using an f-string here for better readability.

test/pyTest/conftest.py
nemadhu: @christian-schilling TargetSession.counter tracks the number of times TargetSession objects that has been instantiated. Pytest fixtures facilitates to create an object of TargetSession for each test.

test/pyTest/target_info.py
nemadhu: Yes, thanks for the review.

test/pyTest/conftest.py
ThoFrankACN: I think this is what @christian-schilling is suggesting:
```suggestion
            self.target_name, TargetSession.counter == 1
```

PR #268, opened 2025-11-06
Fix a bug in eepromDriver for posix

marcmo:
originates from CR #136656

### DIFF
--- a/platforms/posix/bsp/bspEepromDriver/include/eeprom/EepromDriver.h
+++ b/platforms/posix/bsp/bspEepromDriver/include/eeprom/EepromDriver.h
@@ -12,6 +12,14 @@ namespace eeprom
 class EepromDriver : public IEepromDriver
 {
 public:
+    EepromDriver();
+    ~EepromDriver();
+
+    EepromDriver(EepromDriver const&)            = delete;
+    EepromDriver& operator=(EepromDriver const&) = delete;
+    EepromDriver(EepromDriver&&)                 = delete;
+    EepromDriver& operator=(EepromDriver&&)      = delete;
+
     bsp::BspReturnCode init() override;
 
     bsp::BspReturnCode write(uint32_t address, uint8_t const* buffer, uint32_t length) override;
@@ -21,6 +29,6 @@ public:
 private:
     static constexpr size_t EEPROM_SIZE = 4096; // 4KB
     std::string const eepromFilePath    = EEPROM_FILEPATH;
-    int eepromFd{-1};
+    int eepromFd;
 };
 } // namespace eeprom
--- a/platforms/posix/bsp/bspEepromDriver/src/eeprom/EepromDriver.cpp
+++ b/platforms/posix/bsp/bspEepromDriver/src/eeprom/EepromDriver.cpp
@@ -13,45 +13,77 @@
 namespace eeprom
 {
 
-bsp::BspReturnCode EepromDriver::init()
+EepromDriver::EepromDriver() : eepromFd(-1)
 {
-    bool success = true;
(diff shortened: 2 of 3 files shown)


platforms/posix/bsp/bspEepromDriver/include/eeprom/EepromDriver.h
```diff
@@ -12,8 +12,18 @@ namespace eeprom
 class EepromDriver : public IEepromDriver
 {
 public:
+    EepromDriver() = default;
+    ~EepromDriver();
```
rolandreichweinbmw, line 16: virtual? Since we implement a virtual interface IEepromDriver?

platforms/posix/bsp/bspEepromDriver/src/eeprom/EepromDriver.cpp
```diff
@@ -90,4 +90,16 @@ EepromDriver::read(uint32_t const address, uint8_t* const buffer, uint32_t const
...
+{
+    if (-1 != eepromFd)
+    {
+        fsync(eepromFd);
+        close(eepromFd);
+        eepromFd = -1;
+    }
+}
+
+EepromDriver::~EepromDriver() { deinit(); }
```
rolandreichweinbmw: Isn't this a questionable asymmetry with virtual init() called directly from the StaticBsp while deinit() is called from this destructor?

amit-warbhe: I have addressed this,  please have a look.
Thank you.

amit-warbhe:
@christian-schilling I have modified the commit message, Please have a look.
Thank you.

christian-schilling (dismissed):

platforms/posix/bsp/bspEepromDriver/src/eeprom/EepromDriver.cpp
hbragge-acn: I guess another way to address this asymmetry would be to let the destructor do the cleanup instead of using a public deinit() method. That way IEepromDriver and StaticBsp can be left unchanged. The only "asymmetry" then would be that some IEepromDriver subclasses define a destructor and others don't, which makes sense because only one subclass has a need to do cleanup (for now at least). Opinions?

Edit: note that in case the file descriptor would be closed in the destructor, it might also make more sense to open it in the constructor instead of in the init() method

amit-warbhe: @rolandreichweinbmw what's your opinion on this?
I am okay with @hbragge-acn suggestion to proceed with.

rolandreichweinbmw: Fine for me!

platforms/posix/bsp/bspEepromDriver/src/eeprom/EepromDriver.cpp
```diff
@@ -13,45 +13,53 @@
...
+    if (ftruncate(eepromFd, EEPROM_SIZE) != 0)
     {
-        // File already exists: open it without O_EXCL
-        eepromFd = open(eepromFilePath.c_str(), O_RDWR);
-        success  = (-1 != eepromFd);
+        return ::bsp::BSP_ERROR;
     }
-    else
+
+    // Fill with 0xFF
```
hbragge-acn: This would fill the file with 0xFF on every init, meaning that any existing data in an already existing EEPROM file gets lost at startup. I think you should do this in the constructor instead, only when the file was newly created.

amit-warbhe: fixed, thank you

hbragge-acn (approved):
Looks good

PR #269, opened 2025-11-07
Pin specific ubuntu version in dockerfiles

marcmo:
originates from CR #137051

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:jammy-20251001
 
 RUN for i in 1 2 3; do \
         apt-get update && apt-get install -y \
--- a/docker/Dockerfile.docs
+++ b/docker/Dockerfile.docs
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:jammy-20251001
 
 RUN for i in 1 2 3; do \
         apt-get update && apt-get install -y python3.10 python3-pip openjdk-11-jdk graphviz \


christian-schilling (approved):

christian-schilling (approved):

docker/Dockerfile.dev
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:jammy-20251001
```
rolandreichweinbmw, line 1: Is there a precedent case where 22.04 is too ambiguous, to make sure this change actually improves the situation?

For first sight, ubuntu:22.04 is much easier to grasp, instead of everybody needing to know the ubuntu codenames and current versions by heart.

alinemundt: The two version specifiers are very different.

22.04 refers to the latest available image for that version, which gets updated periodically.
jammy-20251001 is a specific point-in-time snapshot of the image, and includes all updates and changes up to that point. This build is fixed and thus improves reproducibility, unlike 22.04.

I expect many developers are familiar with the names of the two latest ubuntu releases, but even if not I don't think it is too much to ask to just look it up, if someone is interested in the docker image.

christian-schilling: Readability does not really matter much here. Reproducibility does.
With targeting a branch it is very hard to tell in retrospect what versions were actually used to build a given image and by extension given firmware. Making obtaining an accurate SBOM next to impossible.
I'd even prefer to do `FROM ubuntu@sha256:09506232a8004baa32c47d68f1e5c307d648fdd59f5e7eaa42aaf87914100db3` (what jammy-20251001 resolves to) since it requires less trust:
With `22.04` we *know* the actual packages will change and we *don't know* what we are actually using.
With `jammy-20251001` we *trust* the maintainers of the image to keep tier promise to not change the tag. That includes trust in their ability to defend against malicious actors atempting a supply chain attack.
With `@sha` we know exactly what version we get and can even *verify* that we got what we expected.

What we really need is to document somewhere that absolutely nothing is allowed to be downloaded from the internet by CI without a checksum verification.
Or even better have one dedicated step at the beginning (could even be git lfs) to download all required files and then do `docker build --no-network` (same for all following steps of the build: no network access)

rolandreichweinbmw: I agree that checksumming is a good idea for this purpose.

To prevent this from being rather theoretical, we would need to follow this approach consistently. Especially, as long as we download other parts of the toolchain (LLVM and ARM GCC) from the internet, the issue still persists.

Is it reasonable at this point to enforce this for all downloads, e.g. in the Dockerfile?

christian-schilling: You are absolutely right, this needs to be followed consistently. And yes for other parts of the toolchain as well.
A first step would be to just check the checksums for all downloads. And yes, can be done in the Dockerfile, even though I think even better would be to download & check before even starting to build.
This is still only halfway there since it does not cover the risk that at some point in the future (we do intend this to be a long lived project after all...) like a decade from now some URL are simple not available or moved. So we also need to store all files in a location we control.

In any case as long as we are not all the way there yet we should take every opportunity to get a bit closer and not just resign with an "we can't do it for everything so we just don't do it for anything" approach.

PR #270, opened 2025-11-10
Add commit message guideline

marcmo:
originates from CR #136856

### DIFF
--- a/doc/guidelines/commit_message.rst
+++ b/doc/guidelines/commit_message.rst
@@ -1,0 +1,262 @@
+.. _commit_message:
+
+Commit Message
+==============
+
+Our commit messages are read by many people, e.g.:
+
+- experienced developers with deep knowledge about basis software
+- inexperienced developers or developers from other domains
+- project managers, problem managers, quality managers, etc.
+- you
+
+A consistent style makes it easier for all of them to read and understand the commits.
+
+Rules
+-----
+
+.. contents::
+    :local:
+    :depth: 1
+
+Use a Subject Line and a Body
++++++++++++++++++++++++++++++
+
+A well-formed commit message consists of
+
+- a brief summary line,
+- followed by a blank line,
+- and an optional detailed description.
+
+Git treats everything before the first blank line as the title,
+which appears in logs, history views, and other tools.
+
+The format is as follows:
+
+.. code-block::
+
+   <single subject line>
+
+   <optional body>
+   <which may span>
+   <multiple lines>
+
+
+For straightforward changes, the body may be omitted entirely.
+
+Example:
+
+.. code-block:: shell
+
+   $ git commit -m "Fix typos in ethernet driver documentation"
+
+Stick to Plain Text in the Subject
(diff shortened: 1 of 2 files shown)


christian-schilling (approved):

doc/guidelines/commit_message.rst
```diff
@@ -0,0 +1,259 @@
...
+
+.. code-block:: shell
+
+   $ git commit -m "Fix typos in ethernet driver documentation"
+
+Stick to Plain Text in the Subject
+++++++++++++++++++++++++++++++++++
+
+Subject lines should be as consistent as possible, giving only the utmost important information
+to the reader. To help keep consistency in titles, stick to plain text only - no markup syntax
```
rolandreichweinbmw, line 57: While we are at it, maybe explicitly mention UTF-8 or 7-bit ASCII? (Depending on what we agree on.)

doc/guidelines/commit_message.rst
```diff
@@ -0,0 +1,259 @@
...
+A project based on OpenBSW can grep for these commits and evaluate,
+if they need to synchronize the corresponding modules.
+
+Beside for bugs there are no other keywords specified. The audiences and use cases differ way too
+far to decide which keywords should be highlighted.
+
+Use Natural Language
+++++++++++++++++++++
+
+To make it easy to read the commit messages use simple natural words instead of inventing a funny
```
rolandreichweinbmw, line 192: English?

doc/guidelines/commit_message.rst
```diff
@@ -0,0 +1,259 @@
+.. _commit_message:
+
+Commit Message
```
rolandreichweinbmw, line 3: Also: PR Message Guideline? (Separate Commit or PR?)

doc/guidelines/commit_message.rst
aschaal: As discussed today in the Eclipse PR meeting, we don't specify that here. Contributors simply have to stick to the Git(Hub) standard.

If a committer does not agree to a commit messages, it can be rejected.

However, I will add to "Use Natural Language" that English has to be used. This excludes "uncommon" characters from other languages.

doc/guidelines/commit_message.rst
aschaal: Yes, as discussed in the Eclipse PR meeting, I will prepare that, but in a separate PR after this has been merged.

doc/guidelines/commit_message.rst
rolandreichweinbmw: Agree.

Looks like git makes sure that UTF-8 is the default encoding with deviations being recorded explicitly with commits: https://git-scm.com/docs/git-commit (`i18n.commitEncoding` in the `encoding` header)

doc/guidelines/commit_message.rst
aschaal: Yes, I have added a sentence, see some lines below.

PR #275, opened 2025-11-12
Use ETL_OR_STD for swap()

rolandreichweinbmw:
As documented at https://www.etlcpp.com/no_stl.html, the intended pattern for some ETL interfaces is to use ETL_OR_STD instead of plain etl namespace because some interfaces can't be implemented in namespace etl due to conflicts with std when in ETL_USING_STL mode (e.g. due to ADL).

This change enables OpenBSW to be compiled both with and without ETL_NO_STL in etl_profile.h


rolandreichweinbmw:
I built and tested both versions, for comparison:

ETL_NOT_USING_STL (via ETL_NO_STL as currently in etl_profile.h):
```
  4 .text         000299d8  00000600  00000600  00010600  2**5
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
```

ETL_USING_STL (ETL default):
```
  4 .text         000299b8  00000600  00000600  00010600  2**5
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
```

I.e. difference in .text section is minimal

PR #279, opened 2025-11-12
Update minimum cmake version in docs

marcmo:
originates from CR #135832


doc/learning/setup/setup_posix_ubuntu_build.rst
```diff
@@ -17,11 +17,18 @@ For Ubuntu :prop:`tool:ubuntu_version`, the ``apt`` package ``build-essential``
 
 You will also need ``cmake`` version >= :prop:`tool:cmake_version`. To install it, simply run:
 
+.. code-block:: bash
+
+    sudo snap install cmake --classic
+
+which (at time of writing) installs ``cmake`` version 4.1.1 on Ubuntu :prop:`tool:ubuntu_version`.
```
rolandreichweinbmw: Commit message suggests 3.28?

christian-schilling: Who cares what version this installs at the time of writing?
Does mentioning it mean this version is supported?
If yes do we test supporting it?
If no why mention this version specifically?

doc/dev/learning/setup/setup_posix_build.rst
```diff
@@ -17,14 +17,21 @@ For Ubuntu :prop:`tool:ubuntu_version`, the ``apt`` package ``build-essential``
 
     sudo apt install build-essential ninja-build
 
-You will also need ``cmake`` version >= :prop:`tool:cmake_version` . To install it, simply run:
+You will also need ``cmake`` version >= :prop:`tool:cmake_version`. To install it, simply run:
```
christian-schilling, line 20: Could we rearrange this to avoid the duplicated instuctions?

doc/dev/properties.yaml
```diff
@@ -1,6 +1,6 @@
 tool:
   gcc_version: "11.x"
-  cmake_version: "3.22"
+  cmake_version: "3.28"
```
christian-schilling, line 3: Can we check in CI that the version(s) mention here actually work(s)?

doc/learning/setup/setup_posix_ubuntu_build.rst
jan-bruckner-esr: I certainly don't care about the specific version at time of writing. But there was a similar sentence below, so I just adapted it. Perfectly fine for me to remove it.

I don't think there is any testing done regarding the cmake version, since this has been incorrect for some time now and no one seems to have noticed.

doc/dev/properties.yaml
jan-bruckner-esr: I would like that also. But that is not the goal of this PR and I think we should first focus on correcting the documentation.

doc/dev/learning/setup/setup_posix_build.rst
jan-bruckner-esr: I merged the posix instructions for Ubuntu and Windows into a single file.

doc/learning/setup/setup_posix_ubuntu_build.rst
jan-bruckner-esr: Removed references to the specific version getting installed.

PR #280, opened 2025-11-12
Cleanup std usage to use ETL consistently

rolandreichweinbmw:
Some typo fixed in a separate commit.


christian-schilling (changes requested):
The separate type fix commit needs to be it's own PR.

rolandreichweinbmw:
> The separate type fix commit needs to be it's own PR.

Ok, removed this commmit from the PR.

PR #285, opened 2025-11-17
Remove AbstractDiagnosisConfiguration

christian-schilling:
By moving the templated container types out and using the ETL provided interface types the need for an abstract class was removed.

### DIFF
--- a/executables/referenceApp/application/src/systems/UdsSystem.cpp
+++ b/executables/referenceApp/application/src/systems/UdsSystem.cpp
@@ -30,15 +30,17 @@ UdsSystem::UdsSystem(
 , _jobRoot()
 , _diagnosticSessionControl(_udsLifecycleConnector, context, _dummySessionPersistence)
 , _communicationControl()
-, _udsConfiguration(
+, _udsConfiguration{
       udsAddress,
       transport::TransportConfiguration::FUNCTIONAL_ALL_ISO14229,
-      ::busid::SELFDIAG,
       transport::TransportConfiguration::DIAG_PAYLOAD_SIZE,
+      ::busid::SELFDIAG,
+      true,  /* activate outgoing pending */
       false, /* accept all requests */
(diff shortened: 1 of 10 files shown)


rolandreichweinbmw (dismissed):
LGTM!

libs/bsw/uds/src/uds/DiagDispatcher.cpp
```diff
@@ -323,7 +328,7 @@ IncomingDiagConnection* DiagDispatcher::requestIncomingConnection(TransportMessa
     {
         ::async::LockType const lock;
         pConnection = acquireIncomingDiagConnection(
-            fConfiguration.IncomingDiagConnectionPool, ::etl::move(fConfiguration.Context));
+            incomingDiagConnectionPool, ::etl::move(fConfiguration.Context));
```
matthiaskessler: I think `etl::move` is most likely wrong here. I know it wasn't part if this change but I just noticed it.

executables/referenceApp/application/src/systems/UdsSystem.cpp
```diff
@@ -30,15 +30,17 @@ UdsSystem::UdsSystem(
 , _jobRoot()
 , _diagnosticSessionControl(_udsLifecycleConnector, context, _dummySessionPersistence)
 , _communicationControl()
-, _udsConfiguration(
+, _udsConfiguration{
```
matthiaskessler, line 33: Using `{` here instead of `(` makes the style of this class a but non-uniform.

christian-schilling: I think we are going to see more {} in the future 🔮

libs/bsw/uds/src/uds/DiagDispatcher.cpp
christian-schilling: True, fixed.

Issue #284, opened 2025-11-14
Deterministic Docker Image Build


christophruethingbmw:
In the context of #261 we started a discussion on how to deal with our docker images in the future. This issue is to preserve the results of this discussion and document the basic ideas and principals we agreed on.

- We want to keep the `Dockerfile` inside of our source code repository and closely related to the code. This simplifies changes compared to a release-based docker image maintained in another repository and provided as a pre-built image. Changes to the `Dockerfile` can be made with potential changes in code and are tested by the same CI pipeline as any other change.
- We do not provide a pre-built image right now, both developers and CI will build the image whenever needed. To increase stabililty and performance in the CI, we use the GitHub actions cache to cache the image build itself and avoid a full rebuild for each workflow. For local developers image builds are typically cached anyways.
- Instead of mitigating the issue of non-reproducibility using pre-built images, we try to work in the future on a fully reproducible docker build which ensures that an image can be re-build at and time. This include pinning the base image, ensuring that all external artifacts are pinned (by sha) to a very specific version, that all external artifacts are stored in local mirrors to be independent of external source, ... This approach allows to still adapt even very old image in case necessary, know the exact contents of the image (e.g. in case of security vulnerabilites), ... The exact technical solutions for this are still to clarify.
- To avoid that the image get's outdated quickly and to follup on the latest development, we also need to think on how to regularly update the image content. This requires human interaction but should be automated as much as possible , e.g. by scripting the update and only reviewing the new state (similar to a pip lock file). Then we can easily update but still ensure that the resulting state is fully reproducible.

PR #286, opened 2025-11-17
Refactor Docker Image with top-level docker-compose File

christophruethingbmw:
Based on the general discussion in #284, this PR simplifies usage of the development docker image and introduces caching of the image in the CI using the Github Actions Cache. 

- A developer can run the development docker container using the provided `development` service in the top-leve `docker-compose.yaml` by just executing `docker compose run --build development`. Inside the container, it is possible to run 1:1 the same commands as in the workflow files of the CI, making it easy to reproduce the behavior.
- The CI now utilizes caching for building the docker image. This reduces the access of external resources and avoids running into rate limits while downloading. Since we use the plain runner images provided by Github we still need to import the image into the runner (taking around 2min), but that's unavoidable.

### DIFF
--- a/.ci/buildoperations.py
+++ b/.ci/buildoperations.py
@@ -31,6 +31,12 @@ def get_environment(cxxid: str, platform: str):
     env["CTEST_PARALLEL_LEVEL"] = str(threads)
     env["CMAKE_BUILD_PARALLEL_LEVEL"] = str(threads)
 
+    # FIXME: We should not hard-code the paths/versions of the compiler in this
+    # python script. The build should also work with cmake only without using
+    # this script and we even have path's as well in the toolchain files. In the
+    # future we might use more tailored docker images which only contain a
+    # single compiler, so there weill be no longer a question of choice here.
+
     if platform == "arm":
         if cxxid == "gcc":
             env["CC"] = get_full_path("/opt/arm-gnu-toolchain/bin/arm-none-eabi-gcc")
--- a/.ci/code_coverage.py
+++ b/.ci/code_coverage.py
@@ -5,6 +5,8 @@ import shutil
 import subprocess
 import sys
 
+# FIXME: Hard-coding the compiler version in this python script feels wrong. It
+# should be changed in the future to provide the compiler from the outside.
 GCC_VERSION = 11
 
 def get_full_path(command):
@@ -111,6 +113,11 @@ def generate_coverage():
 
 
 def generate_badges():
+    # FIXME: It's questionable whether we want to have a dependency to an
+    # external service for generating these badges. This introduces a possible
+    # cause of instabilities in case the external service becomes unavailable,
+    # as already happened in the CI.
+
     result = subprocess.run(
         [
             "lcov",


README.md
```diff
@@ -39,6 +39,33 @@ allowing developers to get up and running quickly.
...
+> In case you want to use a custom history file for the commands you run in the container you can
+> also set the `DOCKER_HISTORY` variable, which defaults to the `~/.docker_history` file.
+>
+> Note, that we bind mount your current working directory into the container and use it as working
+> directory. This makes sure, that you will have the same paths inside and outside of the container
+> when for example loading a generated elf into a debugger or following symlinks created within the
+> container.
+
+```
+host> DOCKER_UID=$(id -u) DOCKER_GID=$(id -g) docker compose run --build development
```
christian-schilling, line 64: Using a docker container as a persistent, interactive shell for developement is not a good idea and should not the the documented or encouraged as a way of working:
- It assumes that all build steps happen in the same contaier, which might be true at the moment, but is most certainly not sustainable as soon as more tools are added to the build process.
- It interacts badly with version control. With every change of the worktree due to "git pull" or swtiching branches for example the image needs to be potentially rebuilt and the container restarted. Keeing the container running can lead to both false positives and false negatives regarding build and test status.

.github/workflows/format.yml
```diff
@@ -6,13 +6,21 @@ jobs:
...
 
-     - name: Run the format.py inside the docker container
-       run: |
-          docker run --rm -v "$PWD:/home/jenkins" -w /home/jenkins --user $(id -u):4996 dev python3 .ci/format.py
+      - name: Build development Image
+        uses: docker/build-push-action@v6
+        with:
+          context: "{{defaultContext}}:docker/development"
+          push: false
+          tags: openbsw-development:latest
```
christian-schilling: We never ever want to refer to anything by "latest".
The image needs to be tagged with a suitable hash of the inputs used to build it.

.github/workflows/puncover_tool.yml
```diff
@@ -13,56 +13,41 @@ jobs:
...
-      - name: Set up Python
-        uses: actions/setup-python@v5
+      - name: Build development Image
+        id: build-docker-development
+        uses: docker/build-push-action@v6
         with:
-          python-version: '3.12'
+          context: "{{defaultContext}}:docker/development"
+          push: false
+          tags: openbsw-development:latest
```
christian-schilling: Once again: "latest" is the same as "random" when it comes to ci. Never ever use it!
This needs to be a deterministic key (hash).

.github/workflows/rim.yml
```diff
@@ -6,13 +6,21 @@ jobs:
...
 
-     - name: Run the third party libraries check via RIM inside the docker container
-       run: |
-          docker run --rm -v "$PWD:/home/jenkins" -w /home/jenkins --user $(id -u):4996 dev rim status -w -d --verify-clean
+      - name: Build development Image
+        uses: docker/build-push-action@v6
+        with:
+          context: "{{defaultContext}}:docker/development"
+          push: false
+          tags: openbsw-development:latest
```
christian-schilling: Once more: Don't use "latest". Ever.

.github/workflows/rim.yml
```diff
@@ -6,13 +6,21 @@ jobs:
   rim:
     runs-on: ubuntu-latest
```
christian-schilling, line 7: Same here: No "ubuntu-latest". This also needs to pinned to an exact version.

.github/workflows/run-pytest-on-posix.yml
```diff
@@ -3,48 +3,54 @@ name: Run pytest for posix platform
...
-            docker run --rm --network=host \
-              --cap-add=NET_ADMIN --cap-add=SYS_MODULE \
-              --device /dev/net/tun \
-              --volume /lib/modules:/lib/modules \
-              -v "$PWD:/home/jenkins" \
-              -w /home/jenkins \
-              --user "$(id -u):4996" \
-              dev python3 .ci/pytest.py
+  run-pytest-on-posix:
+    runs-on: ubuntu-latest
```
christian-schilling, line 7: Again...

.github/workflows/run-pytest-on-posix.yml
```diff
@@ -3,48 +3,54 @@ name: Run pytest for posix platform
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build development Image
+        uses: docker/build-push-action@v6
+        with:
+          context: "{{defaultContext}}:docker/development"
+          push: false
+          tags: openbsw-development:latest
```
christian-schilling: Again...

.github/workflows/run-pytest-on-posix.yml
```diff
@@ -3,48 +3,54 @@ name: Run pytest for posix platform
...
+
+      - name: Set up virtual ethernet interface
+        run: |
+          sudo tools/enet/bring-up-ethernet.sh
+
+      - name: Run the pytest inside the docker container
+        run: >-
+          DOCKER_UID=$(id -u) DOCKER_GID=$(id -g) DOCKER_HISTORY=/dev/null docker compose run --rm development-sil
+          bash -c "
+          pip3 install -r test/pyTest/requirements.txt &&
```
christian-schilling: Requirements should be baked into a (cached) docker image, not installed every time.

docker/development/files/.bash_profile
```diff
@@ -0,0 +1 @@
+[[ -f ~/.bashrc ]] && . ~/.bashrc
```
christian-schilling, line 1: I don't see that this file is for. Using docker interactively is bad practice as outlined in a previous comment.

docker/development/files/.bashrc
```diff
@@ -0,0 +1,2 @@
+[[ $- != *i* ]] && return
+export PS1="openbsw> $PS1"
```
christian-schilling, line 2: same

docker/development/Dockerfile
```diff
@@ -11,8 +11,9 @@ RUN for i in 1 2 3; do \
             iputils-ping \
             lcov \
             ninja-build \
-            python3-pip \
             python3.10 \
+            python3-pip \
+            python3.10-venv \
```
christian-schilling: Setting up venv inside docker seems really pointless to me: Docker *is* the tool manage diffent enviorentments per tool why stack another tool that serves essentially the same purpose but is less capable on top of it?

docker/development/Dockerfile
```diff
@@ -61,31 +61,11 @@ RUN wget https://github.com/mozilla/sccache/releases/download/v0.10.0/sccache-v0
...
-
-RUN groupadd -g 4996 docker
-RUN useradd --shell /bin/bash --uid ${user_id} -g docker -m jenkins
+RUN pip3 install cmakelang pytest rich pyelftools
 
-USER jenkins
+RUN useradd build
+RUN install -d /home/build --mode 0777 --owner build --group build
+COPY --chown=build:build --chmod=0666 files/.bashrc /home/build/.bashrc
+COPY --chown=build:build --chmod=0666 files/.bash_profile /home/build/.bash_profile
```
christian-schilling, line 52: Once again: Don't encourage interactive use of docker containers!

docker/docker-compose.yaml
```diff
@@ -1,19 +1,4 @@
...
-    working_dir: /home/jenkins
-    volumes:
-      - ..:/home/jenkins
-      - ../.docker_history:/root/.bash_history
-    command: /bin/bash
-    user: "${UID}:4996"
-    tty: true
-
   docs:
     build:
```
christian-schilling, line 3: I thought the idea was to have one top level docker-compose.yaml. This looks like a second one.

tools/puncover_tool/pyproject.toml
```diff
@@ -6,7 +6,7 @@ authors = [
     {name = "suhashini.naik",email = "suhashini.naik@accenture.com"}
 ]
 readme = "README.md"
-requires-python = ">=3.12"
+requires-python = ">=3.10"
```
christian-schilling, line 9: Don't use ">=" for versions, always pin to exact versions.

docker-compose.yaml
```diff
@@ -0,0 +1,101 @@
+services:
+  development:
```
christian-schilling, line 2: I really doubt that docker compose is a useful tool for our purpose at all. But that might be something to clarify.

docker-compose.yaml
```diff
@@ -0,0 +1,101 @@
...
+    image: openbsw-development:latest
```
christian-schilling: One last time: No random versions. Please! 🙏

tools/puncover_tool/pyproject.toml
christophruethingbmw: I did not introduce this and it's not in the scope of this PR to change/fix this. I just adapted the requirement to allow the puncover_tool to run with the currently existing docker image (which only comes with python 3.10).

docker/docker-compose.yaml
christophruethingbmw: I just focused as a first step on the `development` image, this file is still there for the `docs` image which I did not touch at all.

.github/workflows/rim.yml
christophruethingbmw: This is not related to the changes in this PR. It is the ubuntu image running the github runner. The images are anyway out of our control right now, otherwise we would need to provide our own images and I am not sure whether that's possible right now (not sure how billing works for github right now).

Also, I do not think this has a big impact, since we anyway run the build in the docker container which gives us the encapsulated and reproducible behavior.

christophruethingbmw:
> I get really triggered whenever see the word "latest" in a build/ci context.

I already thought that you were misleaded by this term, I agree that (to avoid heart attacks) we can rename it to `local`.

Regarding venv: I think we can install the pip dependencies (from the different requirement files we have right now) into the container. To make this more reproducible we could also check-in a compiled lock file (using pip-compile) and use this for installing it into the container. Or we just do it like we did until now of directly installing the `requirements.txt` from the source tree.

Regarding interactive: I think that's something we better discuss directly.

Regarding the link you shared: I am not sure about all details at the first glance, but do I see it correctly that they have a separate ci tool which runs "jobs" and does "things"? From what I see they have a similar structure for the `Dockerfile`s like we have started here, one folder for an image with a `Dockerfile` inside, we just have (as discussed) at the moment the single development docker image and might split it in the future in multiple smaller ones. From what I see, the `Dockerfile`s in the link are also not fully reproducible, they also use `ubuntu:25.10` base image, install pacakges via `apt`, ... and when looking into it at the first glance it runs quite some scripts which makes it harder to follow what's exactly done.

CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 cmake_minimum_required(VERSION 3.28 FATAL_ERROR)
 
+# We do not support modules at the moment
+set(CMAKE_CXX_SCAN_FOR_MODULES FALSE)
```
rolandreichweinbmw, line 4: Great! This fixes clang builds with C++20+

.ci/code_coverage.py
```diff
@@ -5,6 +5,7 @@
 import subprocess
 import sys
 
+GCC_VERSION = 11
```
christian-schilling, line 10: While this is slightly better than before, even better would be to pass this in from somewhere outside.
Can be done later, FIXME comment maybe?

.ci/buildoperations.py
```diff
@@ -33,22 +33,22 @@ def get_environment(cxxid: str, platform: str):
 
     if platform == "arm":
         if cxxid == "gcc":
-            env["CC"] = get_full_path("arm-none-eabi-gcc")
-            env["CXX"] = get_full_path("arm-none-eabi-g++")
+            env["CC"] = get_full_path("/opt/arm-gnu-toolchain/bin/arm-none-eabi-gcc")
```
christian-schilling, line 42: This I don't quite understand:
Shouldn't the build work without this script as well? Is this not purely a cmake thing?

.ci/buildoperations.py
```diff
@@ -33,22 +33,22 @@ def get_environment(cxxid: str, platform: str):
...
-            env["CXX"] = get_full_path("/usr/bin/llvm-arm/bin/clang++")
+            env["CC"] = get_full_path("/opt/llvm-et-arm/bin/clang")
+            env["CXX"] = get_full_path("/opt/llvm-et-arm/bin/clang++")
             return env
 
     if platform == "linux":
         if cxxid == "gcc":
-            env["CC"] = get_full_path("gcc")
-            env["CXX"] = get_full_path("g++")
+            env["CC"] = get_full_path("gcc-11")
```
christian-schilling, line 52: Hardcoding the compiler versions here in a python script does not seem right to me.
It mixes the "what" and "how" of the build.
All of this path related code here is just working around the issues created by having a single image for all toolchains. It would completely go away once we have focused docker images.
Also something to fix later...But can we at least use a constant with a FIXME at the top of the file to not forget about it?

.ci/code_coverage.py
```diff
@@ -133,7 +139,7 @@ def generate_badges():
                 "wget",
                 f"https://img.shields.io/badge/coverage-{line_value}%25-brightgreen.svg",
```
christian-schilling, line 147: Also something to fix later in a separate change:
Last week checks were blocked because shields.io was down. Calling out to an external service just to substitute a few values in an SVG file seems just crazy to me.
But on top of that: Why does the pre-merge check deal with this badge at all?
And on top of that: Do we really need coverage in the pre-merge check? (maybe..maybe not)


`#  FIXME` ?

.ci/code_coverage.py
christophruethingbmw: I agree, this hard-coding (also as noted in the later comments) in the python scripts looks strange to me. Anyhow, I am not sure whether we want to support these script as they are right now in the future, at least it's questionable whether they are really needed. I will add a todo comment here.

.ci/buildoperations.py
christophruethingbmw: Yeah, I agree, I would prefer to have cmake presets for this, I don't know why it was implemented in this way but I did not want to refactor this as well in this PR. The entire `buildoperations.py` script is something we can thing about for the future.

.ci/buildoperations.py
christophruethingbmw: I agree, it can potentially go away, as you say, I think that's something we can discuss on for future change. Once we have more tailured docker images, you choose the compiler/version by selecting the docker image and then there is anyway only one choice in the image.

.ci/code_coverage.py
christophruethingbmw: I fully agree, I also wonder why we need an external service to generate an SVG, that's something we could also do locally in case we need it. I think calculating coverage in pre-merge check can be useful to later be able to fail the build in case of reduced / too low coverage. But I doubt we need any badges in the pre-merge check.

Issue #192, opened 2025-09-20
EepromDriver on s32k1xx: Executable code on stack


rolandreichweinbmw:
In `platforms/s32k1xx/bsp/bspEepromDriver/src/eeprom/EepromDriver.cpp`, the function `launchCommandFromRAM()` is copied to RAM (to be executed in a context without flash access). More precisely, 256 bytes, starting at the symbol `launchCommandFromRAM` are copied to an array on the stack, and executed.

This is highly questionable.

Wouldn't it be more appropriate to create a (small) text section in SRAM for that via the linker script, instead of the above solution?

BTW: The issue surfaced because it is the root cause of https://github.com/eclipse-openbsw/openbsw/issues/188 : The new ETL is using compiler builtins for memcpy etc. However, in this scenario, GCC optimizes `launchCommandFromRAM` away silently (most probably since no user code is referring to it anymore?), leaving us with undefined behaviour by executing it. Restoring the old ETL implementation is only a workaround (although non-critical since the compiler optimizes the rest of the code to basically the same algorithms).

I started with explicitly making sure that `launchCommandFromRAM()` is still linked, reachable and executed. But this wouldn't make the current code better.

PR #289, opened 2025-11-17
Fix launch command from RAM

rolandreichweinbmw:
Executable code on the stack needs to be abandoned and is now solved via a separate section in the linker script.

This fixes https://github.com/eclipse-openbsw/openbsw/issues/192

### DIFF
--- a/executables/referenceApp/platforms/s32k148evb/main/linkerscript/application.dld
+++ b/executables/referenceApp/platforms/s32k148evb/main/linkerscript/application.dld
@@ -26,6 +26,7 @@ MEMORY
   /* SRAM_L */
   Data (RW) : ORIGIN = SRAM_L_START, LENGTH = SRAM_L_END + 1 - SRAM_L_START
   /* SRAM_U */
+  ApplicationRAM (RX) : ORIGIN = SRAM_U_START, LENGTH = 0x100
   NoInit (RW) : ORIGIN = SRAM_U_END + 1 - 0x400, LENGTH = 0x400
 }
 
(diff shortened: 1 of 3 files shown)


platforms/s32k1xx/bsp/bspEepromDriver/src/eeprom/EepromDriver.cpp
```diff
@@ -112,23 +116,7 @@ bsp::BspReturnCode EepromDriver::launchCommand()
         FTFC->FCCOB[i + 3U] = _cmd[i];
     }
 
-    /* The maximum possible size of the launchCommandFromRAM function is 0x88 bytes,
-     * which was checked with different compiler optimization options.
-     * Use slightly bigger RAM buffer to store the function */
-    constexpr size_t FUNCTION_BUF_SIZE = 0x100U;
-    __attribute__((aligned(32))) ::etl::array<uint8_t, FUNCTION_BUF_SIZE> copyOfLaunchFunction;
-
-    /* Some EEPROM commands (ex. Program partition) must not be launched from flash memory,
```
hbragge-acn, line 121: Thanks for the fix, looks good and seems to work. Regarding this comment however, maybe it should be kept (except the thumb state part)? Eventually someone will stumble into this and wonder what's it about.

rolandreichweinbmw: Thanks for the hint! Will re-add the comment to the declaration of the function with the section attribute.

executables/referenceApp/platforms/s32k148evb/main/linkerscript/application.dld
```diff
@@ -227,7 +228,21 @@ SECTIONS
...
+    KEEP(*(.launchFromRAM))
+    KEEP(*(.launchFromRAM*))
+    . = ALIGN(32);
+    __LAUNCH_FROM_RAM_END = .;
+  }
+  > ApplicationRAM = 0xffffffff
+
   ___data_size = _edata - _sdata;
 
   __DATA_ROM_START = __DATA_ROM;
```
manzoorAJ, line 247: One observation: in the computation of `__USED_FLASH_END` linker symbol does not include the `.launchFromRAM` section, even though this section is stored in flash. As a result, the flash utilization and CRC calculations that rely on this symbol become inaccurate.

rolandreichweinbmw: Good catch! Adjusting accordingly.

PR #297, opened 2025-11-19
Remove empty PACK_STRUCT_STRUCT definition

AGO-Development:
Empty definition was overriding LWIP's automatic compiler detection, causing struct packing issues and crashes.

### DIFF
--- a/platforms/posix/lwipSysArch/include/arch/cc.h
+++ b/platforms/posix/lwipSysArch/include/arch/cc.h
@@ -14,8 +14,6 @@ typedef unsigned char bool_t;
 
 #define LWIP_CHKSUM_ALGORITHM 3
 
-#define PACK_STRUCT_STRUCT
-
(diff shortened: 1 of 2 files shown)


christian-schilling:
Is there a way to reproduce/test the issues you observed? Can those be included as some form of test?

AGO-Development:
> Is there a way to reproduce/test the issues you observed? Can those be included as some form of test?

We faced with this issue in internal project. I can try to reproduce it on openbsw reference application, but root cause is clear. Since with empty redefinition of PACK_STRUCT_STRUCT you get struct ip_hdr alligned by 4 bytes and when iphdr = (struct ip_hdr *)p->payload is called in ip4.c it can be the case when payload is aligned by 2 bytes (as in our internal project) and it leads to crash, because from C standard perspective, it is undefined bahavior case.
@christian-schilling do you have the reason why the PACK_STRUCT_STRUCT is defined empty instead of definition done in arch.h by LwIP?

PR #312, opened 2025-11-25
Fix launchCommandFromRAM

ghost:
It was being inlined by the compiler, defeating the purpose of running from RAM and causing hard faults

Change-Id: Idb3283e11b9164fef0cffa59478c77c387ac06da


rolandreichweinbmw:
Yes, a description in the commit message how to reproduce would be best. Especially when we want to regression test later.

I also remember that in the past I needed a "second startup" but the root cause is not clear to me yet.

lmgivelaccenture:
In our case, I think the fact that it was triggering a hard fault in that particular scenario was pure chance. It just happened that those particular conditions meant that we reliably tried to read from code flash at just the right time when the eeprom configuration command was running. Why on second startup this wasn't an issue is hard to guess, and probably has to do with the timing of the relevant operations.

rolandreichweinbmw (approved):

PR #313, opened 2025-11-25
Skip stats related test for posix-threadx as stats commands are not currently supported

marcmo:
originates from CR #137381

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,7 +82,8 @@ jobs:
         id: build-docker-development
         uses: docker/build-push-action@v6
         with:
-          context: "{{defaultContext}}:docker/development"
+          context: .
+          file: docker/development/Dockerfile
           push: false
           tags: openbsw-development:local
           outputs: type=docker
--- a/.ci/pytest.py
+++ b/.ci/pytest.py
@@ -19,9 +19,8 @@ def build_reference_app():
 
 def run_pytest():
     os.chdir("./test/pyTest")
-    subprocess.run(["pytest", "--target=posix-freertos"], check=True)
-    subprocess.run(["pytest", "--target=posix-threadx"], check=True)
-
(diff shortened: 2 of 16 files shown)


.github/workflows/build.yml
```diff
@@ -82,7 +82,8 @@ jobs:
         id: build-docker-development
         uses: docker/build-push-action@v6
         with:
-          context: "{{defaultContext}}:docker/development"
+          context: .
```
christophruethingbmw, line 85: @johannes-esr Actually we were really happy to not need the entire repository as the build context. But of course, in case we want to use something from a top level directory during image build, there is no other option than using root of the repo as build context. I could imagine different possibilites here:

- Make the uds tool a standalone tool maintained separately, release it as a separate pypi package and just install it as all the other pypi packages we have.
=> Most likely not what we want right now,  since the tool is just very tailored for us and not ment as a generic tool/library.
- Just allow running the tool directly by e.g. `python udsTool.py` such that we do not need to install anything in the docker at all. Only the dependencies can be installed via the existing `requirements.txt` (lock) file we have.
=> This would also solve the issue that by doing another pip install now we potentially screw up lor locked `requirements.txt` used in the docker.
- Add a `.dockerignore` to the root to limit the context to at least only the `udsTool` folder.
=> We still have the context quite big but we at least limit it to the smallest possible set of files.

In any case, in case we adapt the context we should also adapt the `docker-compose.yaml` in the root for the local build.

nemadhu: @christophruethingbmw .dockerignore file is added limiting the context and docker-compose.yaml is also updated. Kindly let me know, if this looks okay.

aschaal: I would prefer the second option.

aschaal: After thinking again and a direct discussion I am now fine with `context: .`. In worst case we can easily optimize it later.

aschaal: Will approve. We can change later if needed.

PR #318, opened 2025-11-26
Require CC to be provided for cross-compiling

hbragge-acn:
This makes the hardcoded paths in the .cmake files unnecessary. For arm-gcc using PATH would also work, but not for clang because the names of the POSIX and ARM compilers are identical. For the sake of consistency, use the same approach regardless of the compiler.

### DIFF
--- a/cmake/toolchains/ArmNoneEabi-clang.cmake
+++ b/cmake/toolchains/ArmNoneEabi-clang.cmake
@@ -2,15 +2,11 @@ include_guard(GLOBAL)
 
 include("${CMAKE_CURRENT_LIST_DIR}/ArmNoneEabi-header.cmake")
 
-# Note: When using a toolchain file, it normally specifies the compiler itself
-# and it is unexpected that somebody tries to overwrite it using e.g. the CC
-# environment variable. However, thats exactly what we are doing at the moment,
-# therefore, we only set a compiler when the CC environment variable is not set.
-# Further note, that we only set the C compiler, since CMake deduces the others
+# NOTE: we only check the C compiler, since CMake deduces the others
 # automatically from it (respecting the order in the project call).
 
 if (NOT DEFINED CMAKE_C_COMPILER AND NOT DEFINED ENV{CC})
-    set(CMAKE_C_COMPILER "/opt/llvm-et-arm/bin/clang")
+    message(FATAL_ERROR "CC environment variable must be set")
 endif ()
 
 set(CMAKE_C_COMPILER_TARGET ${ARM_TARGET_TRIPLE})
--- a/cmake/toolchains/ArmNoneEabi-gcc.cmake
+++ b/cmake/toolchains/ArmNoneEabi-gcc.cmake
@@ -2,15 +2,11 @@ include_guard(GLOBAL)
 
 include("${CMAKE_CURRENT_LIST_DIR}/ArmNoneEabi-header.cmake")
 
-# Note: When using a toolchain file, it normally specifies the compiler itself
-# and it is unexpected that somebody tries to overwrite it using e.g. the CC
-# environment variable. However, thats exactly what we are doing at the moment,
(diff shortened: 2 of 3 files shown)


cmake/toolchains/ArmNoneEabi-clang.cmake
```diff
@@ -10,7 +10,7 @@ include("${CMAKE_CURRENT_LIST_DIR}/ArmNoneEabi-header.cmake")
 # automatically from it (respecting the order in the project call).
 
 if (NOT DEFINED CMAKE_C_COMPILER AND NOT DEFINED ENV{CC})
-    set(CMAKE_C_COMPILER "/opt/llvm-et-arm/bin/clang")
+    find_program(CMAKE_C_COMPILER clang)
```
rolandreichweinbmw: Doesn't this confuse the host clang with the cross clang?

rolandreichweinbmw: I.e. if you include the cross clang in the PATH, you fix the target build, but break the unit tests when using clang for the unit tests.

I think it's more consistent to use CC and CXX. See also https://github.com/eclipse-openbsw/openbsw/pull/319

rolandreichweinbmw: Update: My comments hold only for the PATH part.

hbragge-acn: Good point, forgot about the posix clang. I guess that's why it was originally mandatory to provide CC when building with clang. My suggestion would then be to require CC+CXX for both clang and gcc (for the sake of consistency) and remove the hardcoded paths from the .cmake files.

rolandreichweinbmw: Agree!

cmake/toolchains/ArmNoneEabi-gcc.cmake
```diff
@@ -2,15 +2,11 @@ include_guard(GLOBAL)
 
 include("${CMAKE_CURRENT_LIST_DIR}/ArmNoneEabi-header.cmake")
 
-# Note: When using a toolchain file, it normally specifies the compiler itself
-# and it is unexpected that somebody tries to overwrite it using e.g. the CC
-# environment variable. However, thats exactly what we are doing at the moment,
-# therefore, we only set a compiler when the CC environment variable is not set.
-# Further note, that we only set the C compiler, since CMake deduces the others
+# NOTE: we only check the C compiler, since CMake deduces the others
```
rolandreichweinbmw, line 5: Shouldn't we put this in cmake/toolchains/ArmNoneEabi-clang.cmake also?

rolandreichweinbmw (commented):
The s32k148 build now fails since we are missing CC and CXX in the cmake call.

cmake/toolchains/ArmNoneEabi-clang.cmake
hbragge-acn: Updated accordingly, however I only added a check for CC for now since cmake seems to figure out the rest quite well in my experience.

cmake/toolchains/ArmNoneEabi-gcc.cmake
hbragge-acn: Will do.

The CI is now failing because the elf-analyzer step calls cmake without setting CC. In order to avoid duplicating the compiler paths in the workflow file (.github/workflows/build.yml), this step would need to be moved into .ci/buildoperations.py where the other steps are. I'll look into this tomorrow.

hbragge-acn: After looking into the build scripts, I decided to just duplicate the CC in build.yml for now. It doesn't seem like the elf-analyzer step can be easily moved into .ci/build.py. For example, elf-analyzer only supports C++14 but currently there's no option to skip newer standards (without hacking). I'll try to find someone to work on that.

PR #326, opened 2025-12-08
Remove unnecessary memcopy from SocketCanTransceiver

hbragge-acn:
Copy outgoing CAN frames directly from user to the TX queue, instead of via a temporary stack object. This reduces the number of memcopies for outgoing frames (in SocketCanTransceiver) from 4 to 3.

### DIFF
--- a/platforms/posix/bsp/socketCanTransceiver/src/can/SocketCanTransceiver.cpp
+++ b/platforms/posix/bsp/socketCanTransceiver/src/can/SocketCanTransceiver.cpp
@@ -101,35 +101,29 @@ void SocketCanTransceiver::shutdown() {}
 
 ICanTransceiver::ErrorCode SocketCanTransceiver::write(CANFrame const& frame)
 {
-    if (!_writable.load(std::memory_order_relaxed))
-    {
-        return ErrorCode::CAN_ERR_ILLEGAL_STATE;
-    }
-    FrameWithListener slot{frame, nullptr};
-    ::etl::span<uint8_t> memory = _txWriter.allocate(sizeof(slot));
-    if (memory.size() == 0)
-    {
-        return ErrorCode::CAN_ERR_TX_HW_QUEUE_FULL;
-    }
-    ::std::memcpy(memory.data(), &slot, sizeof(slot));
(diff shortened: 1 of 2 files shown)


platforms/posix/bsp/socketCanTransceiver/src/can/SocketCanTransceiver.cpp
```diff
@@ -259,12 +251,13 @@ void SocketCanTransceiver::guardedRun(int maxSentPerRun, int maxReceivedPerRun)
         {
             break;
         }
-        _txReader.release();
         if (slot.listener != nullptr)
         {
             slot.listener->canFrameSent(canFrame);
         }
         notifySentListeners(canFrame);
+        slot.~FrameWithListener();
```
rolandreichweinbmw: To prevent this pattern of using placement new and directly calling the destructor, would it be possible to use etl::typed_storage instead which wraps this via .create(...) and .destroy()? (Similar to etl::typed_storage use in app.cpp)

Should be similarly efficient.

rolandreichweinbmw: etl::destroy_at or etl::destroy_object_at is also available.

hbragge-acn: Thanks for the hints, at least `destroy_at` seems useful. Using typed_storage/typed_storage_ext on the other hand doesn't seem to help in this case, because after writing the typed_storage contents into the queue, there doesn't seem to be a way to read it "on the other side" (in the guardedRun() method) without reinitializing the contents (except by using reinterpret_cast, which would defeat most of the purpose).

Btw. I turned the PR into a draft since the new approach might have alignment issues that I need to look at first (thanks @matthiaskessler for pointing it out)

hbragge-acn:
I've removed the placement-new/reinterpret_cast due to misaligned memory access. Now using the same basic approach as originally, just with one less temporary object / memcopy. (@matthiaskessler FYI)

rolandreichweinbmw (approved):

PR #328, opened 2025-12-11
Update to ETL 20.44.2 in OpenBSW

rolandreichweinbmw:
Adjusting the tests to reflect the fact that ETL doesn't support the assignment between arrays/spans of different fixed sizes anymore.

Considering the fact that it was just the two lines in the tests without substantial demonstrated use cases, I preferred to just remove this "weak use case" from the test instead of changing ETL itself.

No production code affected btw.


rolandreichweinbmw:
recheck

hbragge-acn (approved):

PR #338, opened 2025-12-30
Configure ci clang tidy report

filipe-cuim:
## Description

The goal of this PR is to configure a .clang-tidy that respects openbsw coding conventions, and a ci job that runs clang-tidy checks on all compilation units that are generated in compile_commands.json.

In this first step we start with fewer clang-tidy checks enabled, because currently with all checks enabled the repository has around 3924 findings (excluding all files in test, 3rdparty and mock) and around 23386 findings (when excluding only 3rdparty folder).

Each clang-tidy CI workflow will print a report of all findings for each checker in stdout. After that it will also upload an artifact with a yaml file generated by clang-tidy, which will contain the location (both file name and file offset) of all findings.

There may also exist some checkers that we do not want to activate, and as such it's better to do small PRs after this one, where only few checkers are considered, in order to limit the amount of changes in the repository.


.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,67 @@
...
+
+on: [ merge_group, push, pull_request ]
+
+jobs:
+  clang-tidy:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        preset: ["tests-posix-debug", "tests-posix-release", "tests-s32k1xx-debug", "tests-s32k1xx-release"]
```
simon-d-bmw: Is there any relevant difference between debug and release regarding compiler flags? If not, we could just run clang-tidy on release.

libs/bsp/bspOutputManager/src/outputManager/Output.cpp
```diff
@@ -90,31 +92,17 @@ bsp::BspReturnCode Output::invert(OutputId const chan)
...
             }
-        }
-        else
-        {
             return bsp::BSP_ERROR;
         }
-    }
-    else
-    {
         return bsp::BSP_ERROR;
```
simon-d-bmw: Those returns can be removed, too.

libs/bsp/bspInputManager/src/inputManager/DigitalInput.cpp
```diff
@@ -130,20 +130,13 @@ bsp::BspReturnCode DigitalInput::get(DigitalInputId const channel, bool& result)
...
-        else
-        {
+
             return bsp::BSP_ERROR;
         }
-    }
-    else
-    {
         return bsp::BSP_ERROR;
     }
```
simon-d-bmw: those returns could be removed.

libs/bsp/bspOutputManager/src/outputManager/Output.cpp
```diff
@@ -229,11 +211,13 @@ bsp::BspReturnCode Output::set(OutputId const chan, uint8_t const vol, bool cons
...
     }
-    else if (chan == NUMBER_OF_INTERNAL_OUTPUTS)
+
+    if (chan == NUMBER_OF_INTERNAL_OUTPUTS)
     {
         return bsp::BSP_NOT_SUPPORTED;
     }
-    else if (chan < TOTAL_NUMBER_OF_OUTPUTS)
+
+    if (chan < TOTAL_NUMBER_OF_OUTPUTS)
```
simon-d-bmw: why not 

```
if ((chan > NUMBER_OF_INTERNAL_OUTPUTS) && (chan < TOTAL_NUMBER_OF_OUTPUTS))
```

then you could remote the first if clause

libs/bsp/bspOutputManager/src/outputManager/Output.cpp
```diff
@@ -244,20 +228,11 @@ bsp::BspReturnCode Output::set(OutputId const chan, uint8_t const vol, bool cons
...
-            }
-        }
-        else
-        {
             return bsp::BSP_ERROR;
         }
-    }
-    else
-    {
         return bsp::BSP_ERROR;
```
simon-d-bmw: remove the returns?

libs/bsp/bspOutputManager/src/outputManager/Output.cpp
```diff
@@ -277,16 +252,10 @@ bsp::BspReturnCode Output::setDynamicClient(
...
-        else
-        {
-            fLock.resume();
-            return bsp::BSP_ERROR;
-        }
-    }
-    else
-    {
+        fLock.resume();
         return bsp::BSP_ERROR;
```
simon-d-bmw: this isn't needed. (not adding any more comments for similar findings)

doc/dev/guidelines/practices.rst
```diff
@@ -7,6 +7,36 @@ secure coding practices.
...
+
+  run-clang-tidy-17 -p <Path_to_compile_commands_directory> -export-fixes <Output_file_for_ct_findings> <regex_for_files_to_analyze>
+
+Alternatively you can run the same script that is used in the CI pipeline:
+
+.. code-block:: bash
+
+  python3 .ci/clang-tidy.py <Path_to_compile_commands_directory> <Output_file_for_ct_findings>
+
+In case you want to run clang-tidy directly, you can also use the following command:
```
simon-d-bmw:
```suggestion
In case you want to run clang-tidy on a single file, you can also use the following command:
```

.ci/clang-tidy.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    parser = argparse.ArgumentParser(
+        description="Run clang-tidy on a CMake build directory and analyze findings.",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="""
+Examples:
+  %(prog)s build/tests/posix/Debug ct-findings.yaml
+  %(prog)s build/tests/posix/Debug ct-findings.yaml --exclude "test|mock"
+  %(prog)s build/tests/posix/Debug ct-findings.yaml --quiet
+  %(prog)s build/tests/posix/Debug ct-findings.yaml --verbose
```
simon-d-bmw:
```suggestion
  %(prog)s --build_directory build/tests/posix/Debug --output_file ct-findings.yaml
  %(prog)s --build_directory build/tests/posix/Debug --output_file ct-findings.yaml --exclude "test|mock"
  %(prog)s --build_directory build/tests/posix/Debug --output_file ct-findings.yaml --quiet
  %(prog)s --build_directory build/tests/posix/Debug --output_file ct-findings.yaml --verbose
```

.ci/clang-tidy.py
```diff
@@ -0,0 +1,219 @@
...
+        python3 .ci/clang-tidy.py build/tests/posix/Debug ct-findings.yaml
+
+    With custom exclusion pattern (files matching 'test' or 'mock' will be excluded):
+        python3 .ci/clang-tidy.py build/tests/posix/Debug ct-findings.yaml --exclude "test|mock"
+
+    Quiet mode (no progress output from run-clang-tidy):
+        python3 .ci/clang-tidy.py build/tests/posix/Debug ct-findings.yaml --quiet
+
+    Verbose mode (show run-clang-tidy output):
+        python3 .ci/clang-tidy.py build/tests/posix/Debug ct-findings.yaml --verbose
```
simon-d-bmw: this is redundant, because the script already provides examples when being executed.

.ci/clang-tidy.py
```diff
@@ -0,0 +1,219 @@
...
+        "--verbose",
+        action="store_true",
+        help="Show run-clang-tidy progress output",
+    )
+
+    parser.add_argument(
+        "--version",
+        action="version",
+        version="%(prog)s 1.0",
+    )
```
simon-d-bmw: really needed?

.github/workflows/clang-tidy.yml
filipe-cuim: Right. I forgot to clean this up, because I ad actually already checked that it is the same. So agreed that we can just use Release.

libs/bsp/bspOutputManager/src/outputManager/Output.cpp
filipe-cuim: I think you're right. Didn't took too much consideration into it.

.ci/clang-tidy.py
filipe-cuim: No. I can delete it

libs/bsp/bspOutputManager/src/outputManager/Output.cpp
rolandreichweinbmw: @filipe-cuim : You removed the elses but not the returns?

libs/bsp/bspOutputManager/src/outputManager/Output.cpp
rolandreichweinbmw: @filipe-cuim : Same here?

libs/bsp/bspOutputManager/src/outputManager/Output.cpp
filipe-cuim: @rolandreichweinbmw this is because of one of the calng-tidys "llvm-else-after-return". We don't necessarily need to keep it. Theoretically I should only need to remove that commit, and I can suppress that warning. It's more of a code convention than something that prevents any language misuse.

filipe-cuim:
> Only put a few remarks.
> 
> Also: Are the included "Fix" commits directly related to clang-tidy findings? E.g. directly enabling a clean acceptable baseline? Otherwise, "unrelated" changes could be separated out to a cleanup PR.

Theoretically yes, each commit is respective to one clang-tidy finding. So, in principle, if I remove one commit the PR should remain clean, or at least not needed many rebase changes

doc/dev/guidelines/practices.rst
```diff
@@ -7,6 +7,36 @@ secure coding practices.
 
 The following practices apply to all of our code except when stated otherwise, regardless of the programming language.
 
+Clang-tidy
+----------
+
+Additionally, most of these practices are enforced by clang-tidy which is run as part of our CI pipeline.
+The clang-tidy configuration can be found in the file ``.clang-tidy`` in the root directory of openbsw repository.
+The ci script for running clang-tidy is located in ``.ci/clang-tidy.py``, which runs an llvm helper tool called
```
simon-d-bmw: nit:
```suggestion
The CI script for running clang-tidy is located in ``.ci/clang-tidy.py``, which runs an llvm helper tool called
```

.ci/clang-tidy.py
```diff
@@ -0,0 +1,228 @@
...
+  %(prog)s --build_directory build/tests/posix/Debug --output_file ct-findings.yaml --quiet
+  %(prog)s --build_directory build/tests/posix/Debug --output_file ct-findings.yaml --verbose
+        """,
+    )
+
+    parser.add_argument(
+        "--build_directory",
+        type=Path,
+        help="Path to the build directory containing compile_commands.json generated by CMake",
+    )
```
simon-d-bmw, line 45: nit:
```suggestion
    parser.add_argument(
        "--build_directory",
        type=Path,
        help="Path to the build directory containing compile_commands.json generated by CMake",
        required=True
    )
```

.ci/clang-tidy.py
```diff
@@ -0,0 +1,228 @@
...
+
+    parser.add_argument(
+        "--output_file",
+        type=str,
+        help="Path to the output YAML file where clang-tidy findings will be stored",
```
simon-d-bmw, line 50: nit: 
```suggestion
        help="Path to the output YAML file where clang-tidy findings will be stored",
        required=True
```

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: >-
+          DOCKER_UID=$(id -u) DOCKER_GID=$(id -g) DOCKER_HISTORY=/dev/null docker compose run --rm development
+          python3 .ci/build.py --preset ${{ matrix.preset }} --platform "linux" --cxxid "clang" --cxxstd "14" --config "${{ matrix.config }}"
+      - name: Run the clang-tidy.py inside the docker container
+        run: >-
+          DOCKER_UID=$(id -u) DOCKER_GID=$(id -g) DOCKER_HISTORY=/dev/null docker compose run --rm development
+          python3 .ci/clang-tidy.py --build_directory=build/tests/${{ matrix.platform }}/${{ matrix.config }} --output_file=ct-findings.yaml --exclude="3rdparty" --ignore-checks="clang-diagnostic-error"
+
+      - name: Upload clang-tidy findings
+        if: always()
```
simon-d-bmw: nit: can't we skip this upload if the output file doesn't exist?

simon-d-bmw (commented):
@filipe-cuim there are still 7 clang-diagnostic-errors remaining.

platforms/s32k1xx/bsp/canflex2Transceiver/src/can/transceiver/canflex2/CanFlex2Transceiver.cpp
```diff
@@ -220,6 +220,7 @@ uint16_t CanFlex2Transceiver::getHwQueueTimeout() const
 {
     // 64 = 8 byte payload
     // 53 = CAN overhead
+    // NOLINTNEXTLINE(clang-analyzer-core.DivideZero): CAN Baudrate can never be zero here.
```
matthiaskessler, line 223: @filipe-cuim , just a quick question: Have you ever experimented with `__builtin_assume()`? I know this might introduce portability issues, but I was generally curious if the could work in such cases.

libs/bsw/runtime/test/src/StatisticsContainerTest.cpp
```diff
@@ -13,6 +13,7 @@ using namespace ::testing;
 using namespace ::runtime;
 using namespace ::estd::test;
 
+// NOLINTBEGIN(cert-oop54-cpp): This is intentional for testing purposes.
```
matthiaskessler, line 16: Same here, if it's an alias, we should only check for one rule.

libs/bsw/runtime/test/src/SharedStatisticsContainerTest.cpp
```diff
@@ -16,6 +16,7 @@ using namespace ::testing;
 using namespace ::runtime;
 using namespace ::estd::test;
 
+// NOLINTBEGIN(cert-oop54-cpp): This is intentional for testing purposes.
```
matthiaskessler, line 19: According to [here](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/unhandled-self-assignment.html) the two rules are aliased. If that's the case, we should only have suppressions for one.

libs/bsp/bspCharInputOutput/src/charInputOutput/bspIo.cpp
matthiaskessler, line 19: I think, we need to review this since we are using gcc and clang.

libs/bsp/bspCharInputOutput/src/charInputOutput/bspIo.cpp
```diff
@@ -13,6 +13,7 @@ extern "C"
 {
 #endif
 
+// NOLINTBEGIN(cert-dcl51-cpp): Explanation for this suppression is below
```
matthiaskessler: According to [here](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/reserved-identifier.html)
 these rules are aliases. I don't think we should suppress each of them but decide to use only one rule.

libs/bsw/uds/include/uds/base/AbstractDiagJob.h
```diff
@@ -159,6 +159,8 @@ class AbstractDiagJob : public ::etl::uncopyable
     }
 
 #ifdef UNIT_TEST
+    static void unsetDiagSession() { sfpSessionManager = nullptr; }
```
matthiaskessler: nit: I'd call it `unsetDiagSessionManager()` to have a clear distiction between session manager and the actual session.

platforms/s32k1xx/bsp/canflex2Transceiver/src/can/transceiver/canflex2/CanFlex2Transceiver.cpp
filipe-cuim: Hello @matthiaskessler, I will try the builtin and see what clang-tidy interprets from that

libs/bsw/runtime/test/src/SharedStatisticsContainerTest.cpp
filipe-cuim: Seems reasonable to me. Will try to address that

libs/bsw/uds/include/uds/base/AbstractDiagJob.h
filipe-cuim: Seems fine to me. Will adjust that

.github/workflows/clang-tidy.yml
filipe-cuim: Changed it to upload only when the step of clang-tidy fails.

platforms/s32k1xx/bsp/canflex2Transceiver/src/can/transceiver/canflex2/CanFlex2Transceiver.cpp
filipe-cuim: Hey @matthiaskessler, problem with this one is that gcc doesn't support this builtin, so I would either need to do some ifdefs for gcc and clang, or find a different alternative that could suit them both. This could also maybe be solved with C++ [[assume]] but it's a C++23 feature unfortunately.

libs/bsw/runtime/test/src/StatisticsContainerTest.cpp
filipe-cuim: Disabled these certs globally.

libs/bsp/bspCharInputOutput/src/charInputOutput/bspIo.cpp
matthiaskessler: Will do this offline

simon-d-bmw:
this closes #338

PR #341, opened 2026-01-08
Document PR tags regarding hardware tests

rolandreichweinbmw:
As discussed in https://github.com/eclipse-openbsw/openbsw/discussions/317, we add the labels `tested_on_hw` and `no_hw_test_required`.


rolandreichweinbmw:
Please note that we need to agree on a definition of "tested on hardware". Currently, the PR proposes a simple "sanity check" involving a simple bootup without crashing.

An alternative would be `pytest --target=s32k148 --no-restart`. But that requires a more complex setup involving CAN and 100BASE-T1 test hardware. What do you think?

PR #342, opened 2026-01-09
Adapt CMake to support multiple executables

amit-warbhe:
Adapt CMake to support multiple executables

Before this change, the boolean variables BUILD_REFERENCE and
BUILD_UNIT_TESTS were used to select either the reference app or the unit tests.
These variable are replaced by BUILD_EXECUTABLE which is now a string variable.
It can be set to "referenceApp", "unitTest" or any string referring to other(future)
executables.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,9 +24,20 @@ set(BUILD_TARGET_RTOS
     "FREERTOS"
     CACHE STRING "Target RTOS")
 
+set(BUILD_EXECUTABLE
+    ""
+    CACHE STRING "Which build to enable: referenceApp, unitTest")
+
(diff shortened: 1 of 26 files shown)


CMakeLists.txt
```diff
@@ -24,9 +24,20 @@ set(BUILD_TARGET_RTOS
     "FREERTOS"
     CACHE STRING "Target RTOS")
 
+set(BUILD_EXECUTABLE
+    ""
+    CACHE STRING "Which build to enable: referenceApp, unitTest or empty")
```
rolandreichweinbmw: When "empty" is a valid option, how is it defined?

amit-warbhe: Removed it, thank you.

rolandreichweinbmw (approved):

PR #356, opened 2026-01-23
Add Rust build system support

M-Hassanin:
Add Rust build system support
Enable mixed C++/Rust development.

Build System:
- Add BUILD_RUST CMake option (opt-in, default OFF)
- Integrate Corrosion using rim (vendored in libs/3rdparty/corrosion/)
- Configure cross-compilation for S32K148 (thumbv7em-none-eabihf)
- Use manual C header for FFI (cbindgen installed for future use)

CMake Presets:
- posix-rust: POSIX platform with Rust and FreeRTOS support
- s32k148-rust-gcc: S32K148 with Rust and FreeRTOS support

Example Crate:
- Add rustHelloWorld example crate demonstrating FFI
- Use no_std for embedded compatibility
- Export rust_add() function callable from C++

CI/CD:
- Update Dockerfile with Rust 1.90.0, cbindgen 0.27.0, and ARM target
- Add posix-rust and s32k148-rust-gcc to CI build matrix

Documentation:
- Add Rust setup instructions for POSIX and S32K148 builds
- Pin Rust version to 1.90.0 for CI compatibility
- Update NOTICE.md with Corrosion license info


CMakePresets.json
```diff
@@ -207,6 +240,18 @@
...
         },
+        {
+            "name": "posix-rust",
+            "displayName": "build POSIX with Rust",
+            "description": "Build reference application for POSIX-compliant environment with Rust support (Release by default; for Debug use --config Debug)",
+            "configurePreset": "posix-rust"
+        },
+        {
+            "name": "s32k148-rust-gcc",
+            "displayName": "S32K148 with Rust build (GCC)",
```
rolandreichweinbmw: Maybe mention that the new presets use FreeRTOS. (GCC is already mentioned.)

docker/development/Dockerfile
```diff
@@ -44,6 +44,14 @@ RUN wget -O treefmt.tar.gz https://github.com/numtide/treefmt/releases/download/
 
 RUN gem install esr-rim
 
+# Install Rust toolchain
+ENV RUSTUP_HOME=/opt/rustup
+ENV CARGO_HOME=/opt/cargo
+ENV PATH=/opt/cargo/bin:${PATH}
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable --profile minimal \
```
christian-schilling: This installs an unpinned Rust version. We don't want that.
Every external dependency must be pinned to an exact version.
That also applies to toolchains.

CMakePresets.json
rolandreichweinbmw: Thank you for addressing it in this location. However, there are more locations in CMakePresets.json. Please have a look.

executables/referenceApp/application/src/app/app.cpp
```diff
@@ -213,7 +219,11 @@ void staticInit()
 void run()
 {
     staticInit();
-    etl::print("hello\r\n");
```
rolandreichweinbmw, line 216: Can you please tell why? etl::print() was used intentionally here to have better compatibility with modern C++, Python, and Rust format strings.

doc/dev/learning/setup/setup_s32k148_ubuntu_build.rst
```diff
@@ -107,4 +107,22 @@ The build files should be written to a new subdirectory named ``build/s32k148-gc
...
+.. code-block:: bash
+
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+    rustup target add thumbv7em-none-eabihf
+
+Then build using the Rust preset:
+
+.. code-block:: bash
+
+    cmake --preset s32k148-rust-gcc
```
rolandreichweinbmw, line 129: This works only if the gcc toolchain is set up already also. E.g. via installing and setting CC and CXX. I suggest referring to the respective documentation of S32K148+GCC+FreeRTOS build.

executables/referenceApp/rustHelloWorld/include/rust_hello_world.h
```diff
@@ -0,0 +1,24 @@
...
+extern "C"
+{
+#endif
+
+/// Adds two numbers and returns the result.
+///
+/// \param a First number to add
+/// \param b Second number to add
+/// \return The sum of a and b
+uint32_t rust_add(uint32_t a, uint32_t b);
```
rolandreichweinbmw, line 18: This example shows a C-only interface use case.

For interfacing between C++ and Rust, would it be possible to give a more complex example, e.g. mapping between C++ class and Rust crate interface? Maybe put it on the linked agenda/Rust discussion, but we would need it at some point to support reasonably complex interfaces.

executables/referenceApp/application/src/app/app.cpp
M-Hassanin: As agreed in the meeting, changed back to etl::print.

christian-schilling (changes requested):
The commit message needs to be adjusted according the the guidelines.

executables/referenceApp/rustHelloWorld/include/rust_hello_world.h
M-Hassanin: Good point! This initial PR intentionally uses a minimal C-style interface (extern "C") to establish the basic Rust build infrastructure with Corrosion. The upcoming features in the agenda will include more interfaces between C++ and Rust.

rolandreichweinbmw:
Please note that if you changed the PR description, you need to change the commit message also, separately.

M-Hassanin:
@rolandreichweinbmw, can we merge?

Issue #194, opened 2025-09-24
Unified communication framework (Service-oriented, Signal-based)


simon-d-bmw:
**Describe the feature**
OpenBSW shall provide a communication framework which supports time-based, data-driven, and request-driven applications with safety and security properties.
The C++ communication API offered to applications shall abstract from the binding used for a specific communication channel.
The communication framework shall provide bindings to SOME/IP, shared memory, and signal-based communication buses. 

**Proposed Solution**

Issue #86 requests a middleware, supporting communication paradigm such as methods, events, and fields for communication between internal applications.
To support these paradigms, the middleware implements basic techniques such as shared memory communication for internal applications, including queuing, serialization, synchronization and notification mechanisms.

This solution uses the middleware as core part for a unified communication framework in OpenBSW, providing support for service-oriented, and signal-based communication.
Specifically, this solution plans to introduce:

* Translation component (pdu2Svc) which transforms protocol data units (PDUs) into services and vice versa.
  * From a middleware transport layer perspective, PDUs are represented as service fields of type `UInt8[]`.
* Extensions running in the application context are responsible for the serialization, transmission and reception management, end-to-end protection.

The benefits of this solution are:

*Reuse*
  When using the unified communication framework, basic algorithms such as synchronization, queueing,notification and serialization can be reused for all kinds of communication techniques.

*Modularity*
  The approach is modular because it performs all relevant operations for COM processing in the application context.
  Changes in the layout of one PDU only affect local interfaces and classes, and does not require a reconfiguration or rebuild of a central basic software module.

*Freedom from interference*
  From a functional safety perspective, it is beneficial to perform the end-to-end protection of messages close to the producers and consumers of the data to prevent any corruption during transit.
  This isolation of safety-critical operations in the application context vastly reduces the amount of safety-critical basic software code.
  ``pdu2svc``, and the ``msg router`` are not safety-critical, and can therefore be executed in a QM context, because the end-to-end protection in the application context detects any corruption.

*Flexibility*
   This solution transfers the raw messages up to the application.
   Therefore, applications can flexibly switch between using the raw message, or its signals.

**Additional context**

We already have demonstrated the viability of this solution in a Proof of Concept, which includes:

* An end-to-end protected, cyclic, received PDU translated to a service attribute
* A service attribute translated into a cyclically transmitted, end-to-end protected PDU
* A pdu2svc module
* Extension class for the sending application, i.e. skeleton extension, providing a signal-based API.
[...]

PR #363, opened 2026-02-04
Add middleware proxy, skeleton and cluster connections

dcavalei:
Add middleware proxy, skeleton and cluster connections

See also: #194


libs/bsw/middleware/include/middleware/core/MessageAllocator.h
```diff
@@ -34,36 +34,44 @@ class MessageAllocator
...
-            return MessageAllocator::allocateTrivialType_(
-                &obj, sizeof(obj), msg, numberOfReferences);
+            T::AllocationPolicy::serialize(obj, buffer.value());
+            ret = HRESULT::Ok;
         }
+
+        return ret;
+    }
+
+    // Overload 2: Trivially copyable types that fit in MAX_PAYLOAD_SIZE
```
dcavalei: 😢

libs/bsw/middleware/include/middleware/concurrency/LockStrategies.h
```diff
@@ -0,0 +1,91 @@
...
+ */
+#define MIDDLEWARE_SINGLE_CORE_LOCK middleware::concurrency::ScopedCoreLock();
+
+/**
+ * \def MIDDLEWARE_ECU_LOCK
+ * \brief Macro for creating an ECU-wide scoped lock.
+ * \details Creates a ScopedECULock object that will protect the current scope with an ECU-wide
+ * (multi-core) lock. The lock is automatically released when the scope exits.
+ */
+#define MIDDLEWARE_ECU_LOCK middleware::concurrency::ScopedECULock();
```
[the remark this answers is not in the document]
dcavalei: I confirmed and this is indeed a bug, it the macro does not compile when used.
About copilot suggestion, we can not use C++17 syntax but we should look carefully to `::concurrency`, we have to ensure that things are not optimized out by  -O flags.

rolandreichweinbmw:
Hi Copilot review raised some valid points.

Please also adjust the PR and commit messages according to the guidelines at
https://github.com/eclipse-openbsw/openbsw/blob/main/doc/dev/guidelines/commit_message.rst

Squash commits where reasonable (as hinted by same descriptions).

libs/bsw/middleware/include/middleware/core/TransceiverContainer.h
```diff
@@ -0,0 +1,92 @@
...
+    TransceiverContainer(TransceiverContainer const&)            = delete;
+    TransceiverContainer& operator=(TransceiverContainer const&) = delete;
+    TransceiverContainer(TransceiverContainer&&)                 = delete;
+    TransceiverContainer& operator=(TransceiverContainer&&)      = delete;
+
+    etl::ivector<ITransceiver*>* const
+        fContainer;            ///< Pointer to the vector holding transceiver pointers
+    uint16_t const fServiceid; ///< The service ID associated with this container
+    uint16_t fActualAddress;   ///< The current/next available address ID for proxies
+};
```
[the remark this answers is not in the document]
dcavalei: false positive

libs/bsw/middleware/include/middleware/concurrency/LockStrategies.h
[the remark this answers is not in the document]
dcavalei: fixed

libs/bsw/middleware/test/src/time/mock/SystemTimerProviderMock.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+uint32_t getCurrentTimeInUs()
+{
+    if (gSystemTimerProviderMockPtr != nullptr)
+    {
+        return gSystemTimerProviderMockPtr->getCurrentTimeInUs();
+    }
+
+    assert(false);
+}
```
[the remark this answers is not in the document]
dcavalei: fixed

dcavalei:
please do **NOT MERGE**, thanks

libs/bsw/middleware/src/middleware/core/ProxyBase.cpp
```diff
@@ -0,0 +1,159 @@
...
+            fConnection->getTargetClusterId(),
+            getAddressId());
+    }
+    return Message::createFireAndForgetRequest(
+        getServiceId(),
+        memberId,
+        getInstanceId(),
+        fConnection->getSourceClusterId(),
+        fConnection->getTargetClusterId());
+}
```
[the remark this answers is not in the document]
dcavalei: fixed

libs/bsw/middleware/src/middleware/core/IClusterConnectionConfigurationBase.cpp
```diff
@@ -0,0 +1,158 @@
...
+void ITimeoutConfiguration::unregisterTimeoutTransceiver(
+    ITimeout& transceiver, ::etl::ivector<ITimeout*>& timeoutTransceivers)
+{
+    using ETL_OR_STD::swap;
+
+    MIDDLEWARE_SINGLE_CORE_LOCK
+    {
+        auto* const iter
+            = etl::find(timeoutTransceivers.begin(), timeoutTransceivers.end(), &transceiver);
+        if (iter != timeoutTransceivers.cend())
```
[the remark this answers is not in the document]
dcavalei: fix

libs/bsw/middleware/src/middleware/core/IClusterConnectionConfigurationBase.cpp
```diff
@@ -0,0 +1,158 @@
...
+    meta::TransceiverContainer const* const proxiesEnd,
+    Message const& msg)
+{
+    HRESULT result = HRESULT::Ok;
+
+    if (msg.isEvent())
+    {
+        auto const range = meta::DbManipulator::getTransceiversByServiceIdAndServiceInstanceId(
+            proxiesStart, proxiesEnd, msg.getHeader().serviceId, msg.getHeader().serviceInstanceId);
+        for (auto const* it = range.first; it != range.second; it = etl::next(it))
```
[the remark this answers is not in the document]
dcavalei: fix

libs/bsw/middleware/src/middleware/core/ProxyBase.cpp
```diff
@@ -0,0 +1,159 @@
...
+            proxyCc.end(),
+            [sourceCluster](IClusterConnection const* const clusConn)
+            {
+                if (clusConn != nullptr)
+                {
+                    return (clusConn->getSourceClusterId() == sourceCluster);
+                }
+                return false;
+            });
+        if (ccIt != proxyCc.end())
```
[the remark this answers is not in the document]
dcavalei: fix

libs/bsw/middleware/src/middleware/core/DatabaseManipulator.cpp
```diff
@@ -0,0 +1,356 @@
...
+    {
+        internal::DummyTransceiver const dummy(instanceId);
+        return etl::equal_range(
+            transceiversById->fContainer->cbegin(),
+            transceiversById->fContainer->cend(),
+            &dummy,
+            TransceiverContainer::TransceiverComparatorNoAddressId());
+    }
+
+    return etl::make_pair(start->fContainer->cbegin(), start->fContainer->cbegin());
```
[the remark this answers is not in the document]
dcavalei: To avoid deviations with the Internal implementation, I didn't do logic changes!

libs/bsw/middleware/include/middleware/core/DatabaseManipulator.h
```diff
@@ -0,0 +1,286 @@
...
+     */
+    explicit DummyTransceiver(
+        uint16_t const instanceId, uint16_t const addressId = INVALID_ADDRESS_ID)
+    : Base(instanceId), fAddressId(addressId)
+    {}
+
+    virtual ~DummyTransceiver() = default;
+
+private:
+    uint16_t fAddressId; ///< The address ID of the dummy transceiver
```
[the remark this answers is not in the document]
dcavalei: done

libs/bsw/middleware/test/src/logger/DslLogger.h
```diff
@@ -0,0 +1,105 @@
...
+    void setup() {}
+
+    void teardown() {}
+
+    template<typename... Args>
+    void EXPECT_LOG(LogLevel const level, std::string const& format, Args... args)
+    {
+        etl::vector<uint32_t, sizeof(uint32_t) * sizeof...(Args)> vec;
+        push_all(vec, args...);
+
```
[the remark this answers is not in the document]
dcavalei: done

libs/bsw/middleware/include/middleware/logger/Logger.h
```diff
@@ -0,0 +1,69 @@
...
+enum class Error : uint8_t
+{
+    Allocation                  = 0U, ///< Memory allocation failure
+    Deallocation                = 1U, ///< Memory deallocation failure
+    ProxyInitialization         = 2U, ///< Proxy initialization failure
+    SkeletonInitialization      = 3U, ///< Skeleton initialization failure
+    DispatchMessage             = 4U, ///< Message dispatch failure
+    SendMessage                 = 5U, ///< Message sending failure
+    ProxyCrossThreaViolation    = 6U, ///< Proxy cross-thread access violation
+    SkeletonCrossThreaViolation = 7U, ///< Skeleton cross-thread access violation
```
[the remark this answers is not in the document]
dcavalei: done

libs/bsw/middleware/include/middleware/concurrency/LockStrategies.h
```diff
@@ -0,0 +1,78 @@
...
+namespace concurrency
+{
+
+/**
+ * \brief Suspend all interrupts.
+ * \details Platform-specific function that suspends all interrupts to ensure critical sections
+ * are protected. This is used in conjunction with lock strategies for thread-safe operations.
+ * The implementation must be provided for each platform integration.
+ */
+extern void suspendAllInterrupts();
```
matthiaskessler, line 18: This function seems quite dangerous. The comment is quite generic as if it's a took for critical sections. In reality there is no counterpart to `suspendAllInterrupts()` to release them again.
The only usage I found is in error handling with a following `ETL_ASSERT_FAIL`.
We have two other abstractions for locking interrupts:
- `async::Lock` and `async::ModifiableLock`
- `interrupts::SuspendResumeAllInterruptsLock`

I personally would like to get rid of direct usages of `bspInterrupts` in the long run and rely on `async` locking since `async` is our chosen OS abstraction.

libs/bsw/middleware/include/middleware/concurrency/LockStrategies.h
```diff
@@ -0,0 +1,78 @@
...
+     * synchronization mechanism.
+     */
+    ~ScopedCoreLock();
+
+    ScopedCoreLock(ScopedCoreLock const&)            = delete;
+    ScopedCoreLock& operator=(ScopedCoreLock const&) = delete;
+};
+
+/**
+ * \brief Scoped lock for ECU-wide (multi-core) protection.
```
matthiaskessler: Just a question: Do we really want to use this `\brief` tag? I though we would use auto brief in doxygen, which IMHO increases readability of comment blocks a lot.

libs/bsw/middleware/include/middleware/core/ClusterConnection.h
```diff
@@ -0,0 +1,445 @@
...
+     * \details Registers a proxy with the specified service instance ID to receive messages.
+     *
+     * \param proxy reference to the proxy to subscribe
+     * \param serviceInstanceId the service instance ID for the subscription
+     * \return HRESULT indicating success or failure of the subscription
+     */
+    HRESULT subscribe(ProxyBase& proxy, uint16_t const serviceInstanceId) final;
+
+    /**
+     * \brief Subscribe a skeleton (not implemented).
```
matthiaskessler: NIT: IMHO this is too verbose for just a not implemented function. This is no criticism to this particular function but a general topic that we should address at some time in our guidelines.

libs/bsw/middleware/include/middleware/core/ClusterConnectionBase.h
```diff
@@ -0,0 +1,149 @@
...
+     * \brief Main message processing function.
+     * \details Processes incoming messages and releases message resources at the end of the
+     * function. This is the entry point for handling messages received from other clusters.
+     *
+     * \param msg constant reference to the message to process
+     */
+    void processMessage(Message const& msg) const override;
+
+    /**
+     * \brief Get the count of registered transceivers for a service.
```
matthiaskessler: NIT: I think, we should discuss together about comments for "trivial" functions. E.g. I would write here 
`/** Returns count of registered transceivers for a \p serviceId. */`

This is just one line and contains the same information.

libs/bsw/middleware/include/middleware/core/ClusterConnectionBase.h
```diff
@@ -0,0 +1,149 @@
...
+     * \brief Constructor for ClusterConnectionBase.
+     * \param configuration reference to the cluster connection configuration
+     */
+    explicit ClusterConnectionBase(IClusterConnectionConfigurationBase& configuration);
+
+    /**
+     * \brief Get the source cluster ID.
+     * \return the source cluster ID
+     */
+    uint8_t getSourceClusterId() const override { return fConfiguration.getSourceClusterId(); }
```
matthiaskessler, line 56: NIT: I don't think, it makes sense to duplicate comments from the interface. Maybe
`/** \see IClusterConnection::getSourceClusterId() */` is sufficient if no new information is added.

libs/bsw/middleware/include/middleware/core/ClusterConnectionBase.h
```diff
@@ -0,0 +1,149 @@
...
+     * \details Sends an error response back to the sender when an error occurs during message
+     * processing.
+     *
+     * \param error the error state to send
+     * \param msg constant reference to the original message
+     */
+    void respondWithError(ErrorState const error, Message const& msg) const;
+
+    IClusterConnectionConfigurationBase&
+        fConfiguration; ///< Reference to the cluster connection configuration
```
matthiaskessler: NIT: I personally don't like end of line comments. They lead often to line breaks when auto formatting files and make code more unreadable compared to putting comments above the code. Again, this doesn't mean you need to change it here, we should discuss and come up with a common guideline.

libs/bsw/middleware/include/middleware/core/IClusterConnection.h
```diff
@@ -0,0 +1,139 @@
...
+ */
+class ITimeoutClusterConnection
+{
+public:
+    /**
+     * \brief Register a timeout transceiver.
+     *
+     * \param timeout reference to the ITimeout transceiver to register
+     */
+    virtual void registerTimeoutTransceiver(ITimeout&) {}
```
matthiaskessler: Strictly speaking this is no interface because you are providing a default implementation. That's fine, but I think the class should be named differently then, e.g. `TimeoutClusterConnection` without the leading `I`.

libs/bsw/middleware/include/middleware/core/DatabaseManipulator.h
```diff
@@ -0,0 +1,286 @@
...
+
+namespace internal
+{
+/**
+ * \brief Dummy transceiver implementation for testing and placeholder purposes.
+ * \details This transceiver provides minimal functionality and is typically used as a
+ * placeholder in the transceiver database. It returns default values for most operations
+ * and does not perform actual message processing.
+ */
+struct DummyTransceiver final : public ITransceiver
```
matthiaskessler: NIT: I know this is not done everywhere, but to me, using `struct` hints that this is a POD like datatype. Otherwise I would recommend to use `class`.

libs/bsw/middleware/include/middleware/core/IClusterConnectionConfigurationBase.h
```diff
@@ -0,0 +1,497 @@
...
+     */
+    virtual void updateTimeouts() = 0;
+
+    ITimeoutConfiguration(ITimeoutConfiguration const&)            = delete;
+    ITimeoutConfiguration& operator=(ITimeoutConfiguration const&) = delete;
+    ITimeoutConfiguration(ITimeoutConfiguration&&)                 = delete;
+    ITimeoutConfiguration& operator=(ITimeoutConfiguration&&)      = delete;
+
+protected:
+    virtual ~ITimeoutConfiguration() = default;
```
matthiaskessler, line 140: According to C++ Core Guidelines (C.35): a base class destructor should be either public and virtual or protected and non-virtual

libs/bsw/middleware/include/middleware/core/ITransceiver.h
```diff
@@ -0,0 +1,111 @@
...
+     */
+    virtual uint8_t getSourceClusterId() const = 0;
+
+    ITransceiver(ITransceiver const&)            = delete;
+    ITransceiver& operator=(ITransceiver const&) = delete;
+    ITransceiver(ITransceiver&&)                 = delete;
+    ITransceiver& operator=(ITransceiver&&)      = delete;
+
+protected:
+    uint16_t instanceId_; ///< The service instance ID for this transceiver
```
matthiaskessler: Again, strictly speaking this is no interface but an abstract base class.

libs/bsw/middleware/include/middleware/core/ITimeout.h
```diff
@@ -0,0 +1,29 @@
...
+{
+namespace core
+{
+/**
+ * \brief Interface for timeout management.
+ * \details This interface defines the contract for objects that need to handle timeout
+ * updates. Implementations of this interface can register with cluster connections to
+ * receive periodic timeout notifications and process expired timeouts accordingly.
+ */
+struct ITimeout
```
matthiaskessler: NIT: I suggest using `class` and reserve `struct` for POD like types. Also the name `ITimeout` is not quite self explaining...the purpose of the interface seems more like a timeout handler (although I know handler is also a very generic name ;-) )

libs/bsw/middleware/include/middleware/core/LoggerApi.h
```diff
@@ -0,0 +1,149 @@
...
+#include "middleware/core/Message.h"
+#include "middleware/core/types.h"
+#include "middleware/logger/Logger.h"
+
+namespace middleware
+{
+namespace logger
+{
+
+ETL_INLINE_VAR constexpr uint8_t ALLOCATION_FAILURE_LOG_SIZE
```
matthiaskessler: NIT: end of line comments reduce readability IMHO.

libs/bsw/middleware/include/middleware/core/ProxyBase.h
```diff
@@ -0,0 +1,129 @@
...
+{
+public:
+    /**
+     * \brief Set the address ID for this proxy.
+     * \details Updates the unique address identifier for this proxy instance, used for routing
+     * response messages back to this specific proxy.
+     *
+     * \param addressId the new address ID to set
+     */
+    void setAddressId(uint8_t const addressId) final;
```
matthiaskessler: I think, `const` for primitive types passed by value in a function declaration will create a compiler warning.

libs/bsw/middleware/include/middleware/time/SystemTimerProvider.h
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * \brief Get the current system time in milliseconds.
+ * \details Platform-specific function that returns the current system time in milliseconds.
+ * This is used for timeout management and timestamp operations in the middleware. The
+ * implementation must be provided for each platform integration and should return a monotonic
+ * timestamp.
+ *
+ * \return the current system time in milliseconds
+ */
+extern uint32_t getCurrentTimeInMs();
```
matthiaskessler, line 20: I think, long term we should decide if `middleware` wants to keep these abstractions or if it can rely on the ones OpenBSW provides.

libs/bsw/middleware/src/middleware/core/ClusterConnectionBase.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include "middleware/core/ClusterConnectionBase.h"
+
+#include <cstddef>
+#include <cstdio>
+
+#include "middleware/core/IClusterConnectionConfigurationBase.h"
+#include "middleware/core/ITimeout.h"
+#include "middleware/core/LoggerApi.h"
+#include "middleware/core/Message.h"
+// #include "middleware/core/MessageAllocator.h"
```
matthiaskessler, line 12: Maybe we should remove that line if it's unused?

libs/bsw/middleware/src/middleware/core/ClusterConnectionBase.cpp
```diff
@@ -0,0 +1,125 @@
...
+{
+
+ClusterConnectionBase::ClusterConnectionBase(IClusterConnectionConfigurationBase& configuration)
+: fConfiguration(configuration)
+{}
+
+void ClusterConnectionBase::processMessage(Message const& msg) const
+{
+    static_cast<void>(dispatchMessage(msg));
+    // MessageAllocator::deallocate(msg);
```
matthiaskessler, line 28: Is this unused code or a TODO for later?

libs/bsw/middleware/src/middleware/core/DatabaseManipulator.cpp
```diff
@@ -0,0 +1,347 @@
...
+        res = HRESULT::ServiceIdOutOfRange;
+    }
+    else
+    {
+        auto const serviceId = proxy.getServiceId();
+        auto containerIt     = getTransceiversByServiceId(start, end, serviceId);
+        if (containerIt != end)
+        {
+            auto it = DbManipulator::findTransceiver(&proxy, *containerIt->fContainer);
+            if (it != containerIt->fContainer->end())
```
matthiaskessler: Maybe it makes sense to extract `containerIt->fContainer` into a local variable? It's used quite often in the next lines.

libs/bsw/middleware/src/middleware/core/LoggerApi.cpp
```diff
@@ -0,0 +1,188 @@
...
+}
+
+void serialize(etl::byte_stream_writer& writer, uint32_t const value)
+{
+    writer.write_unchecked(value);
+}
+
+void serialize(etl::byte_stream_writer& writer, core::Message const& value)
+{
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers)
```
matthiaskessler, line 39: Maybe a comment, why exactly 10 is needed would help the reader to understand better. E.g. does the `writer` have exactly 10 bytes reserved for it? Because I have kind of a hear time understanding the static assert since we are accessing member variables of the message header later and I assume the assert makes sure, that the sum of bytes resulting from writing these variables must not exceed 10.

libs/bsw/middleware/include/middleware/concurrency/LockStrategies.h
dcavalei: @matthiaskessler , the `suspendAllInterrupts()` is a function that must be implemented by the platform, and not a direct call to a BSP function. There is not counterpart to `suspendAllInterrupts()` because there is no intention of recovering to the previous state, this function is used before a panic situation, without recovery.

I propose to look at Middleware platform dependencies (interfaces like `suspendAllInterrupts()`) and see if there are existing abstractions, but only after the Initial porting. Do you agree?

libs/bsw/middleware/include/middleware/concurrency/LockStrategies.h
dcavalei: I was not aware, I'll just remove the `\brief` tag

libs/bsw/middleware/include/middleware/time/SystemTimerProvider.h
dcavalei: I proposed this as well in some other comment.

libs/bsw/middleware/include/middleware/core/IClusterConnectionConfigurationBase.h
dcavalei: @matthiaskessler , this is a consequence of the current design unfortunately, there is a lot of inheritance.

According to my understanding of the code, the reason why it is `protected`, is to enforce that only the final derived class destructor can be called by the developer, and not an intermediate interface destructor that will not call the derived class destructor. I interpret this as a guard, I'd prefer to keep it as is. I think the guideline is not good to follow for multi-level inheritance of interfaces...

This multi-level inheritance design is smelly, and we are working on a new design pattern.

libs/bsw/middleware/src/middleware/core/ClusterConnectionBase.cpp
dcavalei: TODO

libs/bsw/middleware/src/middleware/core/ClusterConnectionBase.cpp
dcavalei: no external allocation yet, lets keep it as is so that we don't forget.

libs/bsw/middleware/src/middleware/core/LoggerApi.cpp
dcavalei: There was a bug in this assert, thanks for noticing, even if looking into another topic.

Honestly I don't know why this assert exists, I will try to understand if it should be here.

libs/bsw/middleware/include/middleware/concurrency/LockStrategies.h
matthiaskessler: I agree and suggest to create an issue so that we don't forget. And what I wrote about the missing counterpart was more about the comment above the function. It's reads as if this is a general purpose function that can be used for thread safe operations. While it's real purpose is only in error handling. So I would suggest to change the comment but leave the other abstractions investigation for later.

libs/bsw/middleware/include/middleware/concurrency/LockStrategies.h
matthiaskessler: 👍

libs/bsw/middleware/src/middleware/core/ClusterConnectionBase.cpp
matthiaskessler: Maybe for the future we should mark it as one ;-)

libs/bsw/middleware/src/middleware/core/LoggerApi.cpp
matthiaskessler: Maybe we should also create an issue to not forget.

dcavalei: I asked internally and it was just to assert that its 10 bytes, i fixed the bug, nothing todo here

PR #383, opened 2026-03-03
Improve incomingDiagConnection interfacing

guillaumeaub-ctw:


christian-schilling (changes requested):
I'm currently working on a big refactoring of UDS and as part of that refactoring aim to remove the use of interfaces, virtual and mocks as much as possible. This change is in direct conflict with those goals.
What is the underlying need/problem you are trying to address? I'm sure we can find a better solution.

PR #411, opened 2026-03-26
Keep lwIP netif persistent across lifecycle transitions

amit-warbhe:
Keep lwIP netif persistent across lifecycle transitions

After Lifecycle transition 9 → 4 → 9, UDP multicast join failed with:
igmp_joingroup failed with status -6
because EthernetSystem::init() was only executed once during system lifetime,
while run() is re-executed after lifecycle restart.

Network interface configuration was performed inside init().
During LC restart (4 → 9), init() was not called again, therefore
the lwIP netif was not properly reconfigured.

Moving all netif-specific configuration and activation logic into run() solves the issue.

### DIFF
--- a/executables/referenceApp/application/src/systems/EthernetSystem.cpp
+++ b/executables/referenceApp/application/src/systems/EthernetSystem.cpp
@@ -112,26 +112,6 @@ EthernetSystem::EthernetSystem(
 void EthernetSystem::init()
 {
     lwip_init();
-
-    for (size_t i = 0; i < netifs.netifs.size(); ++i)
-    {
-        if (!lwipnetif::initNetifIp4(
-                netifs.netifs[i], netifs.ip4Configs[i], netifs.networkInterfaceConfigsIp4[i], this))
-        {
-            continue;
-        }
-        netifs.netifStates[i].state = ::lwipnetif::State::Initialised;
-        netifs.netifs[i].linkoutput = &linkoutput;
-
-        ::lwiputils::initNetifDriverParameters(::ethX::MAC_ADDRESS, netifs.netifs[i]);
-
-        auto const lwipNetif = &netifs.netifs[i];
-#if LWIP_IGMP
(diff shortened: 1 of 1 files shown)


executables/referenceApp/application/src/systems/EthernetSystem.cpp
```diff
@@ -101,37 +101,51 @@ EthernetSystem::EthernetSystem(
...
+#if LWIP_IGMP
+            netif_set_igmp_mac_filter(&netif, joinMulticastGroupIpV4);
+#endif
+
+            state = ::lwipnetif::State::Initialised;
+        }
+
+        if (state == ::lwipnetif::State::Initialised)
+        {
+            ::lwipnetif::start(netif, netifs.ip4Configs[i]);
```
christian-schilling, line 144: `start` was already in `::run()` so this should have worked.
Could it be that `stop` is also deinitializing? Maybe the poper fix would be to check what happens in stop.

christian-schilling: I checked: The asymetry is that while `netif_add` happens in `init*`, the `netif_remove` happens in stop.
So the correct fix is to create a `shutdown` function in lwipnetif and remove the `netif_remove` from the `stop` function.

amit-warbhe: As per our **lifecycle** design, 
The init() method of a lifecycle component is only ever called once when the lifecycle component's runlevel is reached from a lower level, whereas run() is called every time the lifecycle component's runlevel is reached from a lower level.
so the fix that I applied works.

matthiaskessler: The fix resolves the problem with the lifecycle-test. I think, we can do a redesign of the startup/shutdown later.

matthiaskessler (dismissed):

christian-schilling (dismissed):

matthiaskessler (approved):

PR #412, opened 2026-03-26
Fix ENET ISR race during Ethernet driver initialization

amit-warbhe:
Fix ENET ISR race during Ethernet driver initialization

The ENET ISR could access the Ethernet driver before it was fully initialized.

The ethernet driver pointer (pEnetDriver) was assigned during init(),
which exposed the driver to interrupts before driver.start() completed.

If an ENET interrupt occurred during this window, the ISR could call interruptGroup0() on a
partially initialized driver, leading to undefined behaviour which resulted in HardFault or sometimes
system reset.

Move the assignment of pEnetDriver from init() to run() so the driver is only exposed to the ISR
after driver.start() completes.

### DIFF
--- a/executables/referenceApp/platforms/s32k148evb/main/src/systems/S32K148EvbEthernetSystem.cpp
+++ b/executables/referenceApp/platforms/s32k148evb/main/src/systems/S32K148EvbEthernetSystem.cpp
@@ -112,7 +112,6 @@ S32K148EvbEthernetSystem::S32K148EvbEthernetSystem(
 void S32K148EvbEthernetSystem::init()
 {
     _tja1101LinkStatus = false;
-    pEnetDriver        = &_driver;
     configureEnetPins();
     _driver.init();
     _tja1101.start();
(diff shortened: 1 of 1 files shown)


executables/referenceApp/platforms/s32k148evb/main/src/systems/S32K148EvbEthernetSystem.cpp
```diff
@@ -112,7 +112,6 @@ S32K148EvbEthernetSystem::S32K148EvbEthernetSystem(
 void S32K148EvbEthernetSystem::init()
 {
     _tja1101LinkStatus = false;
-    pEnetDriver        = &_driver;
     configureEnetPins();
     _driver.init();
```
christian-schilling, line 116: Would it be sufficient to just move the assignment after `_driver.init()`?

amit-warbhe: I tried this but it does not solve the hard fault observed in pytest for lc level change from 1->9, when tested with multiple iterations.

christian-schilling (approved):

christian-schilling (approved):

PR #414, opened 2026-03-30
Add STM32 BSP peripheral drivers

nhuvaoanh123:
Adds the STM32 BSP peripheral driver modules on top of the merged MCU foundation (#413):

- `bspClock` — clock tree / PLL configuration for STM32F4 (F413ZH, 96 MHz) and STM32G4 (G474RE, 170 MHz), with HSI fallback if the PLL fails to lock
- `bspUart` — polling UART driver used for the debug console (ST-LINK VCP)
- `bspIo` — register-level GPIO driver (`bios::Gpio`)
- `bspTimer` — system timer built on the DWT cycle counter, implementing the `SystemTimer` BSP API
- `bspInterruptsImpl` — interrupt disable/suspend primitives (FreeRTOS and ThreadX variants)
- `etlImpl` — ETL platform glue (clocks, print)

Each module ships with `module.spec` and a documentation introduction per the module guidelines. Pin assignments and other board decisions are left to the board configuration, which follows in the next PRs of the series together with the executables that consume these drivers.

All sources compile standalone against `main` for both `STM32F413xx` and `STM32G474xx` with arm-none-eabi-gcc (`-Wall -Wextra -Werror`).

PR #415, opened 2026-03-30
Add STM32 CAN drivers and bxCAN transceiver

nhuvaoanh123:
## Purpose of this PR
- [ ] Bugfix
- [x] New Feature
- [ ] Documentation Update
- [ ] Other

## Description

Third PR in the STM32 platform series. Adds CAN peripheral device drivers and the bxCAN transceiver adapter.

### Modules added
| Module | Purpose |
|--------|---------|
| `bspCan` | Low-level CAN device drivers — BxCanDevice (F4) and FdCanDevice (G4) |
| `bxCanTransceiver` | OpenBSW `ICanTransceiver` adapter for bxCAN peripheral |
| `unitTest` | STM32-specific unit test cmake infrastructure |

### Architecture
- `bspCan/CMakeLists.txt` uses `CAN_TYPE` to select the correct driver at compile time
- BxCanDevice: 3 TX mailboxes, 28 filter banks, hardware FIFO with overrun detection
- FdCanDevice: 3 TX FIFO slots, 28 standard + 8 extended filters, FD frame support
- Unit tests use mock register headers (`test/include/mcu/mcu.h`) for host-based verification

### Milestone
`cmake --preset tests-stm32-debug` configures successfully. Unit tests (BxCanDeviceTest, BxCanTransceiverTest) run in CI Docker environment.

Depends on #414 (PR 2: BSP drivers). PR 3 of 10.

## Related Issues
Part of STM32 platform port — see #413 for series overview.

## Breaking Changes
- [ ] Yes
- [x] No

## Test Plan
1. Configure unit tests: `cmake --preset tests-stm32-debug`
2. Build and run: `cmake --build --preset tests-stm32-debug && ctest --preset tests-stm32-debug`
3. Tests: BxCanDeviceTest, BxCanTransceiverTest, BxCanTransceiverIntegrationTest

## Regression Tests
Have tests been added/updated? [x] Yes — BxCanDeviceTest, FdCanDeviceTest, BxCanTransceiverTest [ ] No

### DIFF
--- a/platforms/stm32/bsp/bspCan/src/can/BxCanDevice.cpp
+++ b/platforms/stm32/bsp/bspCan/src/can/BxCanDevice.cpp
@@ -1,0 +1,401 @@
+/********************************************************************************
+ * Copyright (c) 2026 An Dao
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#include <can/BxCanDevice.h>
+
+namespace bios
+{
+
+// Bounded busy-wait for the INAK handshake. A plain cycle count is used
+// instead of bsp::isEqualAfterTimeout so this register-level driver stays
+// free of the bsp timer dependency - init() runs during early board
+// bring-up, before the system timer is guaranteed to be running.
+static constexpr uint32_t INIT_TIMEOUT_CYCLES = 100000U;
+
+BxCanDevice::BxCanDevice(Config const& config)
+: fConfig(config), fRxQueue{}, fRxHead(0U), fRxCount(0U), fInitialized(false)
+{}
+
+void BxCanDevice::enablePeripheralClock()
+{
+    // Enable CAN1 clock on APB1
+    RCC->APB1ENR |= RCC_APB1ENR_CAN1EN;
+    // Small delay for clock stabilization
+    uint32_t volatile dummy = RCC->APB1ENR;
+    (void)dummy;
+}
+
+void BxCanDevice::configureGpio()
+{
+    GPIO_TypeDef* txPort = fConfig.txGpioPort;
+    GPIO_TypeDef* rxPort = fConfig.rxGpioPort;
+
+    // TX pin: AF mode, push-pull, high speed
+    txPort->MODER &= ~(3U << (fConfig.txPin * 2U));
+    txPort->MODER |= (2U << (fConfig.txPin * 2U));
+    txPort->OSPEEDR |= (3U << (fConfig.txPin * 2U));
+    if (fConfig.txPin < 8U)
+    {
+        txPort->AFR[0] &= ~(0xFU << (fConfig.txPin * 4U));
+        txPort->AFR[0] |= (static_cast<uint32_t>(fConfig.txAf) << (fConfig.txPin * 4U));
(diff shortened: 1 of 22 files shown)


rolandreichweinbmw:
Is this PR ready for review and merge?

nhuvaoanh123:
Yes, ready for review, thanks for checking. Rebased onto current main as a single self-contained commit (bspCan + bxCanTransceiver), with a review pass done and unit tests green (656/656).

platforms/stm32/bsp/bspCan/src/can/BxCanDevice.cpp
```diff
@@ -0,0 +1,391 @@
...
+    {
+        rxPort->AFR[1] &= ~(0xFU << ((fConfig.rxPin - 8U) * 4U));
+        rxPort->AFR[1] |= (static_cast<uint32_t>(fConfig.rxAf) << ((fConfig.rxPin - 8U) * 4U));
+    }
+}
+
+bool BxCanDevice::enterInitMode()
+{
+    fConfig.baseAddress->MCR |= CAN_MCR_INRQ;
+#if !defined(UNIT_TEST)
```
rolandreichweinbmw: Please don't behave differently under test. In the concrete case, the interesting part of the function is masked from the unit test so testing it doesn't prove much.

Similar for the other cases below.

platforms/stm32/bsp/bspCan/src/can/FdCanDevice.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+    // Reject non-matching, configure standard ID filter elements
+    fConfig.baseAddress->RXGFC = (2U << FDCAN_RXGFC_ANFS_Pos)   // Reject non-matching std
+                                 | (2U << FDCAN_RXGFC_ANFE_Pos) // Reject non-matching ext
+                                 | (static_cast<uint32_t>(count) << FDCAN_RXGFC_LSS_Pos);
+
+    // Write filter elements to message RAM (standard filter area)
+    uint32_t* filterRam = reinterpret_cast<uint32_t*>(ramBase + STD_FILTER_OFFSET);
+
+    for (uint8_t i = 0U; i < count && i < 28U; i++)
```
rolandreichweinbmw: Can this magic number (size if idList) be replaced? Also, bare pointers and size for idList can easily be replaced with using etl::array and etl::span as (mostly) zero cost abstractions that make the code safer.

platforms/stm32/bsp/bspCan/src/can/BxCanDevice.cpp
nhuvaoanh123: Fixed - the UNIT_TEST conditionals are removed; the INAK wait loops and the RX FIFO drain are now always compiled and covered by the unit tests, including the timeout paths. The ifdefs were a leftover from initial hardware bring-up: I initially treated this as bring-up-only code and did not give the unit tests enough thought.

platforms/stm32/bsp/bspCan/src/can/FdCanDevice.cpp
nhuvaoanh123: Done - the magic number is now a named constant (STD_FILTER_COUNT = 28, FILTER_BANK_COUNT = 14 for bxCAN), and configureFilterList() takes etl::span<uint32_t const> in both drivers; the FDCAN driver stores the filter list as a span member. Rebased onto current main, 736/736 unit tests green.

rolandreichweinbmw (approved):

PR #417, opened 2026-03-30
Add STM32 fault handler and watchdog safety

nhuvaoanh123:
## Purpose of this PR
- [ ] Bugfix
- [x] New Feature
- [ ] Documentation Update
- [ ] Other

## Description

Fifth PR in the STM32 platform series. Adds the fault handling and watchdog safety platform modules.

### Modules added
| Module | Purpose |
|--------|---------|
| `hardFaultHandler` | Cortex-M4 HardFault handler — saves a register dump (R0–R12, SP, LR, PC, xPSR, handler LR/PSR, stack snapshot, checksum) to a no-init RAM region that survives reset, for post-mortem analysis |
| `safeBspMcuWatchdog` | STM32 IWDG driver behind the same `safety::bsp::Watchdog` API as the S32K1xx module of the same name; PR/RLR updates use bounded waits on the `PVU`/`RVU` sync flags |

Both mirror the `platforms/s32k1xx` layout (`hardFaultHandler`, `safety/safeBspMcuWatchdog`). The board-level `safeLifecycle` glue (SafetyManager) arrives with the G474RE board bring-up PR, which adds the `nucleo_g474re` platform directory it lives in.

### Milestone
`WatchdogTest` (104 cases) runs in the `tests-stm32-debug/-release` CI jobs. On hardware, the fault handler and watchdog activate with the FreeRTOS board config PR.

PR 5 of the series.

## Related Issues
Part of the STM32 platform port — #394.

## Breaking Changes
- [ ] Yes
- [x] No

## Test Plan
1. `cmake --preset tests-stm32-debug && cmake --build --preset tests-stm32-debug && ctest --preset tests-stm32-debug` — 911 tests green, including the bounded-wait timeout paths against stuck update flags
2. Cross-compile check for both device families (`STM32G474xx`, `STM32F413xx`): driver `-fsyntax-only` and fault handler assembled for both `NO_INIT_RAM` variants
3. Validated on NUCLEO-G474RE hardware in a local integration build: IWDG armed and serviced by the 10 ms safety loop; `safety wdg` starvation resets the board and the reboot detects `RCC_CSR_IWDGRSTF`; `safety mpu` hard fault chains through the dump handler to reset with an intact, checksum-consistent dump in no-init RAM

## Regression Tests
Have tests been added/updated? [x] Yes [ ] No

### DIFF
--- a/platforms/stm32/safety/safeBspMcuWatchdog/src/watchdog/Watchdog.cpp
+++ b/platforms/stm32/safety/safeBspMcuWatchdog/src/watchdog/Watchdog.cpp
@@ -1,0 +1,131 @@
+/********************************************************************************
+ * Copyright (c) 2026 An Dao
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#include <watchdog/Watchdog.h>
+
+#include <mcu/mcu.h>
+
+namespace safety
+{
+namespace bsp
+{
+
+uint32_t Watchdog::watchdogServiceCounter = 0U;
+
+// Bounded busy-wait for the IWDG PR/RLR update handshake. A PR/RLR update
+// synchronizes into the LSI clock domain within 5 LSI cycles (~156 us at
+// 32 kHz); 100k iterations is orders of magnitude above that. A plain cycle
+// count is used instead of bsp::isEqualAfterTimeout so this register-level
+// driver stays free of the bsp timer dependency - enableWatchdog() runs
+// during early board bring-up, before the system timer is guaranteed to be
+// running.
+static constexpr uint32_t SYNC_TIMEOUT_CYCLES = 100000U;
+
+void Watchdog::computePrescalerAndReload(
+    uint32_t timeoutMs, uint32_t clockHz, uint8_t& prescalerCode, uint16_t& reload)
+{
+    // IWDG prescaler dividers: /4, /8, /16, /32, /64, /128, /256
+    // prescalerCode 0..6 -> divider = 4 << code
+    //
+    // timeout_ms = (reload + 1) * divider * 1000 / clockHz
+    // reload = (timeout_ms * clockHz) / (divider * 1000) - 1
+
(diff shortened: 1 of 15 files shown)


nhuvaoanh123:
Rebased onto current main and marked ready for review.

Scope change vs. the draft: the board-level `safeLifecycle` glue moved to the upcoming G474RE board bring-up PR — per-board `safeLifecycle` under the board's platform directory is the existing convention (posix, s32k148evb), and that directory does not exist yet. This PR now adds only the two platform modules, mirroring the `platforms/s32k1xx` layout.

Driver change vs. the draft: the IWDG PR/RLR update waits are now bounded and ordered per the reference manual (a write is ignored while the corresponding update flag is set, so each register waits for its flag before the write), with timeout-path unit tests against stuck flags.

Validated on a NUCLEO-G474RE in a local integration build: watchdog armed and serviced by the 10 ms safety loop; starving the kick resets the board within the 250 ms timeout and the restart detects `RCC_CSR_IWDGRSTF`; a forced hard fault chains through the dump handler to reset with a checksum-consistent register dump in no-init RAM afterwards.

platforms/stm32/safety/safeBspMcuWatchdog/src/watchdog/Watchdog.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+    // IWDG prescaler dividers: /4, /8, /16, /32, /64, /128, /256
+    // prescalerCode 0..6 -> divider = 4 << code
+    //
+    // timeout_ms = (reload + 1) * divider * 1000 / clockHz
+    // reload = (timeout_ms * clockHz) / (divider * 1000) - 1
+
+    static uint16_t const dividers[] = {4, 8, 16, 32, 64, 128, 256};
+
+    for (uint8_t code = 0U; code < 7U; code++)
```
rolandreichweinbmw: Please use size_t for indexes and iteration variables. Typically saves memory and runtime since on RISC like ARM you don't generate conversion operations then. Also good for safety since you push value limits further away.

nhuvaoanh123: Done in 24980d61: the prescaler search index is now size_t, with the loop bound derived from the divider table instead of the magic count. The uint8_t out-parameter keeps an explicit cast since IWDG->PR is a 3-bit field. Also aligned the loop indexes in the unit tests.

rolandreichweinbmw (approved):

PR #444, opened 2026-04-20
User specific configuration of module uds

hans-guenterkothgasserbmw:
Static configuration in module uds that is project specific should be moved to a configuration file that is placed outside the static code in directory bsw. Preferred file is UdsConfig.h.

### DIFF
--- a/executables/referenceApp/udsConfiguration/include/uds/UdsConfig.h
+++ b/executables/referenceApp/udsConfiguration/include/uds/UdsConfig.h
@@ -1,4 +1,5 @@
 // Copyright 2024 Accenture.
+// Copyright 2026 BMW AG
 
 #pragma once
 #include "platform/estdint.h"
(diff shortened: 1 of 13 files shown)


rolandreichweinbmw:
The pytest was successful on second try (manually triggered).

I guess this is a race condition in the test.

But probably not caused by this PR.

rolandreichweinbmw (dismissed):

executables/referenceApp/udsConfiguration/include/uds/UdsConfig.h
```diff
@@ -18,4 +19,36 @@ class UdsVmsConstants
     static uint32_t const TESTER_PRESENT_TIMEOUT_MS    = 5000U;
     static uint8_t const BUSY_MESSAGE_EXTRA_BYTES      = 7U;
 };
+
+class UdsConstants
```
christian-schilling: A `class` with only public members should be a `struct`.
Why a class/struct at all and not a namespace?

executables/referenceApp/udsConfiguration/include/uds/UdsConfig.h
```diff
@@ -18,4 +19,36 @@ class UdsVmsConstants
...
+public:
+    /**
+     * This is an extended timeout which is used when the programming
+     * session in bootloader is entered. This is due to the fact that
+     * a tester that is connected via ethernet may need more time
+     * to reconnect (e.g. slow DHCP server)
+     *
+     * 60s no longer needed, changed to required 10s
+     */
+    static uint32_t const TESTER_PRESENT_EXTENDED_TIMEOUT = 10000U; // ms
```
christian-schilling: `constexpr` is preferable over `static const`

rolandreichweinbmw (approved):

christian-schilling (approved):

PR #452, opened 2026-05-05
Provide NetworkInterface and NetworkInterfaceCollection classes

christophruethingbmw:
Today we just have a loose collection of functions to handle LwIP network interfaces. This does not feel like a nice interface to use and when using it the first time I had to do quite some research on which LwIP functions are already called through abstractions and which ones I still have to call directly.

This change tries to introduce an easy to use class based abstraction of a LwIP interface and also provide a collection of interfaces with some logic to properly route frames to the matching interface based on its VLAN ID.

Note, that this PR just provides the implementation not yet used in the reference application. We should first do a review to agree on it in general and can then refactor existing implementations to use the new one. For example, the current reference application already has a similar idea of a `Netifs` struct but less object oriented and not generically usable.

Also the driver interface is something we might need to discuss on in another follow-up on how BSP interfaces are done in OpenBSW.

An example on how to use this can look like shown below. The ethernet system just needs to hold a single member `_interfaces` and call the proper methods, I think from perspectiv of somebody integrating this in his/her own system, this feels convenient and straight forward to use?

```cpp
void EthernetSystem::init()
{
   // add all interfaces you want
    _interfaces.add(69U, ::ip::make_ip4(192U, 168U, 249U, 16U), ::ip::make_ip4(255U, 255U, 255U, 128U));
    _interfaces.add(73U, ::ip::make_ip4(192U, 168U, 199U, 16U), ::ip::make_ip4(255U, 255U, 255U, 128U));
    ...
}

void EthernetSystem::run()
{
    // set the interfaces up once the system is started
    _interfaces.up();
    ...
}

void EthernetSystem::shutdown()
{
    // set the interfaces down once the system is stopped
    _interfaces.down();
    ...
}

void EthernetSystem::execute()
{
    // process incoming frames, this will select the correct interface based on VLAN internally
    // => actually we could even have a _interfaces.service() which does this
    _driver.service(delegate::create<InterfaceCollectionType, &InterfaceCollectionType::input>(_interfaces));
    ...
}
```

### DIFF
--- a/libs/bsw/lwipSocket/CMakeLists.txt
+++ b/libs/bsw/lwipSocket/CMakeLists.txt
@@ -1,5 +1,7 @@
 add_library(
     lwipSocket
+    src/lwipSocket/netif/NetworkInterface.cpp
+    src/lwipSocket/netif/NetworkInterfaceCollection.cpp
     src/lwipSocket/netif/LwipNetworkInterface.cpp
     src/lwipSocket/tcp/LwipServerSocket.cpp
     src/lwipSocket/tcp/LwipSocket.cpp
--- a/libs/bsw/lwipSocket/include/lwipSocket/netif/NetworkInterface.h
+++ b/libs/bsw/lwipSocket/include/lwipSocket/netif/NetworkInterface.h
@@ -1,0 +1,184 @@
+#pragma once
+
+#include "ip/IPAddress.h"
+
+#include "lwip/err.h"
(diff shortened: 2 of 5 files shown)


christophruethingbmw:
Converted to draft, there will be a more extensive discussion on how to provide a proper abstraction which also supports all the features and corner cases we need and already know about. Results will then be documented here or in a separate issue once the discussion progresses.

PR #458, opened 2026-05-19
Assign symbols to specific memory regions in ld script

vladyslavmarkovaccenture:
Assign symbols to specific memory regions in ld script


executables/referenceApp/platforms/s32k148evb/main/linkerscript/application.dld
```diff
@@ -229,8 +229,28 @@ SECTIONS
     . = ALIGN(4);
     __DATA_RAM = .;
     __data_start__ = .;
-    *(.data)
-    *(.data*)
+    /* Communication-related components */
+    *app.o(.data .data*)
```
rolandreichweinbmw: The same file list is now listed 7 times in this linker script, in different variants - hard to maintain.

Can this be factored out? E.g. via include construction, or macros? (Preprocessor would need to be employed explicitly, though.)

rolandreichweinbmw:
@vladyslavmarkovaccenture Can you please explain the background behind this change? Which problem it solves? Ideally, in the commit message and or documentation?

executables/referenceApp/platforms/s32k148evb/main/linkerscript/application.dld
vladyslavmarkovaccenture: Unfortunately we need to reference units 7 times. I created a single source of truth file for hot memory. Macros are defined within linkerscript itself.

vladyslavmarkovaccenture:
@rolandreichweinbmw I will add another commit to this PR with changes to documentation. TL;DR functionally it's an expansion of usable memory. All new stuff will be put to SRAM_U by default, "hot" functionality is moved to SRAM_L. Previously SRAM_U was not used (except for 0x400 bytes allocated for NoInit).

PR #459, opened 2026-05-19
Add Bazel CI workflow and buildifier formatting

DominikAFischer:
:information_source: **NOTE:**  The bulk of the changes made by this PR is the automatically generated Bazel lockfile: +2615 / -1

Add Bazel CI workflow and buildifier formatting:
- Add bazel github action workflow, which currently contains these checks:
   - Smoke test: `bazel version`
   - Stale lockfile check: `bazel mod deps --lockfile_mode=error`
   - Bazel format check: `bazel run //:format_check`
   - Query all bazel targets: `bazel query //...`
   - Build all targets for host platform: `bazel build //...`
   - Build all targets for example embedded platform: `bazel build --config=s32k148 //...`   
   - Run all bazel tests: `bazel test //...`
- Refactored all workflows to use a shared setup-docker-env composite action for Docker image build and Bazel cache directory setup.
- Add buildifier_prebuilt 8.5.1.1 dev dependency to MODULE.bazel
- Add //:format_check and //:format_fix targets via buildifier_prebuilt
- Add buildifier to Docker image and add it as a formatter to treefmt.
- Apply Bazel file formatting fixes

-----
**NOTE: Docker bind mount permission issues**:
Bazelisk/Bazel cache directories created by Docker bind mounts via `create_host_path: true` are owned by root, causing permission errors for the unprivileged container user. It seems this is a fundamental, unresolved Docker Engine limitation (https://github.com/moby/moby/issues/2259 has been open since 2013).
Instead `create_host_path: false` is used and it's the users responsibility to set up the cache dirs locally. This will be added to the README and Bazel quick start guide once the Bazel related sections are added.

-----
**NOTE:** We have decided to keep two instances of buildifier for now:
- Docker instance for treefmt support
- Bazel module instance for native Bazel formatting/linting


libs/bsw/util/BUILD.bazel
```diff
@@ -1,47 +1,49 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
 cc_library(
     name = "util",
     srcs = [
```
christophruethingbmw: I think we discussed in our last PR that in case of such a long list we wanted to stay with `glob` although it can have some drawbacks? I would at least say we should keep things consistent everywhere?

.github/workflows/bazel.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: bazel query //...
+        run: |
+          DOCKER_UID=$(id -u) DOCKER_GID=$(id -g) DOCKER_HISTORY=/dev/null docker compose run --rm development \
+            bazel query //...
+
+      # Compiles all targets for the host platform (Linux/posix).
+      # Targets with incompatible platform constraints are skipped automatically.
+      - name: bazel build //...
+        run: |
+          DOCKER_UID=$(id -u) DOCKER_GID=$(id -g) DOCKER_HISTORY=/dev/null docker compose run --rm development \
```
christophruethingbmw: Maybe it is nice to use `run: >-` to break lines in the `yaml` but get a single line as resulting string? I think that's nicer to read and write than the `\`.

.github/workflows/bazel.yml
```diff
@@ -0,0 +1,84 @@
...
+            bazel version
+
+      # Fails if MODULE.bazel.lock is stale (invalid versions, yanked modules, extension errors).
+      # Fix locally with: bazel mod tidy --lockfile_mode=update
+      - name: Stale lockfile check
+        run: |
+          DOCKER_UID=$(id -u) DOCKER_GID=$(id -g) DOCKER_HISTORY=/dev/null docker compose run --rm development \
+            bazel mod deps --lockfile_mode=error || exit_code=$?
+          if [ "${exit_code:-0}" -ne 0 ]; then
```
christophruethingbmw: Is the purpose of this `if` onyl to provide this error message?

docker-compose.yaml
```diff
@@ -5,29 +5,27 @@ services:
...
-        no_proxy: ${no_proxy}
+        http_proxy: ${http_proxy:-}
+        https_proxy: ${https_proxy:-}
+        no_proxy: ${no_proxy:-}
       cache_from:
         - type=gha
       cache_to:
         - type=gha
-    user: build
+    # Must run as root so entrypoint can fix bind-mount ownership.
```
christophruethingbmw: I do not fully get the purpose of this. Is it because in case a bind mount does not exist, docker will automatically create a new directory owned by root? I think that's just the behavior and I would not try to fix this within the docker entrypoint script. In my opinion, whoever runs the container need to make sure to have the proper directories already created.

To make this explicit as a requirement, I typicalls specify the `create_host_path: false` such that docker will not create missing directories but throw a clear error.

For CI execution we should then simply have a pre-step that creates the directory outside of the container.

DominikAFischer: I tried to make the developer experience frictionless: Ideally, cloning the repo and running docker compose should just work without requiring any prior host-side setup. That said I'm really disappointed, that there's no simpler solution to such a common problem and I agree that it feels a bit hacky. So if we drop it, what would be best?

- We push the responsibility to the user and add an entry to the README / quick start guide (what you suggested above).
- Instead, we could consider using mounted volumes. Then things would be very simple and work out of the box again, but the cache would no longer be shared between the host and the container. I believe there's currently no real requirement for a host/container shared cache.

Either way is fine for me

.github/workflows/bazel.yml
DominikAFischer: The errors that Bazel produce can often be somewhat cryptic and it's hard to decipher the root cause. I often try to add some more context to CI errors so that users can identify the issue from a quick look at the logs. This will purely add another line to the error output. If you prefer I can remove these

libs/bsw/util/BUILD.bazel
DominikAFischer: done

.github/workflows/bazel.yml
DominikAFischer: done

docker-compose.yaml
DominikAFischer: As discussed, we chose this direction:

> We push the responsibility to the user and add an entry to the README / quick start guide once the Bazel relevant sections are added.

.github/workflows/bazel.yml
christophruethingbmw: @DominikAFischer , I see your point, but I think it makes the command less readable, I typically like to keep the run commands as short as possible, such in case you later look into them you can easily see what's really happening and do not have to think about error handling and such.

However, I agree that it could be helpful to provide more information to a developer. Ideally, Github would provide a seprate functionality to handle such error messages, but unfortunately it does not have. Always a suprise, that such common features do not exist.

I would propose three options (all fine for me):

- We leave it out and keep it clean.
- We follow your approach but keep it consistent in all run commands, i.e. always use the `if` in a separate line and do the `echo "::error::...`. At the moment we do this differently for the last run.
- We could also check whether the problem matcher using `echo "::add-matcher` could be helpful. I think it only parses the output of a build step (and cannot access exit codes directly) but maybe we could use this as a more generic solution?

docker-compose.yaml
christophruethingbmw: Perfect, sounds good.

.github/workflows/bazel.yml
DominikAFischer: I'll remove the additional messages for now

DominikAFischer:
False positive / flaky test `uds/test_udsToolRDBI.py::test_rdbi[posix]` in workflow `Run pytest for posix platform / run-pytest-on-posix` identified.
[Increasing this timeout](https://github.com/eclipse-openbsw/openbsw/blob/main/test/pyTest/helper/helper.py#L10) fixes the issue and confirms the root cause. For a proper fix this timeout should be replaced by properly handling it using either polling or something like `pexpect`.
**EDIT**: The issue has been raised to the maintainer and a proper fix is underway.

christophruethingbmw (dismissed):
Looks good to me, just wait for the fix of the flaky test, PR for that should come soon.

christophruethingbmw (approved):

PR #465, opened 2026-05-27
Add rustfmt to treefmt config

felixm-acc:
Add rustfmt to treefmt config

This adds a wrapper script for rustfmt so treefmt skips rustfmt silently
when the rustfmt executable is not found (similarly to passing
`--allow-missing-formatter` to treefmt).

### DIFF
--- a/.treefmt.toml
+++ b/.treefmt.toml
@@ -43,3 +43,7 @@ includes = [ "*.c", "*.cc", "*.cpp", "*.h", "*.hh", "*.hpp" ]
 command = "cmake-format"
 options = [ "-i" ]
 includes = [ "**/CMakeLists.txt", "*.cmake", "CMakeLists.txt" ]
+
+[formatter.rust]
+command = "tools/rustfmt-wrapper"
+includes = [ "*.rs" ]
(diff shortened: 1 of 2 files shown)


rolandreichweinbmw (approved):

.treefmt.toml
```diff
@@ -43,3 +43,7 @@ includes = [ "*.c", "*.cc", "*.cpp", "*.h", "*.hh", "*.hpp" ]
 command = "cmake-format"
 options = [ "-i" ]
 includes = [ "**/CMakeLists.txt", "*.cmake", "CMakeLists.txt" ]
+
+[formatter.rust]
+command = "tools/rustfmt-wrapper"
```
rolandreichweinbmw, line 48: Can this be written directly into .treefmt.toml without a separate script?

`treefmt --allow-missing-formatter` might be an option also.

felixm-acc: I decided against `allow-missing-formatter` because that's a global switch that applies to all formatters, even the existing ones. Since some users may not use the Rust parts, it may make sense to just make that one optional for now. 

In https://treefmt.com/v2.1/getting-started/configure/#config-file there is also no way described to make it optional in the .toml file, hence the wrapper script.

PR #467, opened 2026-05-29
Update run_process() implementation for udstool tests

nemadhu:
Update run_process() implementation for udstool tests

Adding run_process() helper using python subprocess
The helper function used to run udstool for test/pyTest/uds/test_udsToolRDBI.py took longer than the one second wait as per the previous implementation and the test was failing. Replacing the implementation of run_process() so as to fix this issue.


test/pyTest/uds/test_udsToolRDBI.py
```diff
@@ -14,13 +13,11 @@ def test_rdbi(target_session, did="CF01"):
     command = create_uds_tool_command(
         target_session.target_info.socketcan["channel"], did
     )
-    print(command)
-    output = helper.run_process(command)
+    output = run_process(command, "PositiveResponse")
```
christophruethingbmw: Should we specify the timeout explicitly here? Otherwise we use the default (of I think 5s) and we are actually not aware what timeout we will have here?

christophruethingbmw (dismissed):
Looks good to me

test/pyTest/uds/helpers/helper_functions.py
```diff
@@ -8,7 +13,65 @@ def encode(self, value: bytes) -> bytes:
...
+    formatted_response = " ".join(
+        response[i : i + 2] for i in range(0, len(response), 2)
+    )
+    return formatted_response
+
+
+def run_process(cmd, output_contains, timeout=5):
+    """Run a shell command and wait until specific output appears.
+
+    The command is executed in ``/bin/sh`` using ``subprocess``. The function waits
```
DominikAFischer: `/bin/sh` vs `/bin/bash `below

test/pyTest/uds/helpers/helper_functions.py
```diff
@@ -8,7 +13,65 @@ def encode(self, value: bytes) -> bytes:
...
+
+    Returns:
+        Captured process output containing `` output_contains``.
+
+    Raises:
+        pytest.fail: If the process times out or exits before producing
+            `` output_contains``.
+    """
+    proc = subprocess.Popen(
+        ["/bin/bash", "-lc", cmd],
```
DominikAFischer: I would use a non-login shell here to isolate: `["/bin/bash", "-c", cmd],`

test/pyTest/uds/helpers/helper_functions.py
```diff
@@ -8,7 +13,65 @@ def encode(self, value: bytes) -> bytes:
...
+        stderr=subprocess.STDOUT,
+        text=True,
+        bufsize=0,
+    )
+
+    poller = select.poll()
+    poller.register(proc.stdout, select.POLLIN)
+
+    output = []
+    deadline = time.time() + timeout
```
DominikAFischer: Nit: you could use `deadline = time.monotonic() + timeout` here, then 

```python
    try:
        while True:
            remaining = deadline - time.monotonic()
            if remaining <= 0:
               # raise timeout errro
```

test/pyTest/uds/helpers/helper_functions.py
```diff
@@ -8,7 +13,65 @@ def encode(self, value: bytes) -> bytes:
...
+
+    try:
+        while time.time() < deadline:
+            remaining_ms = max(0, int((deadline - time.time()) * 1000))
+            events = poller.poll(remaining_ms)
+            if not events:
+                break
```
DominikAFischer: This needs to be a continue, no? Otherwise we would quit in the first iteration if we haven't received any events yet

test/pyTest/uds/helpers/helper_functions.py
```diff
@@ -8,7 +13,65 @@ def encode(self, value: bytes) -> bytes:
...
+            chunk = os.read(proc.stdout.fileno(), 4096).decode()
+            if not chunk:
```
DominikAFischer, line 67: Do we need to check if the process [has finished](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.poll)? 
```python
if not chunk:
   if proc.poll() is not None:
      break
```

test/pyTest/uds/helpers/helper_functions.py
```diff
@@ -8,7 +13,65 @@ def encode(self, value: bytes) -> bytes:
...
+            output.append(chunk)
+            combined = "".join(output)
+
+            if output_contains in combined:
+                return combined
+
+        raise AssertionError(f"[TIMEOUT] Did not find {output_contains} within {timeout}s\n Output:\n{''.join(output)}")
+    finally:
+        proc.kill()
+        proc.wait()
```
DominikAFischer: missing empty line at EOF

DominikAFischer (changes requested):
SOme remarks

test/pyTest/requirements.txt
```diff
@@ -11,3 +11,4 @@ tomli==2.0.1
 typing_extensions==4.12.2
 udsoncan==1.23.1
 wrapt==1.16.0
+pexpect==4.9.0
```
DominikAFischer: Revert the addition of `pexpect`

docker/development/files/requirements.lock
```diff
@@ -115,6 +119,8 @@ poetry==2.2.1
     # via -r docker/development/files/requirements.txt
 poetry-core==2.2.1
     # via poetry
+ptyprocess==0.7.0
+    # via pexpect
```
DominikAFischer: restore the old lockfile to revert the addition of pexpect

test/pyTest/uds/helpers/helper_functions.py
nemadhu: proc.poll() check is added after poller.poll()

test/pyTest/uds/helpers/helper_functions.py
```diff
@@ -8,7 +13,68 @@ def encode(self, value: bytes) -> bytes:
...
+            chunk = os.read(proc.stdout.fileno(), 4096).decode()
+            if not chunk:
+                break
+            output.append(chunk)
+            combined = "".join(output)
+
+            if output_contains in combined:
+                return combined
+
+        raise AssertionError(f"[TIMEOUT] Did not find {output_contains} within {timeout}s\n Output:\n{''.join(output)}")
```
DominikAFischer: This error also gets raised if the subprocess finishes within the timeout. I think we split it up:
- time ran out -> timeout error
- process finished without expected output -> expected output not found error
( you can set a timed_out variable in the if remaining_ms <= 0 block)

test/pyTest/uds/helpers/helper_functions.py
```diff
@@ -8,7 +13,68 @@ def encode(self, value: bytes) -> bytes:
...
+
+    Raises:
+        pytest.fail: If the process times out or exits before producing
+            `` output_contains``.
+    """
+    proc = subprocess.Popen(
+        ["/bin/bash", "-c", cmd],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        text=True,
```
DominikAFischer: `text=True` switches subprocess stdout to a text stream, but further down you're using `os.read()` and decode which normally operates on raw bytes.I think in a polling context this might actually be problematic.? If there's no specific reason switch this to `text=False`

test/pyTest/uds/helpers/helper_functions.py
```diff
@@ -8,7 +13,68 @@ def encode(self, value: bytes) -> bytes:
...
+            if not chunk:
+                break
+            output.append(chunk)
+            combined = "".join(output)
+
+            if output_contains in combined:
+                return combined
+
+        raise AssertionError(f"[TIMEOUT] Did not find {output_contains} within {timeout}s\n Output:\n{''.join(output)}")
+    finally:
```
DominikAFischer, line 76: Probably we also want the `if proc.poll() is None:` condition here. Otherwise it can happen that we try to kill the process that has exited gracefully already

test/pyTest/uds/helpers/helper_functions.py
```diff
@@ -8,7 +13,68 @@ def encode(self, value: bytes) -> bytes:
...
+    Args:
+        cmd: Shell command to execute.
+        output_contains: Output text or pattern to wait for.
+        timeout: Maximum number of seconds to wait for `` output_contains``.
+
+    Returns:
+        Captured process output containing `` output_contains``.
+
+    Raises:
+        pytest.fail: If the process times out or exits before producing
```
DominikAFischer: Nit:; Actually raises `AssertionError`

DominikAFischer (approved):
Already giving you my approval, but I have added a few suggestions I would propose to improve. Other than that the PR looks good 👍

christophruethingbmw (approved):

PR #468, opened 2026-06-01
Replace middleware time interface with etl chrono

lcastela:


rolandreichweinbmw:
We need to discuss the conversion back and forth between ETL lowlevel adapter, and converting back to 32 bit counter in MW.

platforms/s32k1xx/etlImpl/src/clocks.cpp
```diff
@@ -7,16 +7,16 @@ extern "C"
 {
 etl::chrono::high_resolution_clock::rep etl_get_high_resolution_clock()
 {
-    return etl::chrono::high_resolution_clock::rep{static_cast<int64_t>(getSystemTimeNs())};
+    return etl::chrono::high_resolution_clock::rep{static_cast<int64_t>(getSystemTimeUs32Bit())};
```
rolandreichweinbmw, line 10: You are putting a 32 bit value into a 64 bit interface. For the proposed microseconds internal counter, this wraps around after ~72 minutes. Maybe much earlier than expected.

rolandreichweinbmw (changes requested):
Instead of implementing etl_get_high_resolution_clock() etc, I propose acquiring a 32 bit or 64 bit timestamp value explicitly and using etl::chrono just for calculations of durations and timeout comparisons.

In the general case, 64 bit low level timestamp access can be too expensive, while 32 bit can be too small, depending on the use case. Therefore, better make it explicit?

rolandreichweinbmw:
Agreed with @lcastela to close this PR for now.

PR #469, opened 2026-06-02
Add tester address translation to referenceApp TP-Router

felixm-acc:
Add tester address translation to referenceApp TP-Router

Different addressing schemes are used by different buses: E. g. CAN and
our UDS SELFDIAG bus use 1-byte addresses while Ethernet with DoIP uses
2-byte addresses. This makes a difference for tester addresses, which
usually follow the 0x0E__ pattern in DoIP 2-byte addressing and a 0xF_
in 1-byte addressing.

The following changes are made:

Introduce 2-byte <-> 1-byte tester address translation for the
referenceApp transport configuration and use it in the simple transport
router at bus boundaries.

Rename the logical address converter API from `DoIP`/`8-bit` terminology
to `2-byte`/`1-byte` terminology to match the actual addressing model.

Also switch referenceApp transportConfiguration to a compiled library so
the logical address mapping table and converter specialization can live
in a source file.

Align the unitTest transport configuration with the referenceApp by
adding the same address translation functions (is2ByteTesterAddress,
is1ByteTesterAddress, convert2ByteAddressTo1Byte,
convert1ByteAddressTo2Byte) and switching tester addresses to the 0x0E__
/ 0x00F_ pattern. The UT config retains a separate
TESTER_ADDRESS_RANGE_CAN as a second address list for illustrative
purposes.

Refactor the UT mock BusId from common::mock::busid to the ::busid
namespace to match the referenceApp pattern, enabling the
transportRouterSimple library to build in UT mode.

Add unit tests for transportRouterSimple covering ETH <-> SELFDIAG
address translation round-trips across bus boundaries.


felixm-acc:
recheck

libs/bsw/transport/include/transport/LogicalAddress.h
```diff
@@ -12,58 +12,58 @@ namespace transport
...
+    static uint16_t convert2ByteAddressTo1Byte(uint16_t const address)
     {
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wpragmas"
 #pragma GCC diagnostic ignored "-Wundefined-var-template"
         for (auto& l : TESTER_ADDRESS_LISTS)
 #pragma GCC diagnostic pop
         {
-            auto ret = addressfinder::findDoipAddressInSlice(address, l);
+            auto ret = addressfinder::find2ByteAddressInSlice(address, l);
```
matthiaskessler: We are now using etl::span. Better remove the data structure from the name of the function.

felixm-acc: Good point. I changed the function signature to `findBy2ByteAddress(l, address)`. Added `By` to clarify that the 2ByteAddress is what is searched by (and not what is being returned). Also swapped the parameters to mirror `std::find` which has the order start+end range, then search value.

matthiaskessler (approved):

PR #475, opened 2026-06-09
Add BUILD.bazel for libs/bsw targets

SuhashiniNaik:
Add BUILD.bazel for libs/bsw targets

Add Bazel cc_library targets for the following libs/bsw libraries:
- //libs/bsw/timer:timer
- //libs/bsw/asyncImpl:asyncImpl
- //libs/bsw/logger:logger
- //libs/bsw/io:io
- //libs/bsw/middleware
- //libs/bsw/stdioConsoleInput:stdio_console_input


the-nick-fischer (changes requested):
Looks good to me. The issue with the ETL asserts was fixed in the meantime and we should be able to remove the `copts` workaround.

libs/bsw/middleware/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+
+cc_library(
+    name = "middleware",
+    srcs = glob(["src/**/*.cpp"]),
+    hdrs = glob(["include/**/*.h"]),
+    copts = [
+        # TODO: Disable ETL assertion checking to work around pre-existing API misuse:
+        # https://github.com/eclipse-openbsw/openbsw/issues/470
+        "-DETL_NO_CHECKS",
+    ],
```
SuhashiniNaik: Done

libs/bsw/middleware/BUILD.bazel
```diff
@@ -0,0 +1,57 @@
...
+    name = "middleware_lock_types_test_stub",
+    hdrs = ["test/include/middleware/concurrency/lock_types.h"],
+    strip_include_prefix = "test/include",
+)
+
+alias(
+    name = "middleware_lock_types",
+    actual = select(
+        {
+            # TODO: Use test lock_types.h stub since it is the only existing implementation.
```
DominikAFischer: Discussion:
We do not have a referenceApp version of lock_types.h available. On the CMake side I believe also the test stub from here is used: test/include/middleware/concurrency/lock_types.h in the referenceApp build.
That does not seem like a clean solution to me. Probably at least the stub should be copied to some location outside of the test directory.

libs/3rdparty/etl/BUILD.bazel
```diff
@@ -18,7 +30,7 @@ label_flag(
 cc_library(
     name = "etl",
     hdrs = glob(["include/**/*.h"]),
-    strip_include_prefix = "include",
+    includes = ["include"],
```
christophruethingbmw, line 33: Is there a reason we use `includes` here instead of `strip_include_prefix?`

libs/bsw/middleware/BUILD.bazel
```diff
@@ -0,0 +1,57 @@
...
+
+cc_library(
+    name = "middleware_lock_types_test_stub",
+    hdrs = ["test/include/middleware/concurrency/lock_types.h"],
+    strip_include_prefix = "test/include",
+)
+
+alias(
+    name = "middleware_lock_types",
+    actual = select(
```
christophruethingbmw, line 21: About these selects: I think we can merge it like this for now, but I think for the future we have to switch to `label_flags` or a different kind of approach? Because a "user" of OpenBSW want's to use this library but does not have the `executable_config` for his project. The config we have here is actually very OpenBSW specific. Or how would you imagine, that a control unit can then integrate this?

libs/3rdparty/etl/BUILD.bazel
DominikAFischer: On the Cmake side we default to a SYSTEM include specifier (`-isystem`).
```cmake
option(NO_SYSTEM_INCLUDE "Do not include with -isystem" OFF)
if (NO_SYSTEM_INCLUDE)
    set(INCLUDE_SPECIFIER "")
else()
    set(INCLUDE_SPECIFIER "SYSTEM")
endif()
```
 To replicate this in Bazel we can use the `includes` field. I'm actually adding a comment in one of my upcoming PRs:
 `# Use "includes" field here to match CMake SYSTEM include specifier.`

libs/bsw/middleware/BUILD.bazel
DominikAFischer: You're right. I think we should use the same mechanism as we did somewhere else already:
```starlark
cc_library(
    name = "middleware_lock_types_test_stub",
    hdrs = ["test/include/middleware/concurrency/lock_types.h"],
    strip_include_prefix = "test/include",
)

alias(
    name = "middleware_lock_types_default",
    actual = select(
        {
            "//bazel/config/executable_config:unit_test": ":middleware_lock_types_test_stub",
            "//bazel/config/executable_config:reference_app": ":middleware_lock_types_test_stub",
        },
        no_match_error = "Set executable_config to 'reference_app' or 'unit_test', or override --//libs/bsw/middleware:middleware_lock_types",
    ),
)

#   Default:  per executable_config selection above (currently the host stub)
#   Override: --//libs/bsw/middleware:middleware_lock_types=//<your>:lock_types
label_flag(
    name = "middleware_lock_types",
    build_setting_default = ":middleware_lock_types_default",
)
```

It's sometimes hard to decide which config parameters should be overrideable using a label flag. I need to invest some time to understand the Cmake side better so we can find some criteria to decide.

libs/3rdparty/etl/BUILD.bazel
christophruethingbmw: ah, that makes sense, looks good to me

libs/bsw/middleware/BUILD.bazel
christophruethingbmw: Actually I am also not really aware how it is done in cmake because i never use that for integration, we currently also have a custom bazel setup for OpenBSW and thus this works differently.

SuhashiniNaik: Thanks for the context, I'm going to leave it as-is for now and track the label_flag refactor as a follow-up.

christophruethingbmw (approved):

PR #478, opened 2026-06-11
Add Bazel BUILD files for common, bsp, and configuration targets

SuhashiniNaik:
Add Bazel BUILD files for common, bsp, and configuration targets

Migrate the following CMake targets to Bazel:
- libs/bsw/common:common
- libs/bsw/bsp: bsp
- executables/referenceApp/configuration:configuration
- executables/unitTest/configuration:configuration

- Update bazel_migration/README.md

NOTE: 
The public headers exported by //libs/bsw/common:common do not include any headers from the configuration module, so there is currently no direct header dependency requiring it.

In addition, configuration already depends on common . Adding configuration as a dependency of common would therefore introduce a dependency cycle:
common -> configuration -> common
For that reason I kept common dependent only on the libraries required by its exported headers (etl and platform) and did not add a dependency on configuration.


christophruethingbmw (commented):
I wonder why `referenceApp/configuration` looks structurally different compared to `unitTest/configuration`? 

For `referenceApp` we have a header only library called `configuration` and a implementaion library called `common_impl`. For `unitTest` we have quite similar files but the `common_impl` now includes also the headers? Shouldn't we harmonize this more?

SuhashiniNaik:
> I wonder why `referenceApp/configuration` looks structurally different compared to `unitTest/configuration`?
> 
> For `referenceApp` we have a header only library called `configuration` and a implementaion library called `common_impl`. For `unitTest` we have quite similar files but the `common_impl` now includes also the headers? Shouldn't we harmonize this more?

Hi @christophruethingbmw, Thank you for the review. The structural difference in `BUILD.Bazel` is actually inherited from the CMakeLists.txt sources 

In `referenceApp`, the CMake `configuration` target is a INTERFACE library with its own include directory, and `commonImpl` links against it privately , so in Bazel, `common_impl` depends on `:configuration` which owns the headers.

In `unitTest`, the CMake `configuration` is an empty dummy (`add_library(configuration INTERFACE)` with no headers), and `commonImpl` sets `target_include_directories` directly on itself. So the Bazel mirrors that: `common_impl` owns the `hdrs`, and `configuration` is just an empty target.

the-nick-fischer:
As @SuhashiniNaik said, I think she exactly modelled how it is currently implemented on the CMake side. So I'd vote to leave it as is for now. If you think it's required we could open a gh issue for the issue in general?

vladyslavmarkovaccenture:
@christophruethingbmw I will look into configuration/CMakeLists.txt for both UTs and referenceApp. I think the right approach would be to minimize the "noise" and make only one library named **configuration** (since the path is "configuration/CMakeLists.txt"). Thanks for pointing this out.

Update: I managed to cleanup the _configuration/CMakeLists.txt_ for both **unitTest** and **referenceApp**. In new version there will be no **commonImpl** - only **configuration** target.

vladyslavmarkovaccenture:
@christophruethingbmw #488 is created. It is an aforementioned cleanup for both _configuration_ targets.

SuhashiniNaik:
Note:
In the latest change I did not add `configuration` as a dependency of common.
The public headers exported by `//libs/bsw/common:common` do not include any headers from the configuration module, so there is currently no direct header dependency requiring it.

In addition, `configuration` already depends on` common` . Adding `configuration` as a dependency of common would therefore introduce a dependency cycle:
`common -> configuration -> common`
For that reason I kept `common` dependent only on the libraries required by its exported headers `(etl and platform)` and did not add a dependency on` configuration`.

Update:
Used common_headers instead of common to avoid the circular dependency (common → configuration → common) that exists on the CMake side. This way, configuration can depend on common_headers, and common can depend on configuration without a cycle.

executables/unitTest/configuration/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
...
+# *******************************************************************************
+
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+cc_library(
+    name = "configuration",
+    srcs = ["common/src/busid/BusId.cpp"],
+    hdrs = ["common/include/busid/BusId.h"],
+    implementation_deps = [
+        "//libs/bsw/common",
```
DominikAFischer: Replace with: `"//libs/bsw/common:common_headers",`
See comment below.
Nit: This comment would still be valuable: `# TODO: add //libs/bsw/async once its Bazel target is ready`

libs/bsw/common/BUILD.bazel
```diff
@@ -0,0 +1,27 @@
...
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+cc_library(
+    name = "common",
```
DominikAFischer: We still have a circular dependency on the CMake side: `common` -> `configuration` -> `common`
I would advocate to keep `common_headers` as the mechanism to resolve this on the Bazel side (to match the Cmake include graph). Then we also need an alias to select between the configuration variants in the common package:

```python
alias(
    name = "configuration",
    actual = select(
        {
            "//bazel/config/executable_config:reference_app": "//executables/referenceApp/configuration",
            "//bazel/config/executable_config:unit_test": "//executables/unitTest/configuration",
        },
        no_match_error = "Please set executable_config to either 'reference_app' or 'unit_test'",
    ),
)

# Circular dependency present on the CMake side: common -> configuration -> common
# Split off common_headers to break the circular dependency for Bazel
cc_library(
    name = "common_headers",
    hdrs = [
        "include/common/busid/BusId.h",
        "include/util/Mask.h",
        "include/util/concurrent/IFutureSupport.h",
        "include/util/timeout/ITimeoutManager2.h",
    ],
    strip_include_prefix = "include",
    deps = [
        "//libs/3rdparty/etl",
        "//libs/bsw/platform",
    ],
)

cc_library(
    name = "common",
    deps = [
        ":common_headers",
        ":configuration",
    ],
)
```

@vladyslavmarkovaccenture Is there any reason why this circular dependency is required? Could we not just remove the `configuration` dependency from `common`?

DominikAFischer (changes requested):

executables/referenceApp/configuration/BUILD.bazel
```diff
@@ -0,0 +1,29 @@
...
+    hdrs = [
+        "include/app/appConfig.h",
+        "include/busid/BusId.h",
+        "include/systems/ICanSystem.h",
+        "include/systems/IEthernetDriverSystem.h",
+    ],
+    strip_include_prefix = "include",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//libs/bsw/common",
```
DominikAFischer: Would also need to be changed to `//libs/bsw/common:common_headers`

DominikAFischer (approved):
Looks good to me

libs/bsw/common/BUILD.bazel
```diff
@@ -0,0 +1,49 @@
+# *******************************************************************************
+# Copyright (c) 2026 Accenture
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
```
christophruethingbmw, line 11: Nit: Maybe we want to add an empty line here like in the other files?

christophruethingbmw (dismissed):
looks good to me

christophruethingbmw (approved):

libs/bsw/common/BUILD.bazel
vladyslavmarkovaccenture: @DominikAFischer nope. My bad. See #492 .

PR #480, opened 2026-06-12
Integrate Rust -> BSW logger bridge

felixm-acc:
This PR improves Rust support and offline/containerized build reliability:
  
- Rust logging bridge: adds a bridge so Rust code can log through the existing C++ logger.
- Offline builds: vendors first external Rust dependencies so builds work without network access.
- Toolchain: bumps the default Rust version to 1.96.0.
- Container UID handling: makes the cargo and rustup homes accessible to the runtime UID injected into
  the dev container.
- Repo hygiene: adds a .gitattributes file (for exclusion from vendored 3rdparty code from SLOC stats on main page and Github diff view)


doc/dev/learning/setup/setup_posix_build.rst
```diff
@@ -84,11 +84,11 @@ To build OpenBSW with Rust components for the POSIX platform, you need:
 1. The build tools set up as described above (gcc, cmake, make)
 2. The Rust compiler
 
-Install Rust 1.90.0 to be compatible with the CI builds:
+Install Rust 1.96.0 to be compatible with the CI builds:
```
ThoFrankACN: The ubuntu version is defined only once. Probably we should use the same approach for rust:
```
:prop:`tool:ubuntu_version`
```

libs/bsw/logger/doc/index.rst
```diff
@@ -7,4 +7,40 @@ Overview
...
+an index at Rust static-init time would capture ``COMPONENT_NONE`` and never
+see the real value. Fetching live removes any init-ordering constraint.
+
+``log`` facade fallback
+    Plain ``log::info!`` calls (e.g. from third-party crates) are routed to a
+    single default component configured from C++ via ``set_default_component``;
+    if it is unset the message is dropped. The crate name is prepended to the
+    message so the originating module is still visible.
+
+.. warning::
```
ThoFrankACN, line 52: Should we also warn about the performance implications of double-buffering here?

And not sure how cpp behaves, but the formatting overhead for rust messages is still there even if the cpp logger component is disabled.

libs/bsw/logger/rust/src/log_buf.rs
```diff
@@ -0,0 +1,94 @@
...
+            }
+            Ok(())
+        } else {
+            // not enough space
+            // only copy what fits
+            let s = &s[0..(LOG_BUF_SIZE - self.bytes_used - 1)];
+            for &b in s {
+                self.buffer[self.bytes_used] = b;
+                self.bytes_used += 1;
+            }
```
ThoFrankACN: Maybe the loops should be replaced by [copy_from_slice](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice).

libs/bsw/logger/rust/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+pub mod component;
+pub mod ffi;
+pub mod log_buf;
+mod macros;
+
+pub use bsw_logger::BswLogger;
+pub use component::LoggerComponent;
+
+#[doc(hidden)]
+pub use paste as __paste;
```
ThoFrankACN, line 50: Why is there the need to re-export paste?

ThoFrankACN (commented):
Really nice work!

I only have some minor feedback. In my opinion the vendoring approach is good. We have everything available for an offline build and can track changes, but using dependencies from crates.io is still simple.

libs/bsw/logger/doc/index.rst
felixm-acc: Good points, I'll add the two caveats.
I'll also check how it's done in C++, maybe some performance gains hidden there by disabling serialization if it's not going to get printed anyway.

felixm-acc: In the bsw logger it's gated through `if (isEnabled(componentIndex, _level))` in `Logger.h`. I can also (double) check `isEnabled` via FFI from Rust before doing the serialization. Maybe with some profiling of some format heavy logs to see whether that optimization is worth it.

libs/bsw/logger/rust/src/lib.rs
felixm-acc: It's used in the expansion of the log macro. If we don't re-export then it's not available at the callsite of the macro and every caller of our logging macros would need to add a dependency to paste in their `Cargo.toml`. Similar to https://stackoverflow.com/questions/65164156/how-to-set-dependencies-for-a-macro-in-rust#:~:text=lib%2Ers%2E-,Library,attribute

libs/bsw/logger/doc/index.rst
felixm-acc: Test log line with some formatting.

  "rx id={:#05x} dlc={} ts={:>8}ms seq={:05} crc={:#06x}"
     -> "rx id=0x1a4 dlc=8 ts=  123456ms seq=00042 crc=0xbeef"

  Results (200k iterations/run, 3 runs)

  |                    | gated        | naive (always format) | delta                           |
| ------------------ | ------------ | --------------------- | ------------------------------- |
| **disabled component** | ~6.5 ns/call | ~160 ns/call          | **saving ~155 ns (~24x)**       |
| **enabled component**  | ~160 ns/call | ~160 ns/call          | overhead ~ 0  |

definitely seems to be worth it to add the additional is_enabled FFI call also to Rust to skip serializing entirely when the logging component is disabled.

doc/dev/learning/setup/setup_posix_build.rst
felixm-acc: done. There's even a nice mechanism already in place in `doc/dev/conf.py` to replace `x.x` in code blocks with the correct version number depending on which marker string is nearby.

libs/bsw/logger/rust/src/log_buf.rs
felixm-acc: done

ThoFrankACN (approved):
LGTM

christian-schilling (approved):

PR #491, opened 2026-06-19
Migration of bsw libs docan, doip, logger_integration

DominikAFischer:
Migration of bsw libs docan, doip, logger_integration

Lib targets:
- //libs/bsw/docan:docan
- //libs/bsw/doip:doip
- //libs/bsw/loggerIntegration:logger_integration
- //libs/bsw/storage:storage
- //libs/bsw/transport:transport
- //platforms/posix/etlImpl:etl_impl
- //platforms/s32k1xx/etlImpl:etl_impl

Config targets:
- //libs/bsw/loggerIntegration:etl_impl (label_flag injection)
- //libs/bsw/loggerIntegration:etl_impl_default (closed select)
- //libs/bsw/storage:storage_rtos_impl (closed select)


libs/bsw/loggerIntegration/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+            "//bazel/platform/constraints/soc:s32k148": "//platforms/s32k1xx/etlImpl:etl_impl",
+            "@platforms//os:linux": "//platforms/posix/etlImpl:etl_impl",
+        },
+        no_match_error = "No default etl_impl for this platform. Override --//libs/bsw/loggerIntegration:etl_impl with your own ETL implementation target.",
+    ),
+)
+
+#   Default:  select based on the target platform via etl_impl_default (S32K148, POSIX)
+#   Override: --//libs/bsw/loggerIntegration:etl_impl=//path/to:your_impl
+label_flag(
```
christophruethingbmw, line 26: I wonder how this scales? Many libraries will use etl, I think it does not make sense that each library specifies it's own label flag? Shouldn't we have a more generic one at a higher level every library can then depend on?

The `etl_profile` comes in already via the dependency of `//libs/3rdparty/etl` to the label flag `//libs/3rdparty/etl:etl_profile`.

In total we have three things to consider, the etl implemenation itself, the `etl_profile.h` and the integration specific implementations like `print.cpp`.

In our project it is currently handled in a way the both `etl_profile.h` and the implementations are handled via a single `cc_library` (without dependencies, so no cycle) configured as a label flag called `etl_profile`. Then there is a central `etl` library which takes the etl sources and has the `etl_profile` label flag as a dependency. All users of etl can then simply refer to this library, the fact that it requires a label flag is hiddne.

christophruethingbmw (commented):
Ignored the commits from #494, one comment regarding the integration of etl.

libs/bsw/loggerIntegration/BUILD.bazel
DominikAFischer: That's definitely a valid concern and would make things easier as far as I can see.

It also would be a departure from the current CMake implementation and might even need changes to source code if I'm correct. So, I would suggest that we add this to the list of design discussions and handle it separately?

christophruethingbmw: Fine for me, we can do it like this, but we should keep track of this to refactor :)

PR #485, opened 2026-06-12
Apply RAII idiom to Middleware Aggregator constructor

lcastela:
Move pool initialization from the separate `initialize()` static method into the constructor body. This ensures pools are always initialized when an `Aggregator` is constructed, eliminating the risk of using an uninitialized instance and the need for a separate init call.

PR #493, opened 2026-06-22
Queues initialization in ctor

danielfaleirosantos:
These changes move the initialization of the queues and `QueueBase`'s elements to their respective ctors.

### DIFF
--- a/libs/bsw/middleware/include/middleware/queue/Queue.h
+++ b/libs/bsw/middleware/include/middleware/queue/Queue.h
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2025 BMW AG
+ * Copyright (c) 2025, 2026 BMW AG
  *
  * This program and the accompanying materials are made available under the
  * terms of the Apache License Version 2.0 which is available at
@@ -43,14 +43,13 @@ class QueueMutex<T, typename etl::enable_if_t<etl::is_integral<T>::value>>
 public:
     using mutex_t = etl::add_volatile_t<T>;
 
-    /** Initializes the mutex to \p initialValue. */
-    void init(mutex_t initialValue = 0U) { _mutex = initialValue; }
+    constexpr explicit QueueMutex(mutex_t initialValue = 0U) : _mutex(initialValue) {}
 
     /** Returns a pointer to the mutex variable. */
     mutex_t* get() { return &_mutex; }
 
 private:
-    mutex_t _mutex{0U};
+    mutex_t _mutex;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


rolandreichweinbmw:
Can you please document for this PR and also #485 document the conditions under which the plain ctor initialization needs to be done, do prevent initialization issues, e.g. in multicore scenarios.

Note that init() intentionally provides explicit control.

danielfaleirosantos:
@rolandreichweinbmw, @christian-schilling, @matthiaskessler please have a look at the PR. I have added a brief comment on the reasoning behind the initialization in the ctor.

rolandreichweinbmw:
This is the approach of constinit via compile-time initialization only.

The concern in today's PR review meeting is that when you initialize in global/shared memory (instead of in etl::typed_storage), you'll need to ensure to prevent race conditions between cores.

E.g. copying from core0 ROM to RAM, and using it, modifying data. Then copying from core1 ROM to RAM in init, to the same shared RAM would be a race condition, overwriting the live data.

Can you elaborate on that, to clarify how to ensure?

@matthiaskessler please also have a look.

lcastela:
@rolandreichweinbmw, @matthiaskessler , I guess we should have added some context before introducing these type of changes.
 
We want to change the way how middleware is integrated into the target platforms in regards to the shm, to avoid the manual integration of specific and different init calls per core. In order to achieve that, on the middleware generated code we will provide a `MemoryLayout` of the middleware shared data, and also an internal interface that middleware uses to access its shm data. This shm datatype will be "wrapped" in an `etl::typed_storage`, making it a mandatory usage for the target platform. Example:
 
middleware shm generated code with information from the deployment files middleware/shm/config.h:
> struct MemoryLayout
> {
>     uint8_t queueToFooMutex{0U};
>     uint8_t queueToBarMutex{0U};
>     uint8_t defaultAllocatorMutex{0U};
>     MiddlewareQueue<QUEUE_TO_FOO_SIZE> queueToFoo{&queueToFooMutex};
>     MiddlewareQueue<QUEUE_TO_BAR_SIZE> queueToBar{&queueToBarMutex};
>     message_allocator defaultAllocator{&defaultAllocatorMutex};
> };
>
> extern ETL_CONSTINIT etl::typed_storage\<middleware::shm::MemoryLayout\> middlewareData;
>
> extern middleware::shm::MemoryLayout& middleware::shm::getMemoryLayout();
 
shm library .cpp:
> ETL_CONSTINIT etl::typed_storage\<middleware::shm::MemoryLayout\> middlewareData SHARED_BSS;
>  
> middleware::shm::MemoryLayout& middleware::shm::getMemoryLayout()
> {
>     return *middlewareData;
> }
 
integration on controller main core main.cpp:
> void run_core() {
>     middlewareData.create();
> }

And with this, the key is that ETL_CONSTINIT ... SHARED_BSS places the storage in the BSS section of shared RAM — zero-filled at link time, with no per-core ROM-to-RAM copy at startup. The typed_storage wrapper is inert raw bytes until explicitly constructed. Only the master core calls .create(), which constructs MemoryLayout in-place exactly once. All other cores access the already-live object through getMemoryLayout() after the existing infrastructure-ready synchronization barrier, so there is no second initialization and no race.
 
The master core must call .create() on shared objects before any other core accesses them. This ordering must be enforced by a platform-provided synchronization mechanism. The middleware pattern itself does not provide this barrier.
 
ps: the middleware generator is still a wip, hence why this idea was not visible at the moment.

christian-schilling:
> @rolandreichweinbmw, @christian-schilling, @matthiaskessler please have a look at the PR. I have added a brief comment on the reasoning behind the initialization in the ctor.

Looks good to me in principle. But not sure I have enough context to fully understand.

matthiaskessler (approved):

PR #495, opened 2026-06-23
Migrate lwip to Bazel

DominikAFischer:
Migrate lwip to Bazel

- libs/3rdparty/lwip: lwip_core cc_library, lwip_configuration label_flag
- executables/referenceApp/lwipConfiguration: lwip_configuration
  implementation with platform-select for lwipSysArch
- executables/referenceApp/application: application_headers stub
- platforms/posix/lwipSysArch, platforms/s32k1xx/lwipSysArch: sys_arch
  port libraries
- libs/bsw/lwipSocket: lwipSocket cc_library
- Add lwip BUILD file to ignored rim files.


christian-schilling (dismissed):

libs/3rdparty/lwip/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+        "src/netif/*.c",
+        "src/netif/ppp/*.c",
+        "src/netif/ppp/polarssl/*.c",
+    ]),
+    hdrs = glob(["src/include/**/*.h"]),
+    strip_include_prefix = "src/include",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":lwip_configuration",
+        "//executables/referenceApp/application:application_headers",
```
christophruethingbmw: Uhh, why do we need a dependency of lwip to our reference application? I don't think this is a dependency we want, because it is actually prohibiting us to use the same `lwip_core` in a different application?

Looking into the integration in our project we do not have such a dependency, what's really needed from the `application_headers` here actually?

christophruethingbmw (commented):
Ignored the commits from https://github.com/eclipse-openbsw/openbsw/pull/494, one comment regarding the integration of lwip.

libs/3rdparty/lwip/BUILD.bazel
DominikAFischer: I also think this is really bad. But this is how it is also handled on the CMake side and that's why we included it here too.
https://github.com/eclipse-openbsw/openbsw/blob/main/libs/3rdparty/CMakeLists.txt#L58-L60
```
target_include_directories(
    lwipcore PUBLIC "${LWIP_DIR}/src/include"
                    "${OPENBSW_APP_DIR}/application/include")
```
This transitively leaks to all consumers.

If this is not desired, we could just safely remove the depedency and  the `application_headers` target. I was not able to find any target consuming any headers via this path.

christophruethingbmw: then let's remove it, in case it is anyway not used I see no reason to have it.

christophruethingbmw (approved):

PR #496, opened 2026-06-23
Migrate transport, transportRouterSimple and uds to Bazel

DominikAFischer:
Migrate transport, transportRouterSimple and uds to Bazel

- libs/bsw/transport: transport cc_library
- libs/bsw/transportRouterSimple: transport_router_simple cc_library
- libs/bsw/uds: uds_headers cycle-break split, uds cc_library
- executables/referenceApp/transportConfiguration: transport_configuration
  cc_library
- executables/referenceApp/udsConfiguration: uds_configuration,
  uds_configuration_impl cc_libraries

### DIFF
--- a/libs/bsw/transportRouterSimple/BUILD.bazel
+++ b/libs/bsw/transportRouterSimple/BUILD.bazel
@@ -1,0 +1,32 @@
+# *******************************************************************************
+# Copyright (c) 2026 Accenture
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+cc_library(
+    name = "transport_router_simple",
+    srcs = [
+        "src/transport/TpRouterLogger.cpp",
+        "src/transport/routing/TransportRouterSimple.cpp",
+    ],
+    hdrs = [
+        "include/transport/TpRouterLogger.h",
+        "include/transport/routing/TransportRouterSimple.h",
+    ],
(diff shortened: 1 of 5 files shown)


christian-schilling (approved):

libs/bsw/transportRouterSimple/BUILD.bazel
```diff
@@ -0,0 +1,32 @@
...
+        "include/transport/TpRouterLogger.h",
+        "include/transport/routing/TransportRouterSimple.h",
+    ],
+    implementation_deps = [
+        "//libs/bsw/common:configuration",
+    ],
+    strip_include_prefix = "include",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//executables/referenceApp/transportConfiguration:transport_configuration",
```
christophruethingbmw, line 29: Should we hard-code this here or also provide a label flag for it? Like this it would only be usable in combination with the reference app? In case this simple route is anyway tightly coupled to the reference app (I remember fromr another review that it uses some bus id enums which are totally application specific), should it be at all inside the generic `libs` folder?

DominikAFischer: The dependency mirrors the CMake side: `transportConfiguration` is a public dependency via:
```cmake
target_link_libraries(
    transportRouterSimple
    PUBLIC transport transportConfiguration
    PRIVATE configuration)
```
(and essentially hard ties it to referenceApp).

I suspect a label_flag would only be a partial fix,, since the source itself hardcodes referenceApp bus IDs if I'm not mistaken.
So maybe this should be a general CMake / design question?

christophruethingbmw: yes, I agree, this then should be a more generl question, let's move this into a separate PR.

christophruethingbmw (approved):

PR #499, opened 2026-06-24
Add routing and blob

danielfaleirosantos:
The transport channels routing and blob modules have been created.

The communication modes supported are CAN, FlexRay and nPDU (aka PduTransport).


executables/referenceApp/configuration/include/blob/configuration.h
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright (c) 2026 BMW AG
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+// This is a generated file. Please do not edit it.
```
rolandreichweinbmw, line 11: Please make sure we have a documented way to (re-) generate the files marked so.

If possible, we don't even check in generated files since they can be generated in the build process.

matthiaskessler: I think, for now we try to not add generators in CMake if possible for referenceApp. But we can discuss to change that in the future.

danielfaleirosantos: The generator itself is in the blob lib, but yes, it is in CMake. The documentation is now available in https://github.com/danielfaleirosantos/openbsw/blob/addBNRtoOpenSource/libs/bsw/blob/doc/generator.rst @rolandreichweinbmw

libs/bsw/blob/doc/configuration.rst
```diff
@@ -0,0 +1,112 @@
...
+     - 4 bytes
+     - Size, in bytes, of the configuration ``data`` (including padding and the trailing CRC).
+   * - ``data``
+     - ``size``
+     - Configuration payload, optionally padded, ending with a 4-byte CRC.
+
+The last four bytes of every configuration ``data`` hold a big-endian CRC computed over the preceding
+payload bytes. The CRC is verified by ``checkCrc()``, which re-computes a 32-bit CRC over the preceding
+payload bytes using ``etl::crc32`` (standard CRC-32). The payload is padded with ``0xFF`` bytes so that
+its length is a multiple of four before the CRC is appended.
```
simon-d-bmw: 🔴 High: This doesn't match the implementation. `Config::from_bytes` (Config.cpp:53-63) builds `Config.data` as the **entire config record** — `type` + `reserved` + `size` + payload — starting at byte 0, not just the payload after the header. Since `checkCrc()` is always called as `checkCrc(config.data)` (Blob.cpp:46), the CRC actually covers `type`/`reserved`/`size` + payload, not "the preceding payload bytes" as stated here.

This is an interop risk: an external tool implementing "CRC over payload only" per this doc would produce a mismatching CRC. Please correct this to state the CRC covers the full configuration entry (header fields + payload), excluding only the trailing 4-byte CRC itself.

libs/bsw/blob/doc/examples.rst
```diff
@@ -0,0 +1,100 @@
...
+
+               0x00, 0x00, 0x00, 0x00, ``offset``
+
+            ``output message ids``
+               0x00, 0x00, 0x00, 0x00, ``size``
+
+            ``destinations``
+               0x00, 0x00, 0x00, 0x00, ``size``
+
+            0xBD, 0xEA, 0x81, 0x89, ``crc``
```
simon-d-bmw, line 93: 🔴 High: This diagram misplaces `size` — it's nested under `data` (e.g. line 69), but the actual layout has `size` as a sibling field *before* `data` (see configuration.rst's table). This will need to be redrawn once the header/data boundary is clarified (see comment on configuration.rst).

Separately: this mixes generic blob-format concepts with routing-specific payload semantics (`destination offsets`, `output message ids`, `destinations`, lines 89-98). Since this is the generic `blob` module's example page, consider moving routing-specific payload interpretation into the routing library's own docs.

libs/bsw/blob/doc/examples.rst
```diff
@@ -0,0 +1,100 @@
...
+
+      return 0;
+    }
+
+In this example, a ``Blob`` instance is created, after which the loop iterates over its
+configurations, namely, ``CHANNEL_NAMES: 0xDD`` and ``ROUTING: 0x00``, and produces the output
+below. Since the ``ROUTING`` configuration contains valid data, re-computed CRC bytes turn out to be
+equal to pre-computed ``CRC bytes: 0xBDEA8189``.
+
+.. code-block::
```
simon-d-bmw: 🟢 Nitpick: Missing language on this code-block directive — should probably be `console` or `text` (it's showing program output).

libs/bsw/blob/doc/generator.rst
```diff
@@ -0,0 +1,93 @@
...
+``header config-type``
+    Emits a C++ header containing the ``ConfigType`` enum used by the module
+    (see :doc:`configuration`).
+
+``header metadata``
+    Emits a C++ header containing the ``Metadata`` enum derived from the ``meta`` entries in the
+    input.
+
+Each command accepts ``-i``/``--input`` (defaulting to stdin) and ``-o``/``--output``
+(defaulting to stdout).
```
simon-d-bmw: 🔴 High: "Each command accepts -i/--input ... and -o/--output" isn't accurate. Per tools/blob/__main__.py:104-180, `pprint` has no `-o` option, and `header config-type` has no `-i` option — input/output handling differs per subcommand. Please document each command's actual CLI signature rather than a blanket statement.

libs/bsw/blob/doc/generator.rst
```diff
@@ -0,0 +1,93 @@
...
+   SPDX-License-Identifier: Apache-2.0
+   *******************************************************************************
+
+Generator
+=========
+
+The ``blob`` Python tool located in ``tools/blob`` generates blobs from a JSON Lines (``.jsonl``)
+description of the desired configurations. It emits either the binary blob or generated C++ header
+files for use in applications.
+
```
simon-d-bmw, line 18: 🟡 Medium: Missing two things a reader needs to actually use this tool: (1) the `crc==7.1.0` dependency isn't mentioned anywhere, and (2) there's no JSONL input schema or example shown before the CLI usage. Consider adding a short "Input format" subsection with a minimal .jsonl sample.

libs/bsw/blob/doc/generator.rst
```diff
@@ -0,0 +1,93 @@
...
+Commands for generating necessary blob files
+--------------------------------------------
+
+.. code-block:: console
+
+    python -m blob binary -i routing.jsonl -c blob.routing.table | python -m blob header data -n CONFIGURATION_BLOB > include/blob/configuration.h
+    python -m blob header config-type > include/blob/ConfigType.h
+    python -m blob.routing header routing.jsonl > include/routing/channelId.h
+
+This generates the configuration blob, the ``ConfigType`` enum header file, and the routing channel ID header file, which are required for the ``RoutingTable`` module to function correctly.
```
simon-d-bmw: 🟡 Medium: This section ("Routing helper" onward) is entirely routing-specific (routing tables, channelId.h, RoutingTable module, constants.h) and doesn't fit the generic `blob` library's doc set. Consider moving this content to libs/bsw/routing/doc/integration.rst and keeping this page scoped to the generic blob generator (`binary`, `pprint`, `header ...`).

libs/bsw/blob/doc/design.rst
```diff
@@ -0,0 +1,92 @@
...
+``Blob``
+    Wraps a validated blob span and exposes a forward (input) iterator over its configurations. The
+    constructor calls ``checkHeader()`` and verifies that the memory region size matches
+    ``HEADER_SIZE + size``; on failure it logs an error and yields an empty span. Iteration
+    advances by each configuration's declared size and stops once the span is exhausted.
+
+``Config``
+    Represents a single configuration as a ``type`` and a ``data`` span. ``Config::from_bytes()``
+    parses a configuration header (via ``loadConfigHeader()``), rejects ``UNKNOWN`` types and
+    inconsistent sizes, and returns a span bounded to the configuration's extent.
```
simon-d-bmw, line 36: 🟡 Medium: Not wrong on its own — `Config.data` is indeed "bounded to the configuration's extent," matching `Config::from_bytes()`. But read alongside configuration.rst's layout table, which lists `type`/`reserved`/`size` as rows separate from `data`, a reader will assume `data` means "just the payload." In the code, `Config.data` spans the entire configuration record (header fields + payload), not the payload alone.

Once the configuration.rst CRC/layout description is corrected, please also make this passage explicit that `Config.data` includes the header fields, so the two pages reinforce the same mental model instead of contradicting each other.

libs/bsw/blob/doc/integration.rst
```diff
@@ -0,0 +1,21 @@
...
+   terms of the Apache License Version 2.0 which is available at
+   https://www.apache.org/licenses/LICENSE-2.0
+
+   SPDX-License-Identifier: Apache-2.0
+   *******************************************************************************
+
+Integration
+===========
+
+The module defines the API ``load()``, which converts a pointer in memory to a span of data, after
```
simon-d-bmw, line 15: 🟢 Nitpick: "converts a pointer in memory to a span of data" is a bit imprecise — `load()`'s actual signature takes an `::etl::span<uint8_t const>` as input (Blob.h), not a raw pointer. Consider rewording to avoid implying a pointer-based API.

libs/bsw/blob/doc/api.rst
```diff
@@ -0,0 +1,34 @@
...
+    :start-after: [HeaderCheck]
+    :end-before: [HeaderCheck]
+
+The ``checkCrc()``, also called during blob loading, uses ``etl::crc32`` (standard CRC-32).
+It re-computes the CRC on the configuration's data and compares it against the previously
+computed CRC bytes at the end of the configuration's data.
+
+.. sourceinclude:: include/blob/util.h
+    :start-after: [CrcCheck]
+    :end-before: [CrcCheck]
```
simon-d-bmw, line 34: 🟡 Medium: This page covers `checkHeader()` and `checkCrc()` but omits other public, actively-used helpers: `blob::config()`, `loadConfigHeader()`, and the `loadColumn()` template (all referenced in design.rst and used in BlobTest.cpp). Consider adding a short subsection for each so the API reference is complete.

libs/bsw/routing/doc/design.rst
```diff
@@ -0,0 +1,56 @@
...
+
+.. literalinclude:: ../include/routing/PduRoutingTable.h
+   :start-after: PDU_ROUTING_TABLE_BEGIN
+   :end-before: PDU_ROUTING_TABLE_END
+   :language: c++
+
+
+``LegacyTxAdapter`` or ``PduTransportTxAdapter`` creates outgoing messages out of PDUs according to its
+``TxAdapterTable`` and implements the ``IWriter`` interface. For each message, the ``TxAdapterTable``
+determines the payload length and the PDU offset within it.
```
simon-d-bmw, line 47: 🔴 High: "LegacyTxAdapter or PduTransportTxAdapter creates outgoing messages ... according to its TxAdapterTable" isn't accurate for `PduTransportTxAdapter`. Looking at include/routing/PduTransportTxAdapter.h, it has no `TxAdapterTable` member; only `LegacyTxAdapter` uses one. Please split this into two separate descriptions, one per adapter, since they use different configuration mechanisms.

libs/bsw/routing/doc/design.rst
```diff
@@ -0,0 +1,56 @@
...
+Each routing message uses an 8-byte header followed by its payload:
+
+* 4 bytes: message ID (big endian)
+* 4 bytes: payload length in bytes (big endian)
+* N bytes: payload
+
+``RxAdapter`` or ``PduTransportRxAdapter`` extracts PDU(s) from a payload according to its
+``RxAdapterTable`` and implements the ``IReader`` interface. The ``RxAdapterTable`` determines
+which PDUs a payload contains with the message ID and the payload length. For each PDU,
+it defines a length and an offset in bytes, i.e., the position of the PDU in the payload.
```
simon-d-bmw, line 24: 🟡 Medium: "The RxAdapterTable determines which PDUs a payload contains with the message ID and the payload length" overstates what the base classes do. Per RxAdapter.cpp/LegacyRxAdapter.h, lookup is by message ID only — the payload-length check (`messageLengths`) is enforced specifically by `PduTransportRxAdapter`, not universally. Please clarify which adapter enforces length matching.

libs/bsw/routing/doc/interface.rst
```diff
@@ -0,0 +1,144 @@
...
+    :end-before: PUBLIC_API_END
+    :language: cpp
+    :dedent: 4
+
+RxAdapter
+---------
+
+The RxAdapter is an adapter that extracts PDUs from messages in the provided
+IReader, exposing them with the IReader interface.
+It counts the PDUs and bytes extracted from the messages.
```
simon-d-bmw, line 62: 🟡 Medium: This inherits an ambiguity from the code. `pduCounter()`/`_pduCounter` (RxAdapter.h:66, RxAdapter.cpp:126) actually increment once per **input message** read from `_inputReader` (on `isNewMessage`), independent of how many PDUs are subsequently extracted from it — the header's own comment says as much: "keep track of the total amount of data read from the input queue, regardless of the number of PDUs extracted" (RxAdapter.h:64-65).

So the counters are named/described in terms of "PDU" while behaving as per-message counters — the code conflates the two terms, and this doc line just carries that conflation forward. Suggest addressing both: reword this line to describe the counters as tracking input messages (not extracted PDUs), and consider whether `pduCounter()` should be renamed (e.g. `messageCounter()`) or its own comment clarified in RxAdapter.h to avoid the same confusion for future readers of the code itself.

libs/bsw/routing/doc/interface.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Its configuration is achieved using an appropriate :ref:`table<txAdapterTable>`.
+
+Public API
+++++++++++
+.. literalinclude:: ../include/routing/LegacyTxAdapter.h
+    :start-after: PUBLIC_API_BEGIN
+    :end-before: PUBLIC_API_END
+    :language: cpp
+    :dedent: 4
```
simon-d-bmw, line 144: 🟡 Medium: Several public classes aren't documented anywhere in this file or integration.rst: `Integration::init/route/...`, `PduTransportIntegration::init/activate/disable/...`, `PduTransportBufferedWriter`, `PduTransportConfig`, `ErrorHandler`, `io::udp::Receiver`/`Sender`. Given these are all part of the public API surface shipped in this PR, consider adding at least brief coverage for each so the interface reference is complete.

libs/bsw/routing/doc/index.rst
```diff
@@ -0,0 +1,67 @@
...
+   :align: center
+
+   Package routing {
+     () IReader as ir1
+     () IWriter as iw1
+
+     [Router]
+     [RxAdapter]
+     [TxAdapter]
+   }
```
simon-d-bmw, line 44: 🟡 Medium: This diagram shows a `TxAdapter` component/class, but no such class exists in the codebase — the concrete classes are `LegacyTxAdapter` and `PduTransportTxAdapter`. Please update the diagram to reflect the actual class names (same issue appears in integration.rst lines 39/134).

libs/bsw/routing/doc/integration.rst
```diff
@@ -0,0 +1,174 @@
...
+.. uml::
+   :align: center
+
+    namespace io {
+       interface IWriter {}
+       interface IReader {}
+       hide members
+   }
+   namespace routing {
+        class Integration<PDU_TRANSPORT_CHANNELS, CAN_CHANNELS, FLEXRAY_CHANNELS> {}
```
simon-d-bmw: 🔴 High: This UML shows `Integration<PDU_TRANSPORT_CHANNELS, CAN_CHANNELS, FLEXRAY_CHANNELS>` — 3 template parameters. The actual class (include/routing/Integration.h:49-55) takes 6: `MAX_NUM_PDU_TRANSPORT_CHANNELS, NUM_CAN_CHANNELS, NUM_FLEXRAY_CHANNELS, MAX_PDU_TRANSPORT_CHANNEL_ELEMENT_SIZE, MAX_CAN_CHANNEL_ELEMENT_SIZE, MAX_FLEXRAY_CHANNEL_ELEMENT_SIZE`. Please update the diagram to match.

libs/bsw/routing/doc/integration.rst
```diff
@@ -0,0 +1,174 @@
...
+       MemoryQueueReader --> MemoryQueue
+
+       namespace udp {
+           class Receiver{}
+           class Sender{}
+       }
+       hide members
+   }
+   namespace routing {
+        class PduTransportIntegration<PDU_TRANSPORT_CHANNELS> {}
```
simon-d-bmw: 🔴 High: Similarly, `PduTransportIntegration<PDU_TRANSPORT_CHANNELS>` shows 1 template parameter, but the actual class (include/routing/PduTransportIntegration.h:35) takes 2: `MAX_NUM_CHANNELS, MAX_ELEMENT_SIZE`. Please correct.

libs/bsw/routing/doc/integration.rst
```diff
@@ -0,0 +1,174 @@
...
+transport communication, except for the sockets. The sockets are provided at initialization and their type has
+to derive from ``AbstractDatagramSocket``. ``PduTransportIntegration`` separates frame creation from
+socket transmission:
+
+* ``PduTransportBufferedWriter`` batches outgoing PDU transport messages into frame-sized queue
+  entries
+* ``checkTransmissionTimeouts()`` evaluates whether buffered frames must be committed to TX queues because
+  of configured timeouts
+* ``sendUdpFrames()`` begins the actual UDP transmission by reading frames from the TX queue and sending
+  them through sockets
```
simon-d-bmw: 🟡 Medium: "sendUdpFrames() begins the actual UDP transmission by reading frames from the TX queue and sending them" reads as if it drains the whole queue. Per include/routing/PduTransportIntegration.h:168-172, it sends at most 10 frames per call. This is an important detail for callers (e.g. whether they need to call it in a loop) — please document the cap explicitly.

libs/bsw/routing/doc/integration.rst
```diff
@@ -0,0 +1,174 @@
...
+===================
+
+Integration
+-----------
+
+For projects using FlexRay, CAN, and PDU transport, the ``Integration`` class provides
+a ready-to-go implementation of the instantiation of the routing components.
+
+The ``Integration`` class uses the ``blob`` module to access configurations, which can be loaded
+from memory using the ``load`` functions.
```
simon-d-bmw, line 22: 🟡 Medium: This page jumps straight into the Integration/PduTransportIntegration class structure but never walks through the end-to-end integration flow: generating the blob config via tools/blob, loading it at runtime, and wiring it into these classes. Consider adding a concrete "Getting started" walkthrough here.

libs/bsw/routing/doc/resources/routing_example.dot
```diff
@@ -0,0 +1,384 @@
+digraph Routing {
```
simon-d-bmw, line 1: 🟡 Medium: This file isn't referenced or embedded anywhere in the routing docs (no `.. graphviz::` directive, image, or link in any .rst file) — it's dead weight in this PR. It's also inconsistent with sibling modules (doip, uds), which use PlantUML (.puml) for diagrams rather than raw Graphviz .dot files. On top of that, its field labels (inputMessageOffsets, inputMessageIds) are stale — the current PduRoutingTable uses destinations, destinationOffsets, outputMessageIds instead.

Suggestion: remove this file. If a worked example is valuable, it'd be better regenerated fresh (via the `visualize` command in tools/blob/routing) and actually embedded in integration.rst or design.rst — otherwise it'll go stale again.

simon-d-bmw (commented):
Thank you for adding the documentation, I really appreciate it.

danielfaleirosantos:
After discussion with @simon-d-bmw, the comments shall be addressed in a follow-up PR, as to not block the integration of the modules.

executables/referenceApp/application/include/systems/RoutingSystem.h
```diff
@@ -0,0 +1,187 @@
...
+    class CanOutputWriter final : public ::io::IWriter
+    {
+    public:
+        CanOutputWriter(::can::ICanTransceiver& canTransceiver) : _canTransceiver(canTransceiver) {}
+
+        size_t maxSize() const override { return MAX_CAN_QUEUE_ELEMENT_SIZE; }
+
+        ::etl::span<uint8_t> allocate(size_t size)
+        {
+            _currentFrame = ::etl::make_span(_data).first(size);
```
rolandreichweinbmw, line 119: According to the contract of IWriter

https://github.com/eclipse-openbsw/openbsw/blob/b3ab7a808ba8b6bf4ab0afbfbc9757c39d0773db/libs/bsw/io/include/io/IWriter.h#L36

allocate() must return an empty span if size > maxSize().

In the current implementation, this case is ignored.

I.e. potential buffer overflow / UB.

rolandreichweinbmw (changes requested):

rolandreichweinbmw:
I agreed with @danielfaleirosantos that a convenience wrapper to just regenerate all generated files is the right approach.

Also, we need to rebase to latest main. Then, the branch is ready to merge from my point of view.

danielfaleirosantos:
Convenience wrapper and rebase to main are now done. FYI @rolandreichweinbmw

rolandreichweinbmw (approved):

executables/referenceApp/application/src/systems/RoutingSystem.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+void RoutingSystem::handleError(
+    ::routing::ErrorHandler::StatusCode const statusCode,
+    uint8_t const /*channelId*/,
+    uint32_t const /*data*/)
+{
+    switch (statusCode)
+    {
+        case ::routing::ErrorHandler::StatusCode::INVALID_MESSAGE_LENGTH:
+            logger::Logger::debug(logger::ROUTING, "Invalid message length");
```
matthiaskessler, line 164: I find it a bit strange to do Logger::debug in a `handleError` function, but we can change that later. It depends on how frequent the errors are. If some "errors" are expected, we could rename the function to `handleEvent` or sth. like this and then decide in this function what to treat as an error.

executables/referenceApp/application/CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ if (PLATFORM_SUPPORT_ETHERNET)
         target_sources(app.referenceApp
                        PRIVATE src/systems/DoIpServerSystem.cpp)
     endif ()
+
+    if (PLATFORM_SUPPORT_CAN)
```
matthiaskessler, line 27: In `app.cpp` we only add the routing system, when both `PLATFORM_SUPPORT_CAN` and `PLATFORM_SUPPORT_ETHERNET` are true.

libs/bsw/blob/doc/index.rst
```diff
@@ -0,0 +1,30 @@
...
+
+   SPDX-License-Identifier: Apache-2.0
+   *******************************************************************************
+
+.. _blob:
+
+blob
+====
+
+This module loads from memory the binary representation of a generated gateway table, so as to
```
matthiaskessler, line 17: NIT: Maybe we could think of "routing table" instead of "gateway table"? Just a thought and we can do it later. Most importantly we should be consistent :-)

matthiaskessler (approved):

danielfaleirosantos:
@matthiaskessler I'll pass on these comments to Felipe Prado. Thanks!!

executables/referenceApp/application/CMakeLists.txt
matthiaskessler: Oh, I saw it's nested already.

PR #500, opened 2026-06-30
Add TimestampProvider

rolandreichweinbmw:
Abstraction to prevent application code use BSP directly


simon-d-bmw:
[important] The new TimestampProvider does not yet achieve the stated goal of preventing application code from using BSP directly. Reference app code still includes/calls BSP time APIs directly (for example in app.cpp, DoCanSystem.cpp, and DemoSystem.cpp), and the wrapper currently only covers microsecond accessors.

rolandreichweinbmw:
> [important] The new TimestampProvider does not yet achieve the stated goal of preventing application code from using BSP directly. Reference app code still includes/calls BSP time APIs directly (for example in app.cpp, DoCanSystem.cpp, and DemoSystem.cpp), and the wrapper currently only covers microsecond accessors.

Agree. Added the switchover to the new API and also milliseconds interfaces to TimestampProvider.

simon-d-bmw (dismissed):
lgtm

libs/bsw/time/include/time/TimestampProvider.h
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * A 32 bit timestamp is less expensive compared to a 64 bit timestamp, at least on 32 bit
+     * platforms. Therefore prefer 32 bit timestamp if feasible.
+     *
+     * \return  32 bit CPU local timestamp micro seconds.
+     */
+    static uint32_t getTimestampUs32Bit();
+
+    /**
+     * Provides a 64 bit CPU local and relative timestamp in micro seconds resolution.
```
rolandreichweinbmw, line 41: Document that implementations overflow only at 2^64, not earlier.

PR #501, opened 2026-06-30
Migrate ThreadX RTOS to Bazel build system

SuhashiniNaik:
Migrate ThreadX RTOS to Bazel build system

New BUILD files:
- libs/3rdparty/threadx: Core ThreadX library with label_flags for
  port, port_impl, os_hooks and core_configuration selection
- libs/bsw/asyncThreadX: async_threadx, async_threadx_impl and
  threadx_configuration targets with threadx_core_configuration
  label_flag
- platforms/s32k1xx/3rdparty/threadx: Cortex-M4 port implementation and port headers for s32k148
- platforms/posix/3rdparty/threadx: Linux port implementation and
  port headers for POSIX
- executables/referenceApp/platforms/s32k148evb/main: threadx_os_hooks
- executables/referenceApp/platforms/s32k148evb/threadXCoreConfiguration
- executables/referenceApp/platforms/posix/threadXCoreConfiguration

Modified BUILD files:
- libs/bsw/async: Added support_threadx branch to async_platform_default
- platforms/s32k1xx/bsp/bspInterruptsImpl: Added threadx RTOS headers
  and wired into bsp_interrupts_impl_rtos_hdrs alias
- platforms/posix/bsp/bspInterruptsImpl: Added threadx variant

Key Bazel patterns used:
- - label_flag + alias + select() for platform/RTOS-specific selection
- target_compatible_with to restrict platform-specific targets
- async_threadx_impl and async_freertos_impl marked incompatible for
  opposite RTOS configs to prevent build failures in bazel build //...
- Circular dependency threadX -> osHooks -> asyncThreadX -> threadX
  broken by depending on threadx_headers instead of threadx

Added --config=s32k148_threadx to .bazelrc combining s32k148 platform
and threadx RTOS flag.

### DIFF
--- a/libs/3rdparty/threadx/BUILD.bazel
+++ b/libs/3rdparty/threadx/BUILD.bazel
@@ -1,0 +1,99 @@
+# *******************************************************************************
+# Copyright (c) 2026 Accenture
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+alias(
+    name = "threadx_port_default",
+    actual = select(
+        {
+            "//bazel/platform/constraints/soc:s32k148": "//platforms/s32k1xx/3rdparty/threadx:threadx_cortex_m4_port_headers",
+            "@platforms//os:linux": "//platforms/posix/3rdparty/threadx:threadx_linux_port_headers",
+        },
+        no_match_error = "No ThreadX port for this platform. Override --//libs/3rdparty/threadx:threadx_port_headers with your own port target.",
+    ),
+)
+
+# ThreadX port (tx_port.h).
+#   Default:  select based on the target platform (S32K148, POSIX)
+#   Override: --//libs/3rdparty/threadx:threadx_port_headers=//path/to:your_port_headers
+label_flag(
+    name = "threadx_port_headers",
+    build_setting_default = ":threadx_port_default",
+    visibility = ["//visibility:public"],
+)
+
+alias(
+    name = "threadx_port_impl_default",
+    actual = select(
+        {
+            "//bazel/platform/constraints/soc:s32k148": "//platforms/s32k1xx/3rdparty/threadx:threadx_cortex_m4",
+            "@platforms//os:linux": "//platforms/posix/3rdparty/threadx:threadx_linux",
+        },
+        no_match_error = "No ThreadX port implementation for this platform. Override --//libs/3rdparty/threadx:threadx_port_impl with your own port target.",
+    ),
+)
+
+# ThreadX port implementation
+#   Default:  select based on the target platform (S32K148, POSIX)
+#   Override: --//libs/3rdparty/threadx:threadx_port_impl=//path/to:your_port_impl
+label_flag(
+    name = "threadx_port_impl",
+    build_setting_default = ":threadx_port_impl_default",
+    visibility = ["//visibility:public"],
+)
+
+alias(
+    name = "threadx_os_hooks_default",
+    actual = select(
+        {
+            "//bazel/platform/constraints/soc:s32k148": "//executables/referenceApp/platforms/s32k148evb/main:threadx_os_hooks",
+            "@platforms//os:linux": "//executables/referenceApp/platforms/posix/main:threadx_os_hooks",
+        },
+        no_match_error = "No ThreadX OS hooks for this platform. Override --//libs/3rdparty/threadx:threadx_os_hooks with your own hooks target.",
+    ),
+)
+
+# ThreadX OS hooks
+#   Default:  select based on the target platform (S32K148, POSIX)
+#   Override: --//libs/3rdparty/threadx:threadx_os_hooks=//path/to:your_os_hooks
+label_flag(
+    name = "threadx_os_hooks",
+    build_setting_default = ":threadx_os_hooks_default",
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "threadx_headers",
+    hdrs = glob(["common/inc/**/*.h"]),
+    includes = ["common/inc"],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":threadx_port_headers",
+        "//libs/bsw/asyncThreadX:threadx_configuration",
+    ],
+)
+
+# Circular dependency present on the CMake side: threadX -> osHooks -> asyncThreadX -> threadX
+# Split off threadx_headers to break the circular dependency for Bazel
+cc_library(
+    name = "threadx",
+    srcs = glob(["common/src/*.c"]),
+    copts = ["-Wno-error=unused-parameter"],
+    implementation_deps = [
+        ":threadx_os_hooks",
+        ":threadx_port_impl",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":threadx_headers",
+        "//libs/bsw/platform",
+    ],
+)
--- a/libs/bsw/asyncThreadX/BUILD.bazel
+++ b/libs/bsw/asyncThreadX/BUILD.bazel
@@ -1,0 +1,91 @@
+# *******************************************************************************
+# Copyright (c) 2026 Accenture
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+alias(
+    name = "threadx_core_configuration_default",
+    actual = select(
+        {
+            "//bazel/platform/constraints/soc:s32k148": "//executables/referenceApp/platforms/s32k148evb/threadXCoreConfiguration:threadx_core_configuration",
+            "@platforms//os:linux": "//executables/referenceApp/platforms/posix/threadXCoreConfiguration:threadx_core_configuration",
+        },
+        no_match_error = "No ThreadX core configuration for this platform. Override --//libs/bsw/asyncThreadX:threadx_core_configuration with your own configuration target.",
+    ),
+)
+
+# ThreadX core configuration (Hook.h, ThreadCConfig.h and platform wiring).
+#   Default:  select based on the target platform (S32K148, POSIX)
+#   Override: --//libs/bsw/asyncThreadX:threadx_core_configuration=//path/to:your_config
+label_flag(
+    name = "threadx_core_configuration",
+    build_setting_default = ":threadx_core_configuration_default",
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "async_threadx",
+    hdrs = glob(["include/**/*.h"]),
+    strip_include_prefix = "include",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//libs/3rdparty/threadx",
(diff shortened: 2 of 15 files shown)


libs/bsp/bspInterrupts/BUILD.bazel
```diff
@@ -42,3 +42,23 @@ cc_library(
         ":bsp_interrupts_impl",
     ],
 )
+
+alias(
+    name = "bsp_interrupts_threadx_impl_default",
```
christophruethingbmw: Why do we need this here explicitly for threadx? We already have the `bsp_interrupts_impl` above which should  end up through the select to the right one?

`bsp_interrupts_impl_default` selects for s32k vs linux and the platform specific `bsp_interrupts_impl` has a select for freertos vs threadx.

platforms/posix/3rdparty/threadx/BUILD.bazel
```diff
@@ -0,0 +1,54 @@
...
+    name = "threadx_linux_port_headers",
+    hdrs = [
+        "ports/linux/gnu/inc/tx_port.h",
+    ],
+    defines = [
+        "TX_INCLUDE_USER_DEFINE_FILE",
+        "_GNU_SOURCE",
+        "TX_LINUX_DEBUG_ENABLE",
+    ],
+    linkopts = ["-lpthread"],
```
christophruethingbmw, line 47: Do we need this here since it is a header-only thing?

platforms/posix/3rdparty/threadx/BUILD.bazel
```diff
@@ -0,0 +1,54 @@
...
+    visibility = ["//visibility:public"],
+    deps = [
+        "//libs/3rdparty/threadx:threadx_headers",
+        "//libs/bsp/bspInterrupts:bsp_interrupts",
+        "//libs/bsw/asyncThreadX:threadx_core_configuration",
+    ],
+)
+
+cc_library(
+    name = "threadx_linux_port_headers",
```
christophruethingbmw, line 38: Can we name this in the same way here and for the m4 port below? Here we name it `threadx_linux_port_headers`, below we name it `threadx_cortex_m4_port` without the `_headers` postfix.

libs/3rdparty/threadx/BUILD.bazel
```diff
@@ -0,0 +1,99 @@
...
+        },
+        no_match_error = "No ThreadX port for this platform. Override --//libs/3rdparty/threadx:threadx_port with your own port target.",
+    ),
+)
+
+# ThreadX port (tx_port.h).
+#   Default:  select based on the target platform (S32K148, POSIX)
+#   Override: --//libs/3rdparty/threadx:threadx_port=//path/to:your_port
+label_flag(
+    name = "threadx_port",
```
christophruethingbmw: These are just headers, right? Should we also follow the pattern to name it `_headers`?

libs/3rdparty/threadx/BUILD.bazel
```diff
@@ -0,0 +1,99 @@
...
+    visibility = ["//visibility:public"],
+    deps = [
+        ":threadx_port",
+        "//libs/bsw/asyncThreadX:threadx_configuration",
+    ],
+)
+
+# Circular dependency present on the CMake side: threadX -> osHooks -> asyncThreadX -> threadX
+# Split off threadx_headers to break the circular dependency for Bazel
+cc_library(
```
christophruethingbmw, line 86: Just for me to sort it out

```
threadx
-> threadx_headers
  -> threadx_port -> //platforms/.../3rdparty/threadx:threadx_..._port (provides tx_port.h including tx_user.h)
-> threadx_os_hooks -> //executables/referenceApp/platforms/.../main:threadx_os_hooks
-> threadx_port_impl -> //platforms/.../3rdparty/threadx:threadx_... (provides port implementation including tx_user.h)
```

It feels still not right to refer from this very generic part to the reference application, but that's how it is also done in CMake I think, right? So we take it over for now and need to refactor it later in both CMake and Bazel?

libs/bsw/asyncThreadX/BUILD.bazel
```diff
@@ -0,0 +1,77 @@
...
+    name = "threadx_configuration",
+    hdrs = [
+        "threadXConfiguration/ThreadXConfig.h",
+        "threadXConfiguration/async/Hook.h",
+    ],
+    strip_include_prefix = "threadXConfiguration",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":threadx_core_configuration",
+        "//libs/bsw/asyncFreeRtos:async_core_configuration",
```
christophruethingbmw: This looks strange? We have threadx configuration but we refer to freertos?

libs/bsw/asyncThreadX/BUILD.bazel
```diff
@@ -0,0 +1,77 @@
...
+cc_library(
+    name = "threadx_configuration",
+    hdrs = [
+        "threadXConfiguration/ThreadXConfig.h",
+        "threadXConfiguration/async/Hook.h",
+    ],
+    strip_include_prefix = "threadXConfiguration",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":threadx_core_configuration",
```
christophruethingbmw, line 68: Just to understand, we refer here to the `async_core_configuration` which provides the `tx_user.h` right? But nothing else? Do we use `tx_user.h` here directly at all?

christophruethingbmw (commented):
Looks good overall, but some detailled questions as commented below.

libs/bsp/bspInterrupts/BUILD.bazel
SuhashiniNaik: Thank you for the review, You're right, I tried using `bsp_interrupts_impl ` directly and got this error on a  `bazel build //...`:

```
ERROR: libs/bsw/asyncThreadX/BUILD.bazel:63:11: Compiling libs/bsw/asyncThreadX/src/async/Types.cpp failed
In file included from async/AsyncBinding.h:23,
                 from libs/bsw/asyncThreadX/src/async/Types.cpp:11:
async/ThreadXAdapter.h:250:5: error: there are no arguments to 'setThreadXInitialized' that depend on a template parameter, so a declaration of 'setThreadXInitialized' must be available [-fpermissive]
```

The reason is that` //bazel/config/rtos` defaults to 'freertos', so on a `bazel build //... `with no flags, `bsp_interrupts_impl `resolves to `bsp_interrupts_impl_freertos `which doesn't declare `setThreadXInitialized`.

Since `async_threadx_impl ` is a ThreadX-specific compilation unit. It needs the ThreadX interrupt headers unconditionally, regardless of what //bazel/config/rtos is set to. `bsp_interrupts_threadx_impl` solves this by selecting only on platform (s32k148/posix), then resolving to the ThreadX variant of the interrupt impl.

Is there a better approach you'd suggest without changing the global rtos default?

platforms/posix/3rdparty/threadx/BUILD.bazel
SuhashiniNaik: Yes  `linkopts` are needed here,  they're the Bazel equivalent of CMake's INTERFACE properties on` threadXLinuxPort`, which also propagates `Threads::Threads` and `TX_INCLUDE_USER_DEFINE_FILE` without compiling anything.

libs/3rdparty/threadx/BUILD.bazel
SuhashiniNaik: Yes, changed it to` threadx_port_headers`

libs/3rdparty/threadx/BUILD.bazel
SuhashiniNaik: Yes, this mirrors CMake parity. threadX depends on osHooks which comes from the referenceApp. It's not ideal but it's CMake-parity for now. We should refactor in both CMake and Bazel together later.

libs/bsw/asyncThreadX/BUILD.bazel
SuhashiniNaik: `async_core_configuration` provides` async/Config.h` which is directly included by `ThreadXConfig.h`. `tx_user.h `is not included here directly but there is a comment  in  `ThreadXConfig.h`  documents that it gets pulled in transitively through the TX_INCLUDE_USER_DEFINE_FILE mechanism via `threadx_core_configuration`. So `async_core_configuration` is needed here  for `async/Config.h` I think.

libs/bsw/asyncThreadX/BUILD.bazel
SuhashiniNaik: Ah Yes makes sense, 
Removed it and defined a separate async_core_configuration label_flag  pointing to the `executables/referenceApp/asyncCoreConfiguration:async_core_configuration ` .

platforms/posix/3rdparty/threadx/BUILD.bazel
SuhashiniNaik: Done

libs/3rdparty/threadx/BUILD.bazel
christophruethingbmw: Okay, makes sense, I think that's for sure something we should consider in the future. In case CMake does it in the same way right now, we can keep it also in Bazel to keep it synchronous.

libs/bsw/asyncThreadX/BUILD.bazel
christophruethingbmw: Okay, I see not a problem with this, but if it is technically not needed, I would actually keep it out, does CMake also add this dependency? When you leave it out everything still compiles properly?

libs/bsp/bspInterrupts/BUILD.bazel
christophruethingbmw: Let me understand, you mean that we `build //...` and then end up building some threadx target which then takes a freertos implementation? But is putting then this special alias here the right approach? In case we change a default somewhere, this breaks and we would need to adapt it?

Would it make sense to introduce a `target_compatible_with` such that the threadx targets are excluded from `//...`? I agree, that we do not build everything then with `//...` but we anyway don't do this because everything else that deals with the default will also be excluded somehow?

SuhashiniNaik: I investigated this further and you're right that the  `bsp_interrupts_threadx_impl` alias is not the right approach.

I removed the alias again and switched `async_threadx` back to using the `//libs/bsp/bspInterrupts:bsp_interrupts_impl`.

The underlying issue turned out to be that `async_threadx_impl` itself is a ThreadX-specific implementation target. With the default `rtos=freertos`, it ends up pulling in the Freertos binding and fails to compile, whereas:

```bash
bazel build --//bazel/config/rtos=threadx //libs/bsw/asyncThreadX:async_threadx_impl
```

builds successfully.

Following your suggestion, I added `target_compatible_with` to `async_threadx_impl` in [libs/bsw/asyncThreadX/BUILD.bazel](https://github.com/eclipse-openbsw/openbsw/pull/501/files#diff-0aedf84ebf06bae1e95470f4e2505dad611029a45ffd0ecd3aa264d03d2fd7f2) so that it is excluded from non-ThreadX configurations.

libs/bsw/asyncThreadX/BUILD.bazel
SuhashiniNaik: That makes sense, I tried removing `:async_core_configuration`.
`bazel build //libs/bsw/asyncThreadX:threadx_configuration` still succeeds, but in the CI check:
```
bazel build //...
```
fails with:
```
tx_user.h:16:10: fatal error: async/Config.h: No such file or directory
```

`ThreadXConfig.h` also directly includes `async/Config.h`, and CMake models this relationship as:

```
target_link_libraries(
    threadXConfiguration INTERFACE
        asyncCoreConfiguration
        platform
        threadXCoreConfiguration
)
```
So it looks like `:async_core_configuration` is indeed required, and removing it breaks downstream consumers that rely on `tx_user.h`.

christophruethingbmw (approved):

PR #502, opened 2026-06-30
Replace spsc queue by etl implementation and remove dead code

rolandreichweinbmw:
OpenBSW's spsc queue was based on etl::atomic already.


platforms/posix/bsp/tapEthernetDriver/src/TapEthernetDriver.cpp
```diff
@@ -137,7 +136,11 @@ void TapEthernetDriver::readFrame()
         delete driverPbuf;
     };
 
-    sender.write(&frameBuf->buf.pbuf);
+    if (!_queue.push(&frameBuf->buf.pbuf))
```
christian-schilling: This is unreachable (but still overhead):
The function checks for `full()` at entry and since this is single producer the queue can't become full during this function.

platforms/s32k1xx/bsp/bspEthernet/src/ethernet/RxBuffers.cpp
```diff
@@ -125,8 +124,11 @@ void RxBuffers::interrupt()
             {
                 // allocation was successful
 
-                sender.next() = buf;
-                sender.write_next();
+                if (!_queue.push(buf))
```
christian-schilling: Also unreachable.

platforms/posix/bsp/tapEthernetDriver/src/TapEthernetDriver.cpp
rolandreichweinbmw: Agree. Removed now.

platforms/s32k1xx/bsp/bspEthernet/src/ethernet/RxBuffers.cpp
rolandreichweinbmw: Agreed. Also simplified now.

libs/bsw/lwipSocket/src/lwipSocket/utils/FilterFrame.cpp
```diff
@@ -101,7 +101,11 @@ bool processPbufQueue(
     auto const queued = receiver.size();
     for (size_t i = 0; i < queued; ++i)
     {
-        auto* const p      = receiver.read();
+        pbuf* p = nullptr;
+        if (!receiver.pop(p))
```
christian-schilling: This is also unreachable. We get the size at the beginning and only do as much iterations. Since this is single consumer we can't underrun the queue.

rolandreichweinbmw: Right!

christian-schilling (approved):

PR #503, opened 2026-06-30
refactor(Middleware): return payload by const reference

dcavalei:
When sizeof(T) is large, returning by value risks blowing the stack depending on system configuration. readPayload<T>() now returns a T const& directly into the middleware-managed buffer. Callers must consume the reference before the message is deallocated, as the middleware reclaims the buffer immediately after dispatch.

- readPayload<T>() now returns T const& instead of T, eliminating unnecessary copies when reading from message payloads

- FutureDispatcher callbacks now receive etl::reference_wrapper<ArgumentType const> instead of T by value

- Replace allocateTrivialType/allocateNonTrivialType with emplaceExternalPayload (uses placement new for proper construction) and allocAndCopyBytesToExternalPayload (for byte span copies)

- Drop AllocationPolicy-based serialization path; T must now be copy-constructible (enforced via static_assert)

### DIFF
--- a/libs/bsw/middleware/include/middleware/core/FutureDispatcher.h
+++ b/libs/bsw/middleware/include/middleware/core/FutureDispatcher.h
@@ -19,7 +19,7 @@
 #include <etl/array.h>
 #include <etl/delegate.h>
 #include <etl/expected.h>
-#include <etl/tuple.h>
+#include <etl/functional.h>
 #include <etl/vector.h>
 
 #include <cstdint>
@@ -57,7 +57,8 @@ struct CallbackHelper;
 template<typename ArgumentType>
 struct CallbackHelper<ArgumentType, etl::enable_if_t<(etl::is_void<ArgumentType>::value == false)>>
 {
-    using Callback = ::etl::delegate<void(etl::expected<ArgumentType, Future::State> const&)>;
+    using Result   = ::etl::expected<etl::reference_wrapper<ArgumentType const>, Future::State>;
+    using Callback = ::etl::delegate<void(Result const&)>;
 
     static void call(Callback const& callback, Message const& msg, Future::State const state)
     {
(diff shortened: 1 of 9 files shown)


libs/bsw/middleware/include/middleware/core/MessagePayloadBuilder.h
```diff
@@ -53,76 +54,54 @@ class MessagePayloadBuilder final
...
-                T::AllocationPolicy::serialize(obj, buffer.value());
-                ret = HRESULT::Ok;
-            }
-
-            return ret;
-        }
-        else if constexpr (sizeof(T) <= Message::MAX_PAYLOAD_SIZE)
+        static_assert(etl::is_copy_constructible_v<T>, "T must be copy-constructible");
+        static_assert(!etl::is_span_v<T>, "Use the span overload to allocate bytes");
+
```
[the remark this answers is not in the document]
dcavalei: this is by design and it's Okaish,`T` might not always be trivially destructible because middleware supports `etl::vector` (which is not trivially destructable), but the leaf types are always trivially destructable, so its fine to just drop the ptr without calling the destructor.

rolandreichweinbmw (approved):

libs/bsw/middleware/include/middleware/core/FutureDispatcher.h
```diff
@@ -57,7 +57,8 @@ struct CallbackHelper;
 template<typename ArgumentType>
 struct CallbackHelper<ArgumentType, etl::enable_if_t<(etl::is_void<ArgumentType>::value == false)>>
 {
-    using Callback = ::etl::delegate<void(etl::expected<ArgumentType, Future::State> const&)>;
+    using Result   = ::etl::expected<etl::reference_wrapper<ArgumentType const>, Future::State>;
```
rolandreichweinbmw, line 60: Can you please make namespace usage consistent? I.e. fully qualified ::etl?

lcastela: this pull request's changes were taken into https://github.com/eclipse-openbsw/openbsw/pull/509 and this was addressed there: https://github.com/eclipse-openbsw/openbsw/pull/509/changes/7b98607ae81f0971ce5a66884d2f42c7c46e5377

we can close this pr and follow it up there

rolandreichweinbmw: Ok, thanks! Doing so now.

PR #506, opened 2026-07-01
Replace local CRC implementation with using ETL

rolandreichweinbmw:


rolandreichweinbmw:
The remaining specific CRC implementations can also be removed.

PR #509, opened 2026-07-03
Add middleware to ref app demo system

lcastela:
## Summary - Middleware Simulation
This PR introduces a helper utility Jinja2-based C++ code generator for the OpenBSW middleware service layer and demonstrates its use with a stand-alone POSIX simulation and a referenceApp integration with respective pytests.

This should also allow the investigation for the Middleware rust bindings to go further. fyi @ThoFrank @M-Hassanin  @ Ishan Seth

## Summary Message Allocation changes
Changes from: https://github.com/eclipse-openbsw/openbsw/pull/503

When sizeof(T) is large, returning by value risks blowing the stack depending on system configuration. readPayload() now returns a T const& directly into the middleware-managed buffer. Callers must consume the reference before the message is deallocated, as the middleware reclaims the buffer immediately after dispatch.

- readPayload() now returns T const& instead of T, eliminating unnecessary copies when reading from message payloads
- FutureDispatcher callbacks now receive etl::reference_wrapper instead of T by value
- Replace allocateTrivialType/allocateNonTrivialType with emplaceExternalPayload (uses placement new for proper construction) and allocAndCopyBytesToExternalPayload (for byte span copies)
- Drop AllocationPolicy-based serialization path; T must now be copy-constructible (enforced via static_assert)

A follow-up bug was identified: storing non-trivially-copyable types (e.g. etl::vector) in the Message's internal buffer via placement-new is unsafe. The Message is a 64-byte cache-line object transported across cores and queues as raw bytes. A type that holds internal pointers into its own storage (such as any ETL container) produces dangling pointers when the Message is byte-copied.
- The internal buffer path is now restricted to trivially-copyable types only. Non-trivially-copyable types are always routed through external allocation, regardless of sizeof(T), to survive Message relocation safely.


lcastela:
this is now ready for review and merge

libs/bsw/middleware/test/src/core/MessagePayloadBuilderTest.cpp
```diff
@@ -565,4 +604,34 @@ TEST_F(TestMessagePayloadBuilder, AllocateExternalSharedFailsForSpan)
     MessagePayloadBuilder::deallocate(msg3);
 }
 
+TEST_F(TestMessagePayloadBuilder, InternalPayloadSurvivesMessageRelocation)
```
filipe-cuim: What are we really testing here? I was expecting that we just want to guarantee that the payload of a message is still valid after a Message object has been copy constructed (just like what happens when we push a Message into a queue object)

lcastela: you are right, that unit test is redundant because the behavior is already covered in the introduced typed_tests. will remove it. thanks for the catch

filipe-cuim (approved):

rolandreichweinbmw (dismissed):

executables/referenceApp/application/include/systems/DemoSystem.h
```diff
@@ -42,6 +42,13 @@
 #include <udp/util/UdpIperf2Server.h>
 #endif
 
+#ifdef PLATFORM_SUPPORT_MIDDLEWARE
+#include "middleware/ClusterCore0.h"
```
matthiaskessler: I find the names of the clusters a bit confusing and wouldn't use Core0 and Core1.
If you think about potential use cases of middleware come to my mind:
- communication between components on different ECUs
- communication between components on different cores
- communication between components in different tasks
- communication between components in the same task
Since we currently are a single core system, I'd suggest to just use "Cluster0" and "Cluster1" but not use the name Core in it.

libs/bsw/middleware/tools/cpp_generator/templates/jinja/cluster_srcCluster.cpp.jinja
```diff
@@ -0,0 +1,96 @@
...
+{% for connection in connections if connection.source_cluster.name == cluster.name %}
+meta::ClusterConnection{{cluster.name}}To{{connection.target_cluster.name}}Meta ClusterConnection{{cluster.name}}To{{connection.target_cluster.name}}Config;
+core::ClusterConnectionTypeSelector<meta::ClusterConnection{{cluster.name}}To{{connection.target_cluster.name}}Meta>::type ClusterConnection{{cluster.name}}To{{connection.target_cluster.name}}(
+    ClusterConnection{{cluster.name}}To{{connection.target_cluster.name}}Config);
+{% endfor %}
+
+void initialize{{cluster.name}}ClusterConnection()
+{
+{% for connection in connections if connection.source_cluster.name == cluster.name %}
+    static_cast<void>(new (&ClusterConnection{{cluster.name}}To{{connection.target_cluster.name}}Config) decltype(ClusterConnection{{cluster.name}}To{{connection.target_cluster.name}}Config)());
```
matthiaskessler, line 88: AI review:

UB by C++17 [basic.life]/8, which defines the conditions under which using the original variable name after placement-new is legal. 

**What the code does**
```
// (1) File scope — both objects are constructed at static-init time:
meta::ClusterConnectionCore0ToCore1Meta ClusterConnectionCore0ToCore1Config;
ClusterConnectionTypeSelector<...>::type  ClusterConnectionCore0ToCore1(
    ClusterConnectionCore0ToCore1Config);   // ← object A, lifetime begins

void initializeCore0ClusterConnection()
{
    // (2) Placement-new reuses the storage — object A's lifetime ENDS here,
    //     object B's lifetime begins at the same address:
    static_cast<void>(new (&ClusterConnectionCore0ToCore1)
        decltype(ClusterConnectionCore0ToCore1)(
            ClusterConnectionCore0ToCore1Config));  // ← object B
}

// (3) Later, in processCore0Cluster():
ClusterConnectionCore0ToCore1.processMessage(msg);  // ← which object does this name refer to?
```
The rule says that after object A's lifetime ends and object B is created at the same address, the original name transparently refers to B — but only if all four conditions hold, including:

> (8.3) the type of the original object is not const-qualified, and, if a class type, does not contain any non-static data member whose type is const-qualified or is a reference type


Why condition (8.3) fails here
ClusterConnectionCore0ToCore1's type derives from ClusterConnectionBase, which has:
```
// ClusterConnectionBase (line 85):
IClusterConnectionConfigurationBase& _configuration;  // ← reference member

```
Because the type contains a reference-type data member, condition (8.3) is not satisfied. 
Therefore:
• The name ClusterConnectionCore0ToCore1 does not automatically refer to object B
• The returned pointer from placement-new (discarded via static_cast<void>) was the only valid way to reach B
• Every subsequent use of the name — ClusterConnectionCore0ToCore1.processMessage(msg), .updateTimeouts(), etc. — is accessing an object whose lifetime has ended → undefined behaviour


**Possible Fix**
```
// Use uninitialized storage at file scope — no object is constructed yet:
::etl::typed_storage<ClusterConnectionTypeSelector<...>::type>  ClusterConnectionCore0ToCore1;
::etl::typed_storage<meta::ClusterConnectionCore0ToCore1Meta>   ClusterConnectionCore0ToCore1Config;

void initializeCore0ClusterConnection()
{
    // Construct in-place for the first (and only) time:
    ClusterConnectionCore0ToCore1Config.create();
    ClusterConnectionCore0ToCore1.create(*ClusterConnectionCore0ToCore1Config);
}
```

lcastela: @matthiaskessler thanks for the heads-up, will have a look into it and provide the fix in a follow-up pull request.

Issue #507, opened 2026-07-02
TransportRouterSimple references unguarded busid::ETH_1


jiyuanwangbmw:
**Describe the bug**
The generic library `transportRouterSimple` unconditionally references the
project-specific bus ids `::busid::ETH_0` and `::busid::ETH_1` in the helper
`is1ByteDiagAddressBus()` in
`libs/bsw/transportRouterSimple/src/transport/routing/TransportRouterSimple.cpp`:

    bool is1ByteDiagAddressBus(uint8_t const busId)
    {
        return (busId != ::busid::ETH_0) && (busId != ::busid::ETH_1);
    }

Because this reference is not guarded, any project whose `busid/BusId.h` does not
define `ETH_1` (e.g. a single-Ethernet or a no-Ethernet integration) fails to
compile the shared library with
`error: no member named 'ETH_1' in namespace 'busid'`.

This is the same failure mode as [#349](https://github.com/eclipse-openbsw/openbsw/issues/349), which was fixed for `ETH_0` by guarding its
usage behind `PLATFORM_SUPPORT_ETHERNET` (PR [#358](https://github.com/eclipse-openbsw/openbsw/pull/358), commit `4afe34f`). Commit
`120f568` ("Add tester address translation to referenceApp TP-Router") later
introduced `is1ByteDiagAddressBus()` referencing `ETH_1` (and `ETH_0`) without any
guard, re-introducing the regression — now for `ETH_1`.

**Steps to reproduce the bug**
1. Provide a `busid/BusId.h` that defines `SELFDIAG`, `CAN_0`, and `ETH_0` but not
   `ETH_1` (a single-Ethernet integration), or one with no Ethernet buses at all.
2. Build the `transportRouterSimple` library against that configuration.
3. Compilation fails on `is1ByteDiagAddressBus()`:

       error: no member named 'ETH_1' in namespace 'busid'

   (The same happens for `ETH_0` when `PLATFORM_SUPPORT_ETHERNET` is OFF, which is
   exactly the case #349 addressed.)

**Expected behavior and actual behavior**
Expected: the generic `transportRouterSimple` library builds against any project
regardless of how many Ethernet/DoIP buses it defines. The set of "2-byte diag
address" buses is integration-specific and must not be hard-coded in the shared
library.
Actual: the shared library hard-codes `ETH_0`/`ETH_1`, so any integration that does
not define `ETH_1` (or builds without Ethernet) fails to compile.

**Environment**
[...]

Issue #349, opened 2026-01-16
ETH_0 needs to be defined even if PLATFORM_SUPPORT_ETHERNET is OFF


rolandreichweinbmw:
Since the addition of DoIP (commit 7018186f5a8cd12fd617643bab57192426bef72a), the build breaks if ETH_0 is not defined in case PLATFORM_SUPPORT_ETHERNET is OFF:

```
libs/bsw/transportRouterSimple/src/transport/routing/TransportRouterSimple.cpp:121:38: error: no member named 'ETH_0' in namespace 'busid'
  121 |     else if (sourceBusId == ::busid::ETH_0)
```

PR #513, opened 2026-07-09
Fix unguarded ETH_1 reference in transportRouterSimple

jiyuanwangbmw:
The generic transportRouterSimple hard coded busid::ETH_0 and busid::ETH_1 in is1ByteDiagAddressBus(), so integrations without ETH_1 failed to compile. This regressed [#349](https://github.com/eclipse-openbsw/openbsw/issues/349), which fixed the same for ETH_0.

Move the bus classification into the integration: add TransportConfiguration::is1ByteDiagAddressBus() and let the router
delegate to it. The generic router no longer names any Ethernet bus, so each integration decides which of its buses use 2-byte (DoIP) addresses.

Resolves:
[#507](https://github.com/eclipse-openbsw/openbsw/issues/507): transportRouterSimple references unguarded busid::ETH_1

### DIFF
--- a/libs/bsw/transportRouterSimple/src/transport/routing/TransportRouterSimple.cpp
+++ b/libs/bsw/transportRouterSimple/src/transport/routing/TransportRouterSimple.cpp
@@ -23,14 +23,6 @@ namespace transport
 using ::util::logger::Logger;
 using ::util::logger::TPROUTER;
 
-namespace
-{
-bool is1ByteDiagAddressBus(uint8_t const busId)
-{
-    return (busId != ::busid::ETH_0) && (busId != ::busid::ETH_1);
-}
-} // namespace
-
 TransportRouterSimple::TransportRouterSimple() : _transportLayers()
 {
     for (uint8_t i = 0U; i < NUM_BUFFERS; i++)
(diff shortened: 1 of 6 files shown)


rolandreichweinbmw (dismissed):

libs/bsw/transportRouterSimple/src/transport/routing/TransportRouterSimple.cpp
```diff
@@ -27,7 +27,19 @@ namespace
 {
 bool is1ByteDiagAddressBus(uint8_t const busId)
 {
-    return (busId != ::busid::ETH_0) && (busId != ::busid::ETH_1);
+    // A bus uses 2-byte (DoIP/Ethernet) diagnostic addresses only if the
```
rolandreichweinbmw: Instead of handling ETH_0 and ETH_1 specially, TransportRouterSimple needs to be moved to refApplication.

Then, ETH_1 should not be referenced elsewhere (making OpenBSW dependent on ETH_1).

jiyuanwangbmw: As an alternative to moving the complete router, TransportRouterSimple now stays generic and delegates the 1-byte/2-byte decision to TransportConfiguration::is1ByteDiagAddressBus(), so OpenBSW's generic code no longer references ETH_0/ETH_1, each integration provides its own policy.

rolandreichweinbmw (approved):

christian-schilling (approved):

PR #518, opened 2026-07-16
Fix STM32 CAN unit tests on 64-bit hosts

nhuvaoanh123:
Building the STM32 unit tests on a 64-bit Linux host (gcc 11, -Werror) surfaced three issues in the test code added with #415. Production driver code is untouched.

- The mock register masks were defined with `UL` literals, so `~mask` expressions widen to 64 bits on LP64 hosts and fail `-Werror=overflow`. The masks now use `U` literals, matching the 32-bit semantics of the target.
- Three FdCAN ISR tests created `etl::delegate` callbacks from temporary lambdas. The delegate stores a pointer to the functor, so the tests relied on a dangling reference (crashed under `-O2`). The lambdas are now named locals that outlive the delegate.
- `memset` on the fake FDCAN peripheral fails `-Werror=class-memaccess` because its volatile register operators make the struct non-trivially-assignable; the pointer is now cast to `void*`.

All 736 stm32 unit tests pass on Linux gcc-11 (Debug and Release, in the repo docker image) and on Windows MinGW.

This is a prerequisite for a follow-up PR that adds the `tests-stm32-*` presets to CI, which is how these were found.


rolandreichweinbmw (dismissed):

nhuvaoanh123:
Reopening to retrigger CI: the treefmt job failed on GitHub Actions docker-cache infrastructure errors (blob not found during image build) before the formatter ran. No code change needed.

rolandreichweinbmw:
There is an actual code change necessary for clang-format / treefmt:

```
-    auto callback         = ::etl::delegate<void()>::create(onFrameSent);
+    auto callback = ::etl::delegate<void()>::create(onFrameSent);
```

nhuvaoanh123:
Fixed and force-pushed, thanks.

rolandreichweinbmw (approved):

PR #519, opened 2026-07-17
Fix removed SystemTimer calls in STM32 ETL clock implementation

nhuvaoanh123:
### Problem

`platforms/stm32/etlImpl/src/clocks.cpp` no longer compiles on `main`: commit 9d3e89a28 ("Add TimestampProvider") renamed `getSystemTimeUs()`/`getSystemTimeMs()` to `getSystemTimeUs64Bit()`/`getSystemTimeMs64Bit()`, and the STM32 ETL clock implementation (merged via #414, written before the rename) still references the old names:

```
clocks.cpp:23: error: 'getSystemTimeUs' was not declared in this scope
clocks.cpp:28: error: 'getSystemTimeMs' was not declared in this scope
```

No CI configuration cross-compiles `platforms/stm32/etlImpl` yet, so the breakage is silent on `main`.

### Fix

Use the renamed 64-bit accessors, matching `platforms/s32k1xx/etlImpl/src/clocks.cpp`.

Verified with `arm-none-eabi-g++ -std=c++17 -fsyntax-only -Ilibs/bsw/bsp/include -Ilibs/3rdparty/etl/include`: fails on `main`, passes with this change.

### DIFF
--- a/platforms/stm32/etlImpl/src/clocks.cpp
+++ b/platforms/stm32/etlImpl/src/clocks.cpp
@@ -20,11 +20,11 @@ etl::chrono::high_resolution_clock::rep etl_get_high_resolution_clock()
 
 etl::chrono::system_clock::rep etl_get_system_clock()
 {
-    return etl::chrono::system_clock::rep(static_cast<int64_t>(getSystemTimeUs()));
+    return etl::chrono::system_clock::rep(static_cast<int64_t>(getSystemTimeUs64Bit()));
 }
 
 etl::chrono::steady_clock::rep etl_get_steady_clock()
 {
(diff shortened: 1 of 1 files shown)


rolandreichweinbmw:
Is there already a PR that ensures the code is compiled successfully, for the stm32 targets in the CI checks?

nhuvaoanh123:
Not yet. A follow-up PR adding STM32 unit-test builds (tests-stm32-debug/-release) to the CI matrix is prepared and will follow once #518 is merged. Target cross-compile CI jobs need the board executables, so they will come with the board bring-up PRs later in the series.

rolandreichweinbmw (approved):

PR #531, opened 2026-07-27
middleware/rpc: replace FutureDispatcher with ProxyMethod

filipe-cuim:
Introduce ProxyMethod, ProxyMethodImpl, and ProxyFireAndForgetMethod as replacements for FutureDispatcher and FutureDispatcherBase. The new classes provide a cleaner API for proxy-side asynchronous method calls. Replace FutureDispatcher usage in ProxyAttributeBase and ProxyAttribute with the new ProxyMethod and ProxyFireAndForgetMethod classes. Remove methodId from MethodTraits (now a runtime parameter passed to callMethod) and adapt ProxyAttribute, ProxyAttributeBase, and the generated proxy code accordingly.

Key changes:
- Add MethodTraits to describe input/output types and timeout
- ProxyMethodImpl consolidates future pool management, request tracking, message sending, and timeout handling into a single non-templated base
- ProxyMethod provides type-safe callbacks with conditional Result type supporting both void and non-void input/output types
- ProxyFireAndForgetMethod handles fire-and-forget calls with and without input payload
- Method ID is passed as a runtime parameter to callMethod()
- Remove FutureDispatcher.h and FutureDispatcherBase.h


libs/bsw/middleware/include/middleware/core/Future.h
```diff
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2025 BMW AG
+ * Copyright (c) 2025, 2026 BMW AG
```
christophruethingbmw: Not sure what's correct here, do we really put multiple years in here or not just the latest year?

christophruethingbmw (dismissed):

libs/bsw/middleware/include/middleware/core/Future.h
filipe-cuim: Yeah was not sure as well. Maybe the latest year would be enough I guess. I can edit this.

christophruethingbmw:
@matthiaskessler will also have a look

lcastela (approved):

libs/bsw/middleware/include/middleware/rpc/ProxyMethod.h
```diff
@@ -0,0 +1,279 @@
...
+    if (future == nullptr)
+    {
+        return etl::unexpected<core::HRESULT>(etl::in_place, core::HRESULT::RequestPoolDepleted);
+    }
+
+    core::Message msg = proxy.generateMessageHeader(methodId, future->requestId);
+    auto const index  = etl::distance(_futuresArray.begin(), future);
+    _callbacks[index] = cbk;
+    core::HRESULT ret = core::MessagePayloadBuilder::getInstance().allocate(input, msg);
+    return Base::callMethod(proxy, msg, ret);
```
matthiaskessler, line 218: I find the handling of the future a bit weird (without having too much knowledge about the middleware internals). But it seems, that we obtain a future here, but then leave it to `Base::callMethod()` to clean up, in case sth. goes wrong with the allocation. It makes it kind of hard to me to track who really own the future.
We had similar issue in the `transport` module and came to the conclusion, that whoever allocates is also responsible for releasing a resource. Maybe this doesn't apply here, but it could be worth thinking about...

matthiaskessler (approved):

libs/bsw/middleware/include/middleware/rpc/ProxyMethod.h
filipe-cuim: I agree with you @matthiaskessler, whoever allocates a resource, should be responsible for releasing it if something goes wrong. Here the resource is actually not the future, but the message. The reason why this is passed to the base class is simply to avoid the work the template does. We try to follow an approach where we use templated code to guarantee type-safety and reduce written code that is generated. But at the same time we want to be careful with the amount of code present in a template to avoid growing the code size too much (we need to keep in mind that since this is middleware, the template instantiation, which is dependent on user types, might grow a lot).

filipe-cuim: Some of these things could actually be a bit more simplified if the allocation mechanism is done differently or if the callbacks don't pass the type by parameter (so that they could be generic void(void) callbacks), but this requires a bigger refactoring, which was also not the purpose of this PR.

filipe-cuim: If you agree with this I would say to merge this PR, and I will follow-up on this detail in a next one

libs/bsw/middleware/include/middleware/core/Future.h
danielfaleirosantos: From what I know, you put multiple years, yes. However, I think for contiguous years it's `2025-2026`, for non-contiguous years it's, e.g., `2024, 2026`. I can't remember where I saw this, though.

PR #532, opened 2026-07-28
Create proxy mock and adapt jinja files

filipe-cuim:
### Key changes

- Change ProxyAttribute and ProxyEvent class definitions from being an alias, to being a class that simply forwards work to an implementation object.
- Create mocks for Proxy classes.
- Create a different jinja template file for proxy class implementation (proxy.cpp files), that is used when the generator is called in mock mode.

### DIFF
--- a/libs/bsw/middleware/tools/cpp_generator/templates/jinja/mock/service_proxy_mock.cpp.jinja
+++ b/libs/bsw/middleware/tools/cpp_generator/templates/jinja/mock/service_proxy_mock.cpp.jinja
@@ -1,0 +1,210 @@
+/********************************************************************************
+ * Copyright (c) 2026 BMW AG
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+/**
+  *
+  *  __  __ _     _     _ _
+  * |  \/  (_) __| | __| | | _____      ____ _ _ __ ___
+  * | |\/| | |/ _` |/ _` | |/ _ \ \ /\ / / _` | '__/ _ \
+  * | |  | | | (_| | (_| | |  __/\ V  V / (_| | | |  __/
+  * |_|  |_|_|\__,_|\__,_|_|\___| \_/\_/ \__,_|_|  \___|
+  *
+  * WARNING!
+  * This file is generated. Do not edit manually.
+  *
+  */
+#include "{{ service.headers.proxy }}"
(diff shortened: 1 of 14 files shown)


libs/bsw/middleware/tools/cpp_generator/templates/jinja/mock/service_proxy_mock.cpp.jinja
```diff
@@ -0,0 +1,210 @@
...
+
+::middleware::core::HRESULT {{ attribute.name }}Attribute::cancelGetterRequest(uint16_t requestId)
+{
+    auto* mock = {{ service.name }}ProxyMock::getMock();
+    return mock->{{ attribute.variable }}.cancelGetterRequest(requestId);
+}
+{%- if set_as_method %}
+
+etl::expected<uint16_t, ::middleware::core::HRESULT> {{ attribute.name }}Attribute::set(const AttributeType& payload,
+                                                                                       const SetterCallback& cbk)
```
simon-d-bmw: **Residual signature mismatch: `cbk` by reference here vs. by value in the shared header.**

`service_proxy.h.jinja` declares this member as:
```cpp
::etl::expected<uint16_t, ::middleware::core::HRESULT> set(AttributeType const& payload, const SetterCallback cbk);
```
(`cbk` passed **by value**), but this out-of-line definition uses:
```cpp
etl::expected<uint16_t, ::middleware::core::HRESULT> {{ attribute.name }}Attribute::set(const AttributeType& payload,
                                                                                       const SetterCallback& cbk)
```
(`cbk` passed **by reference**).

Value-vs-reference is part of the function signature, so this definition won't match the declaration in `service_proxy.h.jinja` for any service with a `setAsMethod` attribute that has a setter — it'll fail to compile/link. Note the mock header's own `MOCK_METHOD` for `set()` was already correctly updated to by-value (`const SetterCallback cbk`), so it's specifically this `.cpp` forwarding definition that's out of sync. Should be `const SetterCallback cbk` here to match.

simon-d-bmw (commented):
Thanks for the update — re-checked the latest head and most of the earlier signature mismatch is fixed: `get()`'s value/reference mismatch is resolved, and the bogus `SetterArgumentType` reference is gone (now correctly uses `AttributeType`).

One residual issue remains on the setter's callback parameter for `setAsMethod` attributes (see inline comment). Value vs. reference is part of a function's signature, so an out-of-line definition with a mismatched parameter type won't match the corresponding class member declaration — this will still fail to compile for any service with a `setAsMethod` attribute that has a setter.

libs/bsw/middleware/tools/cpp_generator/templates/jinja/mock/service_proxy_mock.cpp.jinja
simon-d-bmw: Confirmed fixed in 142f3e7 — `cbk` is now passed by value here, matching the header's declaration. Thanks for the quick turnaround! 🎉

simon-d-bmw (approved):

PR #533, opened 2026-07-28
Fix static init order UB in generated cluster connection code

lcastela:
Replace typed_storage<T> + create() with direct globals constructed via the single-argument constructor, and update initializeXClusterConnection() to reset only the Config object (explicit destructor + placement-new).

The original UB was calling placement-new on the Connection object which holds a reference member (_configuration), violating [basic.life]/8. The fix eliminates placement-new on the Connection entirely:

- Config and Connection are direct globals; Config is constructed first (same TU, declaration order), Connection's reference binds to it at startup.

- initializeXClusterConnection() now only reconstructs the Config (destructor + placement-new) to clear registered transceivers. The Connection's reference stays valid because Config is at the same address.

- service_common.cpp.jinja: replace function-local statics with static constexpr arrays using & (valid since vtable is always set at startup by the constructor).

### DIFF
--- a/libs/bsw/middleware/include/middleware/core/TransceiverContainer.h
+++ b/libs/bsw/middleware/include/middleware/core/TransceiverContainer.h
@@ -91,9 +91,9 @@ struct TransceiverContainer
     TransceiverContainer& operator=(TransceiverContainer&&)      = delete;
 
     /** Pointer to the vector holding transceiver pointers. */
-    ::etl::ivector<TransceiverBase*>* const _container{};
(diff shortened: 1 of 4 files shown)


lcastela:
@matthiaskessler, have a look when possible. it is a follow-up from: https://github.com/eclipse-openbsw/openbsw/pull/525

lcastela:
`treefmt` code format is fixed by: https://github.com/eclipse-openbsw/openbsw/pull/534

simon-d-bmw (approved):

PR #535, opened 2026-07-29
Support void method responses in middleware response buffer and skeleton generator

andre-silva9975:
# Overview
This PR adds support for `void` method responses in middleware response handling and updates generated skeleton code accordingly.

# What Changed
- Added a `void`-specific `respond(...)` overload in `ResponseBuffer` while keeping the existing payload-based path for non-void responses.
- Updated service skeleton templates to generate `void` response buffer types and `respond(...)` signatures without payload parameters for methods that have no output parameter.
- Added unit tests that validate `void` response behavior for success, invalid response handles, and send failure scenarios.

# Changed Files
- [ResponseBuffer implementation](libs/bsw/middleware/include/middleware/core/ResponseBuffer.h)
- [ResponseBuffer unit tests](libs/bsw/middleware/test/src/core/ResponseBufferTest.cpp)
- [Service skeleton source template](libs/bsw/middleware/tools/cpp_generator/templates/jinja/service_skeleton.cpp.jinja)
- [Service skeleton header template](libs/bsw/middleware/tools/cpp_generator/templates/jinja/service_skeleton.h.jinja)

# Impact
- Enables methods without output payloads to send valid header-only responses.
- Keeps generated APIs aligned with both payload and `void` method signatures.
- Improves confidence via dedicated test coverage for `void` response flows.

### DIFF
--- a/libs/bsw/middleware/include/middleware/core/ResponseBuffer.h
+++ b/libs/bsw/middleware/include/middleware/core/ResponseBuffer.h
@@ -70,9 +70,7 @@ public:
      */
     template<typename T = ResponseType>
     ::etl::enable_if_t<!::etl::is_void<T>::value, HRESULT> respond(
-        SkeletonResponseInfo& response,
-        T const& result,
-        bool const handleResponseFailure = true)
+        SkeletonResponseInfo& response, T const& result, bool const handleResponseFailure = true)
     {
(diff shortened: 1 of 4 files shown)


christophruethingbmw (dismissed):

lcastela (approved):
👍

christophruethingbmw (approved):

PR #536, opened 2026-07-30
Add openbsw_async crate integrating Rust Async framework (Embassy) with C++ Async

M-Hassanin:
Add openbsw_async crate integrating Rust Async framework (Embassy) with C++ Async

Introduce a no_std library that lets a C++ Async task drive
an Embassy executor cooperatively, one poll at a time,
instead of Embassy owning the CPU.


christophruethingbmw:
@M-Hassanin , would it be possible to do the vendoring of the cargo libraries in a separate PR? Then it get's easier to review.

M-Hassanin:
If it helps, filtering it out (ignore the 3rdparty folder) from your diff view using the folder structure should make the rest of the PR much easier to review.

rolandreichweinbmw:
Separate commit instead of separate PR would be fine for me.

M-Hassanin:
Hi @rolandreichweinbmw @christophruethingbmw , gentle reminder on this PR, it's been open for a while. Could you take a look when you get a chance, or let me know if it's ready to merge?

executables/referenceApp/application/src/app/rust/RustRuntime.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <async/Async.h>
+#include <async/Config.h>
+#include <interrupts/suspendResumeAllInterrupts.h>
+
+#include <cstdint>
+#include <rust_runtime.h>
+
+namespace
+{
+constexpr uint8_t MAX_TASKS             = static_cast<uint8_t>(ASYNC_CONFIG_TASK_COUNT);
```
rolandreichweinbmw: Maybe `size_t`? Or is there a hard requirement to limit this to 256?

Same for task ids in the following I guess?

executables/referenceApp/rustHelloWorld/include/rust_runtime.h
```diff
@@ -8,12 +8,21 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 
+#ifndef RUST_RUNTIME_H
+#define RUST_RUNTIME_H
```
rolandreichweinbmw: `#pragma once`

rolandreichweinbmw (commented):
Can you list the added 3rdparty modules / crates in the root `NOTICE.md`?

Also, can you accompany them with `.riminfo` to make that content traceable to the original repos?

executables/referenceApp/application/src/app/rust/RustRuntime.cpp
M-Hassanin: I have changed MAX_TASKS to size_t.
For task ids I kept it as it is, because it maps to ::async::ContextType which is uint8_t.

executables/referenceApp/rustHelloWorld/include/rust_runtime.h
M-Hassanin: Done

libs/3rdparty/cargo-vendor/critical-section/.riminfo
```diff
@@ -0,0 +1,15 @@
+1f96bd48a38e828b6c8a142107f0d5071cf4262a
+
+RIM Info file. You're welcome to read but don't write it.
+Instead, use RIM commands to do the things you want to do.
+BEWARE: Any manual modification will invalidate the file!
+
+remote_url     : https://github.com/rust-embedded/critical-section
+revision_sha1  : ed0f84c5ca3697a2166ed76e656c958cd3915b00
+target_revision: ed0f84c5ca3697a2166ed76e656c958cd3915b00
+ignores        : **/*,.*,.*/**/*,**/.*,**/.*/**/*
```
rolandreichweinbmw, line 10: I understand that your approach is to at least provide the `remote_url` for reference, and leave the synchronization of the files to `cargo vendor`.

However, the major point of using RIM here is to check unmodified sources, which is not possible if you define `ignores:` for everything.

Isn't it possible to still track via RIM (i.e. without ignoring the files as above), even when `cargo vendor` actually copies (overwrites) files maybe?

M-Hassanin: I think yes, technically possible.

RIM could verify the real upstream source if we dropped cargo vendor and instead had RIM check out the raw git repo, with Cargo consuming it via a path dependency ([patch.crates-io] syn = { path = "..." }).RIM could hash the whole directory.

Two reasons I don't prefer that:

1- It fights how Rust actually verifies dependencies. cargo vendor is paired with Cargo.lock, which cryptographically pins every dependency to a specific published version's checksum, which checked automatically on every build, with any change showing up as a normal diff.

2- Rim makes maintenance harder. Bumping a version stops being edit Cargo.toml, run cargo update and instead needs a rim sync per affected crate too. It also makes dependencies opaque to standard tooling that relies on Cargo.lock + crates.io metadata liek cargo audit and cargo tree.

So I think not using RIM at all for Rust dependencies is better, what do you think?

rolandreichweinbmw: Thanks for your explanation! I agree that RIM somehow conflicts with cargo's implementation with the similar feature.

Since RIM is currently an integral concept of OpenBSW's vendoring tooling, and "cargo vendor" offers similar functionality, how about adding "cargo vendor" for the Rust crates in OpenBSW? @matthiaskessler @christophruethingbmw ?

I expect extensive use of this for `libs/3rdparty/cargo-vendor` so we should decide wisely.

Can you add a respective section to '.github/workflows/rim.yml' to cover the crates part in that sense?

I came up with the following for checking this via `cargo vendor`, but maybe you have a different call in mind?

```
cargo vendor /tmp/cargo-vendor-check/cargo-vendor --locked --manifest-path Cargo.toml
diff -rq libs/3rdparty/cargo-vendor /tmp/cargo-vendor-check/cargo-vendor
```

M-Hassanin: Thanks for the suggestion, Added a cargo-vendor job alongside rim in that workflow.
One note: I renamed rim.yml -> third-party-libs-check.yml since it now covers both RIM and cargo-vendor checks, not just RIM.

M-Hassanin:
@rolandreichweinbmw All set on my end! Let me know if this is good to hit merge.

rolandreichweinbmw (approved):

rolandreichweinbmw:
> @rolandreichweinbmw All set on my end! Let me know if this is good to hit merge.

Thank you! Looks good!

PR #537, opened 2026-07-30
Fix common depending on configuration in CMake targets

matthiaskessler:
Fix common depending on configuration in CMake targets

"common" is a foundational, header-only interface used by dozens of
libraries, while "configuration" is executable-specific. "common"
had ended up linking "configuration" directly, inverting the
intended dependency direction and forcing every consumer of
"common" to transitively depend on an application's configuration.

- Remove the leftover duplicate "commonImpl" target from
  executables/referenceApp/configuration/CMakeLists.txt; its source
  file was already compiled into "configuration" and nothing linked
  against it.
- Restore libs/bsw/common as a pure INTERFACE library (include dirs,
  etl, platform) with no reference to "configuration".
- Have "configuration" (referenceApp and unitTest variants) link
  PUBLIC "common", the correct direction.
- common::busid::BusIdTraits::getName() is declared in "common" but
  implemented in each executable's "configuration" library.
  libs/bsw/transport and platforms/s32k1xx/bsp/canflex2Transceiver
  call it directly, so link "configuration" PUBLIC from those
  targets themselves, matching the existing precedent in
  libs/bsw/uds. Reorder link lists so "configuration" appears after
  "transport", since static libraries are resolved left to right.

Add a "module + moduleImpl" section to
doc/dev/guidelines/module.rst describing this pattern: keep the
generic interface target free of implementation-specific
dependencies, and put those in a separate impl target instead.

### DIFF
--- a/doc/dev/guidelines/module.rst
+++ b/doc/dev/guidelines/module.rst
@@ -157,6 +157,50 @@ Use only standard CMake commands. The typical structure is as follows:
         target_include_directories(<module>Mock PUBLIC include)
         target_link_libraries(<module>Mock PUBLIC gmock ... PRIVATE ...)
 
+.. _module_impl_pattern:
+
+The module + moduleImpl pattern
+-------------------------------
+
+Some modules only declare an interface (pure header/``INTERFACE`` CMake target) that is meant to be
+reused by many other modules, while a concrete implementation of that interface is only available,
+or only makes sense, in a more specific context (e.g. it depends on platform- or application-specific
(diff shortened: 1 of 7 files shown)


christian-schilling (dismissed):

rolandreichweinbmw (dismissed):

rolandreichweinbmw (commented):
LGTM - just rebase to resolve the documentation merge conflict.

rolandreichweinbmw (approved):

PR #542, opened 2026-08-04
Fix IPv4 comparison when IPv6 disabled.

andreu-montiel-te:
Hello,

We had a problem at Technica related to this; IP comparison doesn't look right when IPV6 is disabled.
I think this should be the fix, please let me know if I missed something.

Thanks,
Andreu

### DIFF
--- a/libs/bsw/cpp2ethernet/include/ip/IPAddress.h
+++ b/libs/bsw/cpp2ethernet/include/ip/IPAddress.h
@@ -392,7 +392,7 @@ inline bool operator==(IPAddress const& ip1, IPAddress const& ip2)
         && (ip1.be_uint32_at(1) == ip2.be_uint32_at(1))
         && (ip1.be_uint32_at(0) == ip2.be_uint32_at(0)));
 #else
-    return (ip1.raw[0] == ip2.raw[0]);
+    return (ip1.be_uint32_at(0) == ip2.be_uint32_at(0));
 #endif
 }
 
--- a/libs/bsw/cpp2ethernet/test/src/ip/IPAddressTest.cpp
+++ b/libs/bsw/cpp2ethernet/test/src/ip/IPAddressTest.cpp
@@ -272,6 +272,9 @@ TEST(IPAddressTest, equality)
 
     IPAddress ip3 = make_ip4(++addr);
     EXPECT_NE(ip1, ip3);
+
+    IPAddress ip4 = make_ip4(0xABCDFF01);
+    EXPECT_NE(ip1, ip4);
 }
 
 #ifndef OPENBSW_NO_IPV6


rolandreichweinbmw:
Good catch!

Can you please also add a test to IPAddressTest.cpp since TEST(IPAddressTest, equality) does not cover the original issue that you found.

andreu-montiel-te:
I tried but I suspect the affected fragment is not covered by any unit test.

I'm not very familiar with this environment, so I did the experiment of adding a code fragment to IPAddress.h and building all tests, this is the fragment:

```
#ifdef OPENBSW_NO_IPV6
# error "COMPILED WITH OPENBSW_NO_IPV6"
#endif
```

I expected a compilation error somewhere, but every test compiled just fine.  Also tried doing configuration step with -DOPENBSW_NO_IPV6=1, still builds.

Not sure if I missed something .

rolandreichweinbmw:
This is what I would expect (i.e. no compilation error). Please note that `-DOPENBSW_NO_IPV6=1` can't be put on the cmake command line directly but needs to be defined via cmake means.

Anyway, what I meant for unit testing is this:

```
diff --git a/libs/bsw/cpp2ethernet/test/src/ip/IPAddressTest.cpp b/libs/bsw/cpp2ethernet/test/src/ip/IPAddressTest.cpp
index b7f45dc4a..8c2e14121 100644
--- a/libs/bsw/cpp2ethernet/test/src/ip/IPAddressTest.cpp
+++ b/libs/bsw/cpp2ethernet/test/src/ip/IPAddressTest.cpp
@@ -272,6 +272,9 @@ TEST(IPAddressTest, equality)

     IPAddress ip3 = make_ip4(++addr);
     EXPECT_NE(ip1, ip3);
+
+    IPAddress ip4 = make_ip4(0xABCDFF01);
+    EXPECT_NE(ip1, ip4);
 }

 #ifndef OPENBSW_NO_IPV6
```

Feel free to just integrate it into your PR.

andreu-montiel-te:
Ok, added. Just bear in mind that this test wouldn't have covered the bug, because the problematic fragment is not compiled in any UT (#ifdef OPENBSW_NO_IPV6...) .

Andreu

rolandreichweinbmw:
You are right.

Anyway - thanks for adding the test! I can approve this PR now.

This issue showed that the OPENBSW_NO_IPV6 path was not well tested.

Just for the record: You can activate this path with `cmake ... -DCMAKE_CXX_FLAGS="-DOPENBSW_NO_IPV6=1"` - this uncovers several other bugs which I will address in a separate change.

rolandreichweinbmw (approved):

rolandreichweinbmw:
Content-wise and from the CI checks, the PR looks fine now.

Can you please adjust the commit messages according to the guidelines at https://github.com/eclipse-openbsw/openbsw/blob/main/doc/dev/guidelines/commit_message.rst ?

rolandreichweinbmw:
Thanks!

I'm handling the follow-up here: #544

PR #550, opened 2026-08-12
Try to retrieve the stack usage from filling on ThreadX

Wang-Perry:
we still have chance to retrieve the stack usage if stack filling is enabled, even stack checking is disabled.

### DIFF
--- a/libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
+++ b/libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
@@ -289,8 +289,67 @@ bool ThreadXAdapter<Binding>::getStackUsage(size_t const taskIdx, StackUsage& st
                 - reinterpret_cast<UCHAR*>(taskHandle.tx_thread_stack_highest_ptr));
         }
 #else
+#ifndef TX_DISABLE_STACK_FILLING
+        // we still have chance to retrieve the stack usage if stack filling is enabled, even stack
+        // checking is disabled
+        ULONG const* const stackStartPtr
+            = reinterpret_cast<ULONG const*>(taskHandle.tx_thread_stack_start);
+        ULONG const* const stackEndPtr
+            = reinterpret_cast<ULONG const*>(taskHandle.tx_thread_stack_end);
+        ULONG const* const stackCurrentPtr
+            = reinterpret_cast<ULONG const*>(taskHandle.tx_thread_stack_ptr);
+        ULONG const* const stackHighestPtr
+            = reinterpret_cast<ULONG const* const>(taskHandle.tx_thread_stack_highest_ptr);
+        ULONG const stackFillWord = static_cast<ULONG>(TX_STACK_FILL);
+
+        // Downward stack growth
+        if (stackEndPtr >= stackStartPtr)
+        {
+            bool const stackHighestPtrValid
+                = (stackHighestPtr >= stackStartPtr && stackHighestPtr <= stackEndPtr);
+            ULONG const* current = stackHighestPtrValid ? stackHighestPtr : stackCurrentPtr;
+
+            while (current > stackStartPtr && *current != stackFillWord)
+            {
+                --current;
+            }
+            // current points to the last filled word(from stackStart)
+            // move it back to the last used word in the stack, which is the next word from the last
+            // filled word
+            ++current;
+            stackUsage._usedSize = static_cast<uint32_t>(
+                                       reinterpret_cast<UCHAR const*>(stackEndPtr)
+                                       - reinterpret_cast<UCHAR const*>(current))
+                                   + 1; // stackEndPtr is the last valid stack address
+
+            // tx_thread_stack_highest_ptr points to the last used stack word
+            taskHandle.tx_thread_stack_highest_ptr
+                = const_cast<VOID*>(reinterpret_cast<VOID const*>(current));
+        }
+        else // Upward stack growth
+        {
+            bool const stackHighestPtrValid
+                = (stackHighestPtr <= stackStartPtr && stackHighestPtr >= stackEndPtr);
+            ULONG const* current = stackHighestPtrValid ? stackHighestPtr : stackCurrentPtr;
+
+            while (current < stackStartPtr && *current != stackFillWord)
+            {
+                ++current;
+            }
+            --current;
+            stackUsage._usedSize = static_cast<uint32_t>(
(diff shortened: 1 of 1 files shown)


rolandreichweinbmw (approved):

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
```diff
@@ -288,9 +288,68 @@ bool ThreadXAdapter<Binding>::getStackUsage(size_t const taskIdx, StackUsage& st
...
+        ULONG const stackFillWord = static_cast<ULONG>(TX_STACK_FILL);
+
+        // Downward stack growth
+        if (stackEndPtr >= stackStartPtr)
+        {
+            bool const stackHighestPtrValid
+                = (stackHighestPtr >= stackStartPtr && stackHighestPtr <= stackEndPtr);
+            ULONG const* current = stackHighestPtrValid ? stackHighestPtr : stackCurrentPtr;
+
+            while (current > stackStartPtr && *current != stackFillWord)
```
rolandreichweinbmw, line 312: Wouldn't it make more sense to flip the fill-word logic? Searching over fill-words until it is _not_ found anymore?

christophruethingbmw: Yes, that's also my feeling, because the valid data in the stack can be the fill word. I think what Roland proposes is the better heuristic.

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
```diff
@@ -288,9 +288,68 @@ bool ThreadXAdapter<Binding>::getStackUsage(size_t const taskIdx, StackUsage& st
...
+        ULONG const* const stackEndPtr
+            = reinterpret_cast<ULONG const*>(taskHandle.tx_thread_stack_end);
+        ULONG const* const stackCurrentPtr
+            = reinterpret_cast<ULONG const*>(taskHandle.tx_thread_stack_ptr);
+        ULONG const* const stackHighestPtr
+            = reinterpret_cast<ULONG const* const>(taskHandle.tx_thread_stack_highest_ptr);
+        ULONG const stackFillWord = static_cast<ULONG>(TX_STACK_FILL);
+
+        // Downward stack growth
+        if (stackEndPtr >= stackStartPtr)
```
matthiaskessler, line 306: How does this work? I mean is it an OS configuration option which direction the stack grows?
If so, couldn't we use that parameter here? I am very cautious because we are relying with a decision here on values, that in an error case could theoretically be overwritten and corrupted. This could also lead to very long loops later in the code.

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
matthiaskessler, line 1: Maybe we should rephrase the commit message because its suggesting sth. non deterministic and that might be worrying in OS related code. Either we have a solution or not :-)

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
Wang-Perry: Thanks for the feedback. 

We agree that searching from the beginning of the stack for the first non-fill word is more robust, since valid stack data can theoretically contain the fill word.

We actually considered this approach initially. However, this would require scanning the unused part of every thread's stack on each check, which can become expensive when the stacks are relatively large.

The current implementation starts from the previously detected stack usage and searches towards the beginning of the stack. The assumption is that stack usage normally grows incrementally, so this significantly reduces the amount of memory that needs to be scanned on each check.

This does mean that a valid stack value matching the fill pattern could cause the usage to be underestimated. This can be mitigated by using a distinctive fill pattern; ThreadX uses 0xEFEFEFEF by default.

So the current approach is a deliberate trade-off between accuracy and runtime overhead for periodic stack-usage monitoring. We can also document this limitation in the code.

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
Wang-Perry: Thanks for the feedback.

We agree that invalid stackStartPtr or stackEndPtr values could result in an invalid search range and potentially an unbounded loop.

For ThreadX, there is no equivalent of FreeRTOS's portSTACK_GROWTH macro that we can use here. ThreadX's default stack-checking logic assumes downward-growing stacks, while the stack-checking implementation can be overridden by individual ports when a different stack direction is required.

Also, the existing OpenBSW stack-filling implementation already relies on the same stackStartPtr and stackEndPtr information to determine the stack range. This PR uses the same mechanism.

Therefore, we consider validating or defining the stack boundaries to be outside the scope of this PR. The current change focuses on measuring the stack usage based on the existing stack boundaries.

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
Wang-Perry: OK. I agree that the current word "try" can be misleading. How about rephrase the commit message to "Improve stack usage measurement"?

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
```diff
@@ -288,9 +288,68 @@ bool ThreadXAdapter<Binding>::getStackUsage(size_t const taskIdx, StackUsage& st
...
+            stackUsage._usedSize = static_cast<uint32_t>(
+                                       reinterpret_cast<UCHAR const*>(stackEndPtr)
+                                       - reinterpret_cast<UCHAR const*>(current))
+                                   + 1; // stackEndPtr is the last valid stack address
+
+            // tx_thread_stack_highest_ptr points to the last used stack word
+            taskHandle.tx_thread_stack_highest_ptr
+                = const_cast<VOID*>(reinterpret_cast<VOID const*>(current));
+        }
+        else // Upward stack growth
```
rolandreichweinbmw, line 329: I think this "upward stack growth" branch is dead code, since in ThreadX, tx_thread_stack_end = stack_start + size - 1. (With size being unsigned.) Independent of the growth direction.

Same is true for the above pre-existing branch, also a bug.

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
rolandreichweinbmw: Agree.

Please update the PR accordingly. As far as I can see, there was no activity from your side on this PR for a month.

PR #556, opened 2026-08-17
Refactor netif api and state machine

christian-schilling:
Refactor netif api and state machine


executables/referenceApp/application/src/systems/UdsSystem.cpp
```diff
@@ -13,6 +12,7 @@
 
 #include <busid/BusId.h>
 #include <etl/array.h>
+#include <etl/span.h>
```
rolandreichweinbmw: Why? I can't find span declarations in this file.

rolandreichweinbmw (approved):

PR #558, opened 2026-08-18
Add middleware bazel targets

lcastela:
## Summary

Add Bazel build support for the OpenBSW middleware stack, including middleware libraries, unit tests, simulation, and reference-app middleware configuration.

## What Changed

- Added explicit Bazel targets for middleware core, memory, queue, and interface libraries.
- Added Bazel targets for middleware unit tests and their mock implementations.
- Consolidated middleware unit tests on the shared `executables/unitTest/etl_profile`.
- Extended the unit-test ETL profile with middleware test settings:
  - `ETL_USE_TYPE_TRAITS_BUILTINS`
  - `ETL_CHECK_PUSH_POP`
  - `ETL_LOG_ERRORS`
  - `ETL_VERBOSE_ERRORS`
  - `ETL_DEBUG`
- Relocated middleware test sources and mocks to match the project layout.
- Updated Bazel support for middleware simulation.
- Updated Bazel support for reference-app middleware configuration.
- Replaced system include paths with `strip_include_prefix` for middleware-owned targets to preserve compiler warning analysis.

depends-on: https://github.com/eclipse-openbsw/openbsw/pull/559


DominikAFischer:
Hi @lcastela  and thank you for the contribution. Please note that there's a systematic effort underway to migrate OpenBSW to Bazel. Please make sure to include @christophruethingbmw and myself for review so that we can make sure to keep parallel work in check :)

From a first glance this looks great!
One thing that I would suggest is to split the PR up into smaller logical chunks if possible. this would make the review a lot easier. Ideas for chunks:
- Python toolchain and infrastructure
- Codegen Bazel implementation
- Middleware changes (and ReferenceApp integration?)

lcastela:
Hi @DominikAFischer. I've noticed your pull requests with the bazel changes and was already planning to ask for your and @christophruethingbmw review once the changes are finalized. Still need to addressed some issues before that.

lcastela:
@DominikAFischer , @christophruethingbmw , ready! you can now have a look.

rolandreichweinbmw (dismissed):

DominikAFischer:
I will have a look later today. @christophruethingbmw is on leave and will only be back 2nd of September it seems.

libs/bsw/middleware/tools/codegen.bzl
```diff
@@ -0,0 +1,98 @@
...
+        arguments = [
+            "--output",
+            output_directory,
+            "--deployment-yaml",
+            ctx.file.deployment.path,
+            "--clang-format-config",
+            ctx.file.clang_format_config.path,
+        ],
+        tools = [ctx.executable.tool],
+        inputs = depset([ctx.file.deployment, ctx.file.clang_format_config]),
```
DominikAFischer, line 31: The `clang-format` binary is called in this rule (via `jinja2cpp.py`) but is not a declared Bazel input.
This means the rule will try to use a host installed `clang-format` instance, which makes it nonreproducible, cache-unsafe, and it won't work in strict sandbox settings (e.g. RBE). 

**Question:** Is formatting really needed for the generated files? The script should generate syntactically correct files that only ever live inside the `bazel-bin` folder, are never checked in and probably not read by humans.

If formatting is really required, make clang-format a hermetic tool input or properly set it up so it uses `clang-format `instance (`clang-format-17`) that ships with the docker container
If not, maybe add a `--no-format` switch or similar to the script

libs/bsw/middleware/tools/cpp_generator/jinja2cpp.py
DominikAFischer, line 400: This will try to use the `clang-format` host instance. In the docker container currently `clang-format-17` is available but no unversioned `clang-format` symlink is set up. An easy fix would be to setup a symlink in the dockerfile.
This probably means in its current state, formatting is skipped with a warning in all cases anyhow.

MODULE.bazel
```diff
@@ -16,6 +16,16 @@ module(
 bazel_dep(name = "platforms", version = "1.0.0")
 bazel_dep(name = "rules_cc", version = "0.2.17")
 bazel_dep(name = "bazel_skylib", version = "1.9.0")
+bazel_dep(name = "rules_python", version = "0.40.0")
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
```
DominikAFischer, line 21: This will pull a standalone python instance as an online build-time dependency.
I was given to understand that online dependencies should be avoided.

@christian-schilling would this be acceptable here? Essentially, we're already doing the same thing inside the dockerfile during image building.

If not, we could try to reuse the container python instance via [local_runtime_repo](https://github.com/bazel-contrib/rules_python/blob/main/docs/toolchains.md#local-toolchain), although I'm not sure if we would still need to download the pip packages via Bazel then or if we could reuse them too from the host.

The other solution would be to sidestep `rules_python` entirely and use the docker pyhon instance and packages via a plain `ctx.actions.run`. (nonhermetic, will break in strict sandbox / with remote execution)

libs/bsw/middleware/tools/pip_middleware_tools.MODULE.bazel
```diff
@@ -0,0 +1,17 @@
...
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
+pip.parse(
```
DominikAFischer, line 12: Downloads wheels at fetch time as an online dependency.
@christian-schilling is this acceptable?

libs/bsw/middleware/tools/cpp_generator/requirements.lock
DominikAFischer, line 1: Not actually introduced by this PR but still relevant:
Should these python lockfiles not also be pinned with hashes? (via `pip-compile --generate-hashes`)
@christian-schilling FYI

libs/bsw/middleware/interfaces/BUILD.bazel
DominikAFischer, line 1: Each `*_impl_default` alias maps both `unit_test` and `reference_app` to the same `*_mock_impl` located in the module's test subfolder.
Problems:
-  production level packages should not depend on test sources
- `gtest` dep leaks into the reference app build

Instead, suitable versions need to be supplied for the referenceApp build: stubs or empty libraries

executables/referenceApp/middlewareConfiguration/generated_outputs.bzl
DominikAFischer, line 1: This list needs to be hand maintained and could drift silently from the` deployment.yaml`.
We can prevent the drift by using his sort of staleness check:

- Add a `write_file` rule that can be used to generate a `generated_outputs_expected.txt` from `generated_outputs.bzl`
- Add a `--list-output-files` mode on `jinja2cpp.py` that can be used to create a `generated_outputs_actual.txt`
- Add a `diff_test` target that will compare the files and prevent merging inconsistent state (`bazel test //...` is run in CI)

libs/bsw/middleware/simulation/generated_outputs.bzl
DominikAFischer, line 1: This list needs to be hand maintained and could drift silently from the` deployment.yaml`.
We can prevent the drift by using his sort of staleness check:

- Add a `write_file` rule that can be used to generate a `generated_outputs_expected.txt` from `generated_outputs.bzl`
- Add a `--list-output-files` mode on `jinja2cpp.py` that can be used to create a `generated_outputs_actual.txt`
- Add a `diff_test` target that will compare the files and prevent merging inconsistent state (`bazel test //...` is run in CI)

libs/bsw/middleware/BUILD.bazel
DominikAFischer, line 1: Every migrated `libs/bsw` module up until has only a single BUILD file in the module root and no subpackages / nested BUILD files (assumption: module = "atomic" package unit).
I'm sure there's valid exceptions from this convention, but I would argue that there should not be any subpackages defined in any of the standard module folders:

I would argue, that there should not be any subpackages / BUILD files created inside any of the "standard" module subfolders:
- include
- doc
- test
- src
- mock

That said, I understand that middleware is a bigger sized module and BUILD file bloat can be a problem.
Since the tools subfolder is logically separate from the module, I think it makes sense to keep it as a separated package.

MODULE.bazel
DominikAFischer: Every migrated `libs/bsw` module up until has only a single BUILD file in the module root and no subpackages / nested BUILD files (assumption: module = "atomic" package unit).
I'm sure there's valid exceptions from this convention, but I would argue that there should not be any subpackages defined in any of the standard module folders:

I would argue, that there should not be any subpackages / BUILD files created inside any of the "standard" module subfolders:
- include
- doc
- test
- src
- mock

If there are no specific reasons why interfaces needs to be a dedicated package, I would move the target definitions to the module root BUILD file.
Since the tools subfolder is logically separate from the module, I think it makes sense to keep it as a separated package.

libs/bsw/middleware/tools/cpp_generator/BUILD.bazel
DominikAFischer, line 1: This BUILD file should be moved to the tools root package (one level up) instead in my opinion.

libs/bsw/middleware/tools/pip_middleware_tools.MODULE.bazel
DominikAFischer, line 1: If there's no specific reason why segmentation is needed here, would advocate placing this in the OpenBSW root MODULE. Segmented MODULE files are hard to read and harder to maintain and add indirection.

If segmenting the root MODULE is really desired, then it should not be moved her but to a more generic location near the repo root.

DominikAFischer (changes requested):
Looks good in general, but I've added some remarks for discussion

libs/bsw/middleware/tools/cpp_generator/requirements.lock
christian-schilling: Yes they should.

libs/bsw/middleware/tools/pip_middleware_tools.MODULE.bazel
christian-schilling: https://thehackernews.com/2026/08/rust-supply-chain-attack-puts-build.html

What else needs to happen to get the point across that absolutely no network access during a build is acceptable?

MODULE.bazel
christian-schilling: See comment below.
We should first setup our CI to build in containers with `--network=none` as a baseline and then figure out how to setup our build *within* that contraint.

lcastela:
I didn't forget about this, but I had to switch to more prio topics. will follow-up with this soon

libs/bsw/middleware/tools/codegen.bzl
christophruethingbmw: I agree, using `clang-format` of the host makes it less hermetic and less reproducible. I also think that formatting is not strictly necessary, in case the generated code is correct it should be fine? We should make the template to generate the files nice such that it generates readable code and does not require another formatting? And even if the generated files do not follow the formatting rules this would not be a problem?

rolandreichweinbmw: I don't consider generated files as source, since by definition it's a different class of files.

lcastela: Fixed.

libs/bsw/middleware/tools/cpp_generator/jinja2cpp.py
lcastela: Fixed. jinja2cpp.py now takes an explicit --clang-format-binary flag instead of resolving clang-format from $PATH.

MODULE.bazel
lcastela: No longer applicable, changes from latest main didn't need this.

libs/bsw/middleware/tools/pip_middleware_tools.MODULE.bazel
lcastela: No longer applicable, changes from latest main didn't need this.

libs/bsw/middleware/tools/cpp_generator/requirements.lock
lcastela: No longer applicable, changes from latest main didn't need this.

libs/bsw/middleware/interfaces/BUILD.bazel
lcastela: Fixed. The default label_flag values for concurrency_impl/logger_impl/os_impl/time_impl are now plain no-op stubs

executables/referenceApp/middlewareConfiguration/generated_outputs.bzl
lcastela: already addressed in the latest main, just add to adapt the pr

libs/bsw/middleware/tools/pip_middleware_tools.MODULE.bazel
lcastela: No longer applicable, changes from latest main didn't need this.

libs/bsw/middleware/tools/cpp_generator/BUILD.bazel
lcastela: kept it aligned with latest main

libs/bsw/middleware/simulation/generated_outputs.bzl
lcastela: already addressed in the latest main, just add to adapt the pr

lcastela:
@christophruethingbmw , @DominikAFischer , can you have another look?

Issue #562, opened 2026-08-21
Potential Buffer Over-read and incorrect comparison in IPAddress::operator==


krishna127702:
**Describe the bug**
The operator== implementation for IPv6 assumes that both ip1 and ip2 always have at least 4 blocks (128 bits) of data available.
Risk 1 (Memory Safety): If an IPv4 address object is passed to this operator while PLATFORM_SUPPORT_IPV6 is enabled, the code will attempt to access be_uint32_at(1), (2), and (3). If the underlying buffer is only 4 bytes long, this results in an out-of-bounds read, which can cause a crash or undefined behavior (security risk).
Risk 2 (Logic): The operator does not check if ip1 and ip2 belong to the same Family. An IPv4 address and an IPv4-mapped IPv6 address should ideally be treated as equal, but this implementation will fail to identify them as such.

**Steps to reproduce the bug**
Prerequisites:
PLATFORM_SUPPORT_IPV6 must be defined at compile time.
Compiler: GCC 11.4.0 or Clang 14+ (with -fsanitize=address,undefined recommended).

Reproducing Risk 1 — Uninitialized Byte Access (UB):
#define PLATFORM_SUPPORT_IPV6
#include "ip/IPAddress.h"
#include <cstdio>

int main()
{
    // Construct an IPAddress manually WITHOUT using make_ip4() or make_ip6().
    // This leaves bytes [0..11] uninitialized (or zero by accident in some compilers).
    ip::IPAddress addr1;
    // Only fill the last 4 bytes (as if it were a raw IPv4 buffer copy)
    addr1.raw[12] = 192;
    addr1.raw[13] = 168;
    addr1.raw[14] = 1;
    addr1.raw[15] = 1;
    // Upper bytes [0..11] are UNINITIALIZED

    ip::IPAddress addr2 = ip::make_ip4(192, 168, 1, 1); // Correctly initialized

    // operator== reads be_uint32_at(0), (1), (2), (3) on addr1,
    // accessing uninitialized memory — UNDEFINED BEHAVIOR
    bool result = (addr1 == addr2);
    printf("Equal: %d\n", result); // Result is unpredictable
    return 0;
}

**Expected behavior and actual behavior**
Expected: 
No runtime failure
Actual:
runtime error: load of value <uninitialized>, which is not a valid value for type 'uint32_t'

Expected:
Scenario												Expected Result
Two identical IPv4 addresses compared via operator==	TRUE
[...]

PR #569, opened 2026-08-26
Add SOME/IP implementation module cpp2someip

Kim-Woelfel:
Add SOME/IP implementation module cpp2someip

Add SOME/IP (Scalable Service Oriented Middleware
over IP) implementation covering most of the open
source specification (https://some-ip.com/standards.shtml).

It supports:
- SOME/IP Service Discovery (SD)
- SOME/IP RPC via UDP, TCP and TP
- Serialization/Deserialization
- Sending/Receiving notifications
- Reboot Tracking

Current limitations:
- No dynamic switching between unicast
  and multicast based on consumers
- SD option support limited to Ipv4
  and Ipv6 endpoint and multicast options


executables/referenceApp/application/include/someip/ProvidedServiceHandler.h
```diff
@@ -0,0 +1,101 @@
...
+
+template<typename CallbackPoolType, someip::port::type ServicePort>
+inline ProvidedServiceHandler<CallbackPoolType, ServicePort>::ProvidedServiceHandler(
+    CallbackPoolType& methodCallbacks, someip::IEventSender& es)
+: ServiceHandler(methodCallbacks), _request(), _response(), _eventSender(es)
+{}
+
+template<typename CallbackPoolType, someip::port::type ServicePort>
+someip::ISomeIpSerializable*
+ProvidedServiceHandler<CallbackPoolType, ServicePort>::getResponse(uint16_t methodId)
```
rolandreichweinbmw: You can use `[[maybe_unused]] uint16_t methodId` here (and in the following places), since we have C++17.

executables/referenceApp/application/include/systems/SomeIpSystem.h
```diff
@@ -0,0 +1,115 @@
...
+ * Copyright (c) 2025 Accenture
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#ifndef SYSTEMS_SOMEIP_SYSTEM_H
```
rolandreichweinbmw: `#pragma once`

executables/referenceApp/application/src/someip/ServiceListener.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+void ServiceListener::updateEventgroupDescription(::someip::ServiceDescription&, ServiceStatus) {}
+
+::someip::MethodDetail const* ServiceListener::getMethodDetail(uint16_t methodId) const
+{
+    switch (methodId)
+    {
+        case 1U:
```
rolandreichweinbmw: Maybe either mark this magic number with a comment line, or even better give it a symbolic name that speaks for itself?

Similar for 0xFFFF, below.

executables/referenceApp/application/src/systems/SomeIpSystem.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+#include "async/Types.h"
+#include "ethConfig.h"
+#include "ip/IPAddress.h"
+#include "someip/ServiceDescription.h"
+#include "someip/SomeIpConstants.h"
+
+#include <tuple>
+
+std::tuple<uint16_t, ::someip::PortRangeReturnCode>
```
rolandreichweinbmw: Please use `etl::` instead of `std::`.

libs/bsw/cpp2someip/include/someip/util/Delegate.h
```diff
@@ -0,0 +1,579 @@
...
+      Delegate<RT, P1, P2, P3, P4>,
+      typename MFunctionPtrT<RT, P1, P2, P3, P4>::Type>
+{
+public:
+    using tDelegate = Delegate<RT, P1, P2, P3, P4>;
+    using tBase     = DelegateBase<tDelegate, typename MFunctionPtrT<RT, P1, P2, P3, P4>::Type>;
+
+    explicit Delegate(void* callee = nullptr, typename tBase::tMemberFctPtr mfp = 0L);
+
+#if ESR_DEPRECATE < 202310
```
rolandreichweinbmw: ESR_DEPRECATE is defined nowhere, so this expression evaluates to true. So maybe we can remove this conditional?

libs/bsw/cpp2someip/include/someip/util/Delegate.h
```diff
@@ -0,0 +1,579 @@
...
+{}
+
+template<
+    typename RT,
+    typename P1 = void,
+    typename P2 = void,
+    typename P3 = void,
+    typename P4 = void,
+    typename P5 = void>
+class Delegate
```
rolandreichweinbmw, line 142: If this delegate implementation is not SOME/IP specific, etl::delegate, etl::inplace_function or similar could be used instead.

libs/bsw/cpp2someip/include/someip/Closure.h
```diff
@@ -0,0 +1,142 @@
...
+
+#include "someip/util/Delegate.h"
+
+namespace someip
+{
+/**
+ * Base class for all Callback implementations
+ */
+template<typename ParamType, typename ResultType = void>
+class Closure
```
rolandreichweinbmw, line 21: etl::closure ?

libs/bsw/cpp2someip/include/someip/endian_helpers.h
```diff
@@ -0,0 +1,93 @@
...
+#include <cstdint>
+
+namespace someip
+{
+namespace endian
+{
+
+// Read big-endian value from byte buffer
+template<typename T>
+inline T read_be(uint8_t const* ptr)
```
rolandreichweinbmw, line 25: You don't need to re-implement. There is already etl::unaligned_type<>.

E.g. useable as etl::be_uint32_t etc.

libs/bsw/cpp2someip/include/someip/EtlVector.h
```diff
@@ -0,0 +1,216 @@
...
+{
+/// \cond INTERNAL
+template<class T, size_t S>
+struct GetSizeHelper<::etl::vector<T, S>>
+{
+    static uint32_t getSize(::etl::vector<T, S> const& t, uint32_t const lengthSize)
+    {
+        uint32_t size = lengthSize;
+
+        for (uint32_t i = 0U; i < t.size(); ++i)
```
rolandreichweinbmw, line 33: `uint32_t` -> `size_t` ?

Similarly below.

libs/bsw/cpp2someip/include/someip/PrimitiveTypes.h
```diff
@@ -0,0 +1,294 @@
...
+{
+    static_assert(
+        sizeof(T) == sizeof(RT),
+        "floating point to/from integral conversion type not the same size");
+
+    // in C it is possible to use unions to convert from one type to another (type-punning)
+    // in C++ this behavior is compiler dependent and can therefore be undefined behavior
+    // as a solution memcpy can be used to convert from one type into another type
+    RT dest{};
+    std::memcpy(&dest, &value, sizeof(value));
```
rolandreichweinbmw, line 64: `etl::mem_copy`

libs/bsw/cpp2someip/src/QueryManager.cpp
```diff
@@ -0,0 +1,607 @@
...
+        query.timestamp       = timestamp;
+        query.repetitionCount = 1U;
+    }
+}
+
+void QueryManager::updateQueryInRepetitionPhase(uint64_t const timestamp, ServiceQuery& query) const
+{
+    uint32_t const timePassed = static_cast<uint32_t>(timestamp - query.timestamp);
+    uint32_t const repetitionCount
+        = std::min(query.repetitionCount, static_cast<uint32_t>(sizeof(uint32_t) * 8U));
```
rolandreichweinbmw: `etl::min`

libs/bsw/cpp2someip/src/RpcReceiver.cpp
```diff
@@ -0,0 +1,460 @@
...
+    {
+        return false; // Non-empty sorts after empty
+    }
+
+    // Lexicographic comparison for strict weak ordering
+    auto const& lhs_addr = lhs.value().getRemoteEndpoint().getAddress();
+    auto const& rhs_addr = rhs.value().getRemoteEndpoint().getAddress();
+
+#ifdef PLATFORM_SUPPORT_IPV6
+    // IPv4-mapped IPv6 addresses: compare bytes 12-15
```
rolandreichweinbmw: Maybe you meant `IPv6-mapped IPv4 addresses` ?

libs/bsw/cpp2someip/test/src/SdRemoteSizeofTest.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+#include <gtest/gtest.h>
+
+namespace
+{
+
+class SocketDummy : public ::tcp::AbstractSocket
+{
+public:
+    ::ip::IPAddress getRemoteIPAddress() const override { return ::ip::IPAddress(); }
```
rolandreichweinbmw: Unused?

Also, returns an uninitialized IPAddress?

libs/bsw/cpp2someip/test/src/SdSizeofTest.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace
+{
+/**
+ * Make sure sizeof SdEndpoint == 20U, ServiceQuery == 120U, ServiceDescription == 36U,
+ * SubscribedEventGroup == 16U, SubscriptionEndpoint == 40U and SubscriptionManager == 32U.
+ */
+TEST(SdSizeof, Sizeof)
+{
+    if (sizeof(size_t) == 8U)
```
rolandreichweinbmw: Maybe either support all bitwidths, or admit that on 32 bit (typical target HW) we can't ensure the Assertions? And therefore just leave out the test altogether?

libs/bsw/cpp2someip/src/RpcReceiver.cpp
```diff
@@ -0,0 +1,460 @@
...
+                bytesLeft);
+            Statistics::incCounter(Statistics::Counter::RPC_MALFORMED_MESSAGE_RX);
+            return;
+        }
+
+        Statistics::incCounter(Statistics::Counter::PDU_RX);
+
+        SomeIpMessage const message(input.subspan(offset, bytesLeft));
+
+        uint32_t const payloadLength = message.getPayloadLength();
```
rolandreichweinbmw, line 80: offset += (payloadLength + SomeIpMessage::OFFSET_PAYLOAD); happens before the validity check at RpcReceiver.cpp:91. payloadLength is attacker-controlled up to 0xFFFFFFF7 (getPayloadLength() in SomeIpMessage.cpp:217 only guards underflow, not the upper range). With payloadLength == 0xFFFFFFF0, payloadLength + 16 wraps to 0, so:

* the guard at line 91 (bytesLeft < 0) is false and does not fire, and
* offset never advances → the while (offset < length) loop spins forever.

Fix: validate payloadLength <= bytesLeft - OFFSET_PAYLOAD (no addition) before mutating offset, as SdReceiver.cpp:48 already does correctly.

libs/bsw/cpp2someip/src/SdServiceRegistry.cpp
```diff
@@ -0,0 +1,415 @@
...
+    _serviceTracker.init(*this);
+
+    async::scheduleAtFixedRate(
+        _ethernetContext,
+        _cyclicFunction,
+        _cyclicTimeout,
+        CYCLE_TIME_MS,
+        ::async::TimeUnit::MILLISECONDS);
+
+    printf("SdServiceRegistry::init() cyclic schedule done\n");
```
filipe-cuim: Is this printf intentional?

libs/bsw/cpp2someip/include/someip/PrimitiveTypes.h
filipe-cuim: If the goal here is to convert types, etl::mem_copy might not help here, because of the templated type-safety

rolandreichweinbmw: You are right @filipe-cuim !

A compromise would be to just use the prefix `std::` since that would be my primary goal: To keep `std::` out consistently out of production code.

libs/bsw/cpp2someip/src/SdServiceRegistry.cpp
Kim-Woelfel: No. I will remove it in the next update.

libs/bsw/cpp2someip/src/RpcReceiver.cpp
Kim-Woelfel: Good finding. I'd like it even more if such a check would be added at another location in the code, because actually 0xFFFFFFF7 is even for SOME/IP TP quite a big value and should be evaluated as invalid earlier in the pipeline. For not I will add your fix, but this should be investigated further in the future.

libs/bsw/cpp2someip/test/src/SdSizeofTest.cpp
Kim-Woelfel: You are right. I'll remove it.

libs/bsw/cpp2someip/test/src/SdRemoteSizeofTest.cpp
Kim-Woelfel: Actually not unused. It is used further down in that file as UdpSocketType and since it is just a dummy class I see nothing wrong in using the default IP.

libs/bsw/cpp2someip/src/RpcReceiver.cpp
Kim-Woelfel: You are right

libs/bsw/cpp2someip/src/QueryManager.cpp
Kim-Woelfel: Done

libs/bsw/cpp2someip/include/someip/util/Delegate.h
Kim-Woelfel: Done

libs/bsw/cpp2someip/include/someip/util/Delegate.h
Kim-Woelfel: Most likely yes. I will have a look at this in the future.

libs/bsw/cpp2someip/include/someip/Closure.h
Kim-Woelfel: I will also investigate this further in the future.

libs/bsw/cpp2someip/include/someip/endian_helpers.h
Kim-Woelfel: I will have a look at this.

libs/bsw/cpp2someip/include/someip/EtlVector.h
Kim-Woelfel: Can you go further into detail? Since both GetSizeHelper methods that are manipulating this member also return uint32_t I'd say it is ok to leave it like this.

libs/bsw/cpp2someip/test/src/SdRemoteSizeofTest.cpp
rolandreichweinbmw: The issue is, there is currently no default IPAddress, rather left uninitialized. See https://github.com/eclipse-openbsw/openbsw/issues/562

Maybe when the other PR is merged, then you can just do `return ::ip::IPAddress();`. Until then, you need make_ipv4().

libs/bsw/cpp2someip/include/someip/EtlVector.h
rolandreichweinbmw: Technically, it's the same. It's just about the convention that sizes and indexes should be `size_t`. Makes the code more portable, even smaller and faster on non-32-bit.

If it's a local convention in existing OpenBSW code that you don't want to refactor here, leave it.

executables/referenceApp/application/src/systems/SomeIpSystem.cpp
Kim-Woelfel: Good idea.

executables/referenceApp/application/include/someip/ProvidedServiceHandler.h
Kim-Woelfel: Good idea. Done.

executables/referenceApp/application/include/systems/SomeIpSystem.h
Kim-Woelfel: Done.

executables/referenceApp/application/src/someip/ServiceListener.cpp
Kim-Woelfel: Good idea. Done.

libs/bsw/cpp2someip/test/src/SdRemoteSizeofTest.cpp
Kim-Woelfel: I get what you mean. I changed this and also other occurrences in cpp2someip to make_ip4/ip6.

rolandreichweinbmw (dismissed):

rolandreichweinbmw:
There is a consistent test failure here:

```
481 - ServiceDescription.test_triviality_of_ServiceDescription (Failed)
```

Can you please have a look?

rolandreichweinbmw (approved):

PR #570, opened 2026-08-27
Add bazel support for libs/bsw mock targets

SuhashiniNaik:
Add bazel support for libs/bsw mock targets

//libs/3rdparty/freeRtos:freertos_mock
//libs/bsw/util:util_mock
//libs/bsw/asyncImpl:async_impl_mock
//libs/bsw/async:async_mock
//libs/bsw/async:async_mock_impl
//libs/bsw/bsp:bsp_mock
//libs/bsw/io:io_mock
//libs/bsw/logger:logger_mock
//libs/bsw/lifecycle:lifecycle_mock

### DIFF
--- a/libs/bsw/async/BUILD.bazel
+++ b/libs/bsw/async/BUILD.bazel
@@ -73,7 +73,6 @@ cc_library(
     strip_include_prefix = "include",
     visibility = ["//visibility:public"],
     deps = [
-        ":async_platform",
         "//libs/3rdparty/etl",
     ],
 )
@@ -84,5 +83,39 @@ cc_library(
     deps = [
         ":async_binding",
         ":async_headers",
+        ":async_platform",
     ],
+)
+
+cc_library(
+    name = "async_mock",
+    testonly = True,
+    hdrs = [
+        "mock/gmock/include/async/AsyncMock.h",
+        "mock/gmock/include/async/LockMock.h",
+        "mock/gmock/include/async/RunnableMock.h",
+        "mock/gmock/include/async/TestContext.h",
+        "mock/gmock/include/async/TimeoutMock.h",
+        "mock/gmock/include/async/Types.h",
+    ],
+    strip_include_prefix = "mock/gmock/include",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":async_headers",
+        "//libs/3rdparty/etl",
+        "//libs/3rdparty/googletest:gtest",
+        "//libs/bsw/platform",
+    ],
+)
+
+cc_library(
+    name = "async_mock_impl",
+    testonly = True,
+    srcs = [
+        "mock/gmock/src/async/AsyncMock.cpp",
+        "mock/gmock/src/async/TestContext.cpp",
+        "mock/gmock/src/async/Types.cpp",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [":async_mock"],
 )
--- a/libs/bsw/util/BUILD.bazel
+++ b/libs/bsw/util/BUILD.bazel
@@ -21,3 +21,29 @@ cc_library(
         "//libs/bsw/platform",
     ],
 )
+
+cc_library(
+    name = "util_mock",
+    testonly = True,
+    srcs = [
+        "mock/gmock/src/util/StdIoMock.cpp",
+        "mock/gmock/src/util/logger/TestConsoleLogger.cpp",
+    ],
+    hdrs = [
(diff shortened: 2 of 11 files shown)


libs/bsw/util/BUILD.bazel
```diff
@@ -14,10 +14,43 @@ cc_library(
     name = "util",
     srcs = glob(["src/util/**/*.cpp"]),
     hdrs = glob(["include/**/*.h"]),
+    defines = select(
+        {
+            "//bazel/config/executable_config:reference_app": [],
+            "//bazel/config/executable_config:unit_test": ["ESTD_HAS_EXCEPTIONS"],
+        },
+        no_match_error = "executable_config must be 'reference_app' or 'unit_test' to select ESTD_HAS_EXCEPTIONS.",
```
christian-schilling, line 22: What does this mean? The build will fail if not building refapp or UT? That does not sound like something we'd want.

SuhashiniNaik: Thank you for the review.
 executable_config which is a string_flag only allows "reference_app" or "unit_test" (enforced by the flag itself), so any other value fails before select() even runs:

`Error in fail: Error setting //bazel/config/executable_config:executable_config: invalid value 'some_made_up_value'. Allowed values are ["reference_app", "unit_test"]`

If we ever add a third config, we'd need to update the allowed values list first, which would force us to update this select() too , so can we keep `no_match_error` than switch to `//conditions:default`?

christian-schilling: And what about downstream users wanting integrate this into their build?

SuhashiniNaik: That's a good point and it makes sense, 
I think using //conditions:default would be safer here, with ESTD_HAS_EXCEPTIONS enabled for unit_test and disabled by default for other consumers.

downstream users integrating libs/bsw/util may have their own build configuration and may not use `//bazel/config/executable_config:executable_config`. In that case, `no_match_error `could make the library unnecessarily restrictive for downstream builds. I'll change it to  `//conditions:default `

SuhashiniNaik: Hi, I have removed this part `"ESTD_HAS_EXCEPTIONS"` from the Bazel build file entirely.
it does not have any defines or select() for "ESTD_HAS_EXCEPTIONS" . As it has been removed from CmakeLists.txt in PR #578.

christophruethingbmw (dismissed):

libs/bsw/async/BUILD.bazel
DominikAFischer: Move the `":async_platform"` dep from here to the `async` library

libs/bsw/async/BUILD.bazel
```diff
@@ -86,3 +86,50 @@ cc_library(
...
+        "mock/gmock/include/async/TimeoutMock.h",
+        "mock/gmock/include/async/Types.h",
+    ],
+    implementation_deps = [
+        "//libs/3rdparty/etl",
+        "//libs/3rdparty/googletest:gtest",
+        "//libs/bsw/async",
+        "//libs/bsw/platform",
+    ],
+    strip_include_prefix = "mock/gmock/include",
```
DominikAFischer: Completely remove: `hdrs`, `implementation_deps`, `strip_include_prefix`
Instead add:
-  `deps=[":async_mock"]`
- `visibility=public`

DominikAFischer (changes requested):
I think the split on the async mock target is slightly wrong, probably to work around a preexisting error (`async_platform` was attached as a dep to `async_headers` instead of `async`). This should be a cleaner solution while still keeping full Cmake parity

libs/bsw/async/BUILD.bazel
SuhashiniNaik: hello
I moved `async_platform` to `async` as suggested but `async_binding_headers` throws (StaticContextHook.h: No such file or directory) erro, since it was getting that header transitively via async_headers → async_platform.

Reason could be CMake's` asyncBinding` only links runtime, but `AsyncBinding.h` includes RTOS-specific headers. This works in CMake because INTERFACE includes propagate globally across the link graph, regardless of explicit deps. 
hence I have added  async_platform directly as a dep of `async_binding_headers` in `executables/referenceApp/asyncBinding/BUILD.bazel`

DominikAFischer: Sorry I overlooked that, makes sense

DominikAFischer (approved):

christophruethingbmw (approved):

Issue #511, opened 2026-07-06
Warning of build with preset of s31k148-freertos-gcc


nvzentech:
**Describe the bug*
Warnings like warning: _kill is not implemented and will always fail

**Steps to reproduce the bug**
Reproducing the issue is crucial for identifying and resolving the problem. Please include all
necessary code and detailed steps required to replicate the issue. 

**Expected behavior and actual behavior**
Whether warnings creates issue with freertos

**Environment**
Operating System: Ubuntu 22.04
Build System/Tools: GCC 11.4.0, CMake 3.31.12]

Please specify the exact environment or tools you’re using to help us better understand and
replicate the issue.
$ cmake --preset s32k148-freertos-gcc
$ cmake --build --preset s32k148-freertos-gcc

**Additional context**

$ cmake --build --preset s32k148-freertos-gcc
warning: _close is not implemented and will always fail
note: the message above does not take linker garbage collection into account
warning: _kill is not implemented and will always fail

ThePo1es:
Same warnings as #244, where Roland answered that the ELF is fine and that most of these are POSIX symbols the app doesn't use. That holds for four of the six. The other two are worth a note.

They all come from `.gnu.warning.*` sections in `libnosys.a`, which gets linked because of `-specs=nosys.specs` in `cmake/toolchains/ArmNoneEabi-gcc.cmake:26`. ld prints one at every reference site, before `--gc-sections` has run. That's what the `note:` in your log is about.

I built `main` (9b362b8) with arm-gnu-toolchain 14.3.Rel1, the release pinned in `docker/development/Dockerfile:7`, and looked at the map file and the ELF. Different toolchain and distro from yours; on mine ld emits six of these, including the two in your log.

`_close`, `_lseek`, `_read` and `_write` get pulled in by `crt0.o`, which references `exit` unconditionally. That drags in newlib's stdio cleanup (`libc_a-exit.o` -> `libc_a-findfp.o` -> `libc_a-stdio.o`) and from there the four file-op stubs. Unused, and all four get collected away again; `arm-none-eabi-nm app.referenceApp.elf` shows none of them.

`_kill` and `_getpid` are different. They arrive via `std::abort()`, which is what the ETL assert handler calls (`executables/referenceApp/application/src/main.cpp:35`), and they do survive into flash at 0x674 and 0x664. Nothing reaches them at runtime unless an assert fires, so as warnings they're still harmless.

[...]

PR #579, opened 2026-09-01
Fix ETL_ASSERT hang by resetting the ECU explicitly

### LINKED ISSUE #575: ETL_ASSERT on s32k148 hangs the task instead of resetting the ECU
**Describe the bug**

`doc/dev/guidelines/conventions/statements.rst:190-192` says:

> In contrast to a *PC* program where asserts are ignored in release builds, an `ETL_ASSERT` will be present in release code causing an ECU reset and most likely a customer noticeable event. We usually do not distinguish between debug and release builds.

On s32k148 there is no reset. The asserting task spins in newlib's `_exit` forever, nothing is logged, and because FreeRTOS is preemptive it starves every task below its own priority, including idle.

I have not run this on hardware. Everything below is from the disassembly of the built ELF and from the FreeRTOS configuration, so please treat the runtime part as derived rather than observed.

**Steps to reproduce the bug**

```
cmake --preset s32k148-freertos-gcc
cmake --build --preset s32k148-freertos-gcc
```

Flash it, then run `lc assert` on the console. That is the documented "forces an assert" command (`doc/dev/learning/console/lcConsole.rst:26`, registered in `executables/referenceApp/consoleCommands/src/lifecycle/console/LifecycleControlCommand.cpp:42`, fired at `:91`).

**Expected behavior and actual behavior**

Expected: the MCU resets.

Summarised from `arm-none-eabi-objdump -d app.referenceApp.elf`:

```
0x9ea8  etl_assert_function : push {r3, lr} ; bl 5ec <abort>
0x05ec  abort               : push {r3, lr} ; movs r0,#6 ; bl 628 <raise>
                                            ; movs r0,#1 ; bl 684 <_exit>
0x0674  _kill               : errno = 88 (ENOSYS) ; return -1
0x0684  _exit               : e7fe   b.n 684
```

`raise(SIGABRT)` ends up in libnosys' `_kill`, which sets `ENOSYS` and returns -1, so no signal is ever delivered. Control falls through to `_exit`, and `_exit` is a two-byte branch to itself. There is only one `_exit` in the image, nothing overrides it, and nothing on the path executes `cpsid`, so it spins at task level with interrupts still enabled.

The handler is the bare-metal branch at `executables/referenceApp/application/src/main.cpp:35`, installed at `main.cpp:40`. Asserts are enabled: `executables/referenceApp/etl_profile/etl_profile.h:36` defines `ETL_USE_ASSERT_FUNCTION`, and `ETL_NO_CHECKS` is not defined by any openbsw build configuration. `git grep ETL_ASSERT` over `*.c/*.h/*.cpp/*.hpp` excluding `libs/3rdparty` gives 142 occurrences in 66 files, including on-target code such as `executables/referenceApp/platforms/s32k148evb/bspConfiguration/src/bsp/uart/UartConfig.cpp:37` and `executables/referenceApp/application/src/can/DoCanMultiAddressingTransportLayer.cpp:49`.

What that costs depends on which task asserted. `configUSE_PREEMPTION` is 1 (`libs/bsw/asyncFreeRtos/freeRtosConfiguration/FreeRTOSConfig.h:21`) and `FreeRtosAdapter::initTask` passes the context index straight through as the FreeRTOS priority (`FreeRtosAdapter.h:335-341`), so the task enum in `executables/referenceApp/asyncCoreConfiguration/include/async/Config.h` is also the priority order, one task per level. For the `lc assert` repro the command runs in `TASK_SYSADMIN` (priority 7, wired at `app.cpp:344`), which leaves `TASK_BACKGROUND`, `TASK_BSP`, `TASK_UDS`, `TASK_DEMO`, `TASK_ETHERNET`, `TASK_CAN` and idle all starved. CAN and Ethernet processing stops; their interrupts still fire, but nothing drains them.

**Why the watchdog does not stop it**

I expected the watchdog to catch this, but it can't. `SafeWatchdog` is enabled with a 250 ms timeout (`executables/referenceApp/safety/safeWatchdog/include/safeWatchdog/SafeWatchdog.h:46`). `SafetySystem` cycles at 10 ms and `SafetyManager` kicks the watchdog on every 8th pass, so roughly every 80 ms (`executables/referenceApp/platforms/s32k148evb/safety/safeLifecycle/include/safeLifecycle/SafetyManager.h:44`, and the comment saying as much at `SafetyManager.cpp:63`). That runs in `TASK_SAFETY`, priority 8, above every task listed above, so it keeps running and keeps servicing the watchdog while the lower tasks are dead. No reset.

To trip the watchdog the assert would have to happen at or above `TASK_SAFETY`'s priority, or in an ISR. Note `TASK_SAFETY` is not the top priority in the system either: the FreeRTOS timer daemon sits at 9 (`FreeRTOSConfig.h:49`).

There is no task-level supervision to fall back on. `safeMonitor::Watchdog`, the alive monitor template, has no users outside its own unit test.

**Environment**

`main` @ 9b362b8, arm-gnu-toolchain 14.3.Rel1 (gcc 14.3.1), the release pinned in `docker/development/Dockerfile:7`. Ubuntu 24.04.

**Additional context**

Smallest fix I can see is to give the handler a definite end state rather than letting it fall into newlib's `_exit`: log, then call `softwareSystemReset()` (`platforms/s32k1xx/bsp/bspMcu/src/reset/softwareSystemReset.cpp:17`), which is what `SafeSupervisor` already does for the conditions it monitors (`executables/referenceApp/safety/safeSupervisor/src/safeSupervisor/SafeSupervisor.cpp:196-198`).

Happy to write the PR, but I'd rather check that's the direction you want first.

Found this while looking into the linker warnings reported in #511.

ThePo1es:
`etl_assert_function()` called `std::abort()`, which on the bare metal targets ends in newlib's `_exit()`, a two byte branch to itself. The asserting task spins forever. Recovery was expected from the hardware watchdog, but the watchdog is serviced from `TASK_SAFETY`, the highest priority application task, which keeps running while a lower priority task is stuck, so no reset happens. Details and measurements are in #575.

This calls `softwareSystemReset()` directly, matching `HardFault_Handler_Final` in `isr_sys.cpp` and `libs/rust/panic_handler/src/lib.rs`.

The `__linux__` split is dropped because `softwareSystemReset()` exists on all four platforms. `ETL_MINIMAL_ERRORS` leaves `what()`, `file_name()` and `line_number()` empty, so the return address is printed instead; `addr2line` resolves it against the ELF from the same build.

On POSIX the exit code changes from 134 to 0, because `platforms/posix/bsp/bspMcu/src/reset/softwareSystemReset.cpp` is `terminal_cleanup(); _exit(0);`. Nothing observes it today: `test/pyTest/process_mgmt.py` relaunches on `returncode is not None` without reading the value, and no pytest sends `lc assert`.

Dropping `std::abort()` also removes the `raise(SIGABRT)` chain, so `_kill` and `_getpid` leave the link. Two of the six `not implemented` warnings in #511 disappear. Text grows by 1792 bytes (260532 to 262324) for the `etl::print` call.

No unit test: `application/src/main.cpp` is not built by any `tests-*` preset.

Verified: `s32k148-freertos-gcc` and `posix-freertos` build; `lc assert` on POSIX prints `ETL_ASSERT failed at 0x...`; `arm-none-eabi-nm` shows `abort`, `raise`, `_exit`, `_kill` and `_getpid` all gone; cr_checker passes. I have no S32K148 board, so this is not verified on hardware.

Resolves #575. See also #511.

### DIFF
--- a/executables/referenceApp/application/src/main.cpp
+++ b/executables/referenceApp/application/src/main.cpp
@@ -10,30 +10,30 @@
 
 #include "app/app.h"
 
+#include "reset/softwareSystemReset.h"
+
 #include <etl/error_handler.h>
+#include <etl/platform.h>
+#include <etl/print.h>
 
-#if defined(__linux__) || defined(__APPLE__)
-#include <cstdlib>
-#include <iostream>
-#if __has_include("execinfo.h")
-#include <execinfo.h>
-#endif
+#include <cstdint>
 
-void etl_assert_function(etl::exception const& exception)
+/**
+ * Reports a failed ETL_ASSERT and resets the ECU.
+ *
+ * etl_profile.h defines ETL_MINIMAL_ERRORS, so what(), file_name() and line_number() of the
+ * exception carry no information. The return address is printed instead, which resolves back
+ * to the ETL_ASSERT call site with addr2line against the ELF of the same build.
+ *
+ * The reset is explicit rather than left to the watchdog, because the watchdog is serviced
+ * from TASK_SAFETY, which keeps running when a lower priority task stops. See #575.
+ */
+ETL_NORETURN void etl_assert_function(etl::exception const&)
 {
-#if __has_include("execinfo.h")
-    void* callstack[128];
-    backtrace_symbols_fd(callstack, backtrace(callstack, 128), 2 /* = stderr */);
-#endif
-    std::cout << "Assertion at " << exception.file_name() << ':' << exception.line_number() << " ("
-              << exception.what() << ") failed" << std::endl;
-    std::abort();
+    etl::print(
+        "ETL_ASSERT failed at {:#x}\r\n", reinterpret_cast<uintptr_t>(__builtin_return_address(0)));
+    softwareSystemReset();
 }
-#else // defined(__linux__) || defined(__APPLE__)
-#include <cstdlib>
-
-void etl_assert_function(etl::exception const&) { std::abort(); }
-#endif
 
 void app_main()
 {


executables/referenceApp/application/src/main.cpp
```diff
@@ -10,30 +10,29 @@
...
+ * Reports a failed ETL_ASSERT and resets the ECU.
+ *
+ * etl_profile.h defines ETL_MINIMAL_ERRORS, so what(), file_name() and line_number() of the
+ * exception carry no information. The return address is printed instead, which resolves back
+ * to the ETL_ASSERT call site with addr2line against the ELF of the same build.
+ *
+ * The reset is explicit rather than left to the watchdog, because the watchdog is serviced
+ * from TASK_SAFETY, which keeps running when a lower priority task stops. See #575.
+ */
+[[noreturn]] void etl_assert_function(etl::exception const&)
```
rolandreichweinbmw: ETL_NORETURN

ThePo1es: Done, thanks.

rolandreichweinbmw: Thank you for the update. @christophruethingbmw will have a look also.

christophruethingbmw (approved):
looks good to me

rolandreichweinbmw (approved):

PR #586, opened 2026-09-04
Support middlewareConfiguration and codegen infrastructure in Bazel (2/5)

DominikAFischer:
Support middlewareConfiguration and codegen infrastructure in Bazel

middleware/tools/cpp_generator:
- middleware_codegen
- generator

middlewareConfiguration:
- middleware_configuration
- generated_outputs_drift_test

application:
- demo_logger

Note: Python invoked via dev-container venv; generator deps are
pre-installed in the image, no network access at build time.

### DIFF
--- a/executables/referenceApp/middlewareConfiguration/BUILD.bazel
+++ b/executables/referenceApp/middlewareConfiguration/BUILD.bazel
@@ -1,0 +1,86 @@
+# *******************************************************************************
+# Copyright (c) 2026 Accenture
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load(
+    "//libs/bsw/middleware/tools/cpp_generator:middleware_codegen.bzl",
+    "middleware_codegen",
+)
+load(":generated_outputs.bzl", "GENERATED_OUTPUTS")
+
+middleware_codegen(
+    name = "middleware_configuration_generated",
+    deployment_yaml = "model/deployment.yaml",
+    generated_outputs = GENERATED_OUTPUTS,
+    visibility = ["//visibility:public"],
+    deps = ["//libs/bsw/middleware"],
+)
+
+# Drift guard: serialise the declared output list for comparison.
+write_file(
+    name = "generated_outputs_expected",
+    out = "generated_outputs_expected.txt",
+    content = sorted(GENERATED_OUTPUTS) + [""],
+    newline = "unix",
+)
+
+# Drift guard: query the generator for its actual output list without rendering.
+genrule(
+    name = "generated_outputs_actual",
+    srcs = [
+        "model/deployment.yaml",
+        "//libs/bsw/middleware/tools/cpp_generator:generator",
+    ],
+    outs = ["generated_outputs_actual.txt"],
+    cmd = ("/opt/venv/bin/python3" +
+           " libs/bsw/middleware/tools/cpp_generator/jinja2cpp.py" +
+           " --input libs/bsw/middleware/tools/cpp_generator" +
+           " --deployment-yaml $(execpath model/deployment.yaml)" +
+           " --list-outputs | LC_ALL=C sort > $@"),
+    target_compatible_with = ["@platforms//os:linux"],
+)
+
+# Drift guard: fail if generated_outputs.bzl is out of sync with the generator.
+diff_test(
+    name = "generated_outputs_drift_test",
+    failure_message = "generated_outputs.bzl is stale; regenerate it (see the header of that file).",
+    file1 = "generated_outputs_expected.txt",
+    file2 = "generated_outputs_actual.txt",
+    target_compatible_with = ["@platforms//os:linux"],
+)
+
+cc_library(
+    name = "middleware_configuration",
+    srcs = [
+        "platform_integration/logger/src/LoggerImpl.cpp",
+        "platform_integration/os/src/OsDefinitions.cpp",
+        "platform_integration/time/src/SystemTimeProvider.cpp",
+        "src/MemoryLayout.cpp",
+    ],
+    hdrs = ["include/MemoryLayout.h"],
+    implementation_deps = [
+        "//executables/referenceApp/application:demo_logger",
+        "//libs/bsw/bsp",
+    ],
+    strip_include_prefix = "include",
+    target_compatible_with = ["@platforms//os:linux"],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":middleware_configuration_generated",
+        "//executables/referenceApp/asyncBinding:async_binding",
+        "//libs/bsw/logger",
+        "//libs/bsw/middleware",
+    ],
+    # NOTE: alwayslink is required because LoggerImpl.cpp implements
+    # middleware::logger::logBinary(), a symbol declared inside //libs/bsw/middleware.
+    alwayslink = True,
+)
--- a/executables/referenceApp/middlewareConfiguration/BUILD.bazel
+++ b/executables/referenceApp/middlewareConfiguration/BUILD.bazel
@@ -8,8 +8,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
 
-load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
-load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
 load(
     "//libs/bsw/middleware/tools/cpp_generator:middleware_codegen.bzl",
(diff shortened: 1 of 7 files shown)


executables/referenceApp/middlewareConfiguration/generated_outputs.bzl
```diff
@@ -0,0 +1,49 @@
...
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+"""Output layout of the middleware code generator for model/deployment.yaml.
+
+The list is data-dependent (it varies with the deployment model). It mirrors the
+hard-coded output list in the former CMake `middlewareConfiguration/CMakeLists.txt`
```
rolandreichweinbmw, line 14: Please change the wording since it is misleading:

I recommend:
* "former CMake" -> "mirroring CMake"

Please note that we don't replace CMake with Bazel, it is an additional build system. At least for now.

libs/bsw/middleware/tools/cpp_generator/middleware_codegen.bzl
```diff
@@ -0,0 +1,96 @@
...
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+"""Middleware C++ code generation macro.
+
+`middleware_codegen` runs `jinja2cpp.py` over a deployment YAML model and wraps
+the generated C++ in a `cc_library`. It replaces the CMake `add_custom_command`
```
rolandreichweinbmw, line 14: Similarly here.

libs/bsw/middleware/tools/cpp_generator/README.md
```diff
@@ -46,19 +46,45 @@ python3 jinja2cpp.py \
...
 | `--clang-format-config` | No | Path to a `.clang-format` config file; auto-detected from repo root if omitted |
 | `--no-clean` | No | Skip deleting the output directories before generation |
 | `--generation-mode` | No | Default generation mode: `normal` (full generation) or `mock` (interface-only for testing). Default: `normal` |
+| `--list-outputs` | No | Print the relative path of every file a generation run would produce for the given `--deployment-yaml` (one per line, on stdout) and exit **without rendering or writing anything**. Loader diagnostics go to stderr so stdout stays a clean path list. Used to declare the generated output set to build systems (see [Bazel build integration](#bazel-build-integration)) |
+
+## Bazel build integration
+
+The Bazel build wraps this script in the `middleware_codegen` macro
+(`libs/bsw/middleware/tools/cpp_generator/middleware_codegen.bzl`), which replaces
+the CMake `add_custom_command` + `add_library` pair. A `genrule` runs the script
```
rolandreichweinbmw, line 69: Dito.

rolandreichweinbmw (commented):
Please also adjust the PR and commit message: We support bazel, but we don't "migrate" to it.

executables/referenceApp/middlewareConfiguration/generated_outputs.bzl
DominikAFischer: done

libs/bsw/middleware/tools/cpp_generator/middleware_codegen.bzl
DominikAFischer: done

libs/bsw/middleware/tools/cpp_generator/README.md
DominikAFischer: done

DominikAFischer:
@rolandreichweinbmw  Agree with your remarks and fixed them. Generally, I believe migration is still the correct technical term for what we're doing even if we won't drop support for CMake but support the two build automation tools side by side. Would it be okay with you If we keep the term in general?

rolandreichweinbmw:
> @rolandreichweinbmw Agree with your remarks and fixed them. Generally, I believe migration is still the correct technical term for what we're doing even if we won't drop support for CMake but support the two build automation tools side by side. Would it be okay with you If we keep the term in general?

Migration by definition is a move from A (cmake?) to B (bazel?), which implicitly means "drop A". I think that can be easily understood that way, even if we agree on "supporting bazel" semantically.

Thanks for adjusting the other phrasings.

libs/bsw/middleware/tools/cpp_generator/README.md
```diff
@@ -46,19 +46,42 @@ python3 jinja2cpp.py \
...
+
+The Bazel build wraps this script in the `middleware_codegen` macro
+(`libs/bsw/middleware/tools/cpp_generator/middleware_codegen.bzl`), which mirrors
+the CMake `add_custom_command` + `add_library` pair. A `genrule` runs the script
+and a wrapper `cc_library` exposes the generated headers/sources.
+
+Key points:
+
+- **Interpreter:** the genrule invokes `/opt/venv/bin/python3` by absolute path . Bazel therefore 
+  relies on the development Docker image already having the dependencies installed in `/opt/venv`.
```
rolandreichweinbmw, line 75: As you suggest by this description, you intentionally restrict the use of bazel to the combination of bazel with the shipped docker image.

Until now, I was able to run bazel in other environments as well, what I really like.

Even within the docker container, would it harm to make `python3` available as `/opt/venv/bin/python3`, unambiguously (e.g. via PATH)? That way, both use cases could be supported, which broadens the user base for bazel. I.e. in everyone's interest? Fine for me if we limit official support to the environment defined by the Dockerfile.

Just a food for thought. I could live with the fact that bazel is limited to a narrowed-down environment, leaving cmake for the more general use cases.

rolandreichweinbmw (dismissed):

libs/bsw/middleware/tools/cpp_generator/README.md
DominikAFischer: Good point, thank you for the feedback!

In general, using rules_python with Bazel managed pip deps would be the go-to approach for hermeticity. Since online dependencies should be avoided (if that is a hard requirement), the docker approach allows us still to be somewhat hermetic.

That said, I think we need to make a switch anyhow, since the code generators will also not work for external integrators that won't use the docker setup at all.

My proposal for a proper solution: replace the hardcoded reference with a `py_runtime` toolchain, defaulting to the Docker path. External users register their own `py_runtime` pointing to their local interpreter via register_toolchains() in their MODULE.bazel.
This makes the interpreter an overridable, Bazel-tracked concept rather than a hardcoded string, and is fully offline-compatible. Python dependencies for full hermeticity may then be vendored inrepo.

libs/bsw/middleware/tools/cpp_generator/middleware_codegen.bzl
```diff
@@ -0,0 +1,91 @@
...
+    cc_library(
+        name = name,
+        srcs = srcs,
+        hdrs = hdrs,
+        strip_include_prefix = gen_root + "/include/generated_code",
+        # alwayslink is required because the generated .cpp files implement
+        # symbols declared inside //libs/bsw/middleware, e.g.
+        # AllocatorSelectorDefinitions.cpp implements middleware::memory::getAllocFunction().
+        alwayslink = True,
+        target_compatible_with = _LINUX_ONLY,
```
christophruethingbmw: Why is the generated library only compatible with linux, isn't the same middleware generation also used for the target later?

executables/referenceApp/middlewareConfiguration/BUILD.bazel
```diff
@@ -0,0 +1,86 @@
...
+middleware_codegen(
+    name = "middleware_configuration_generated",
+    deployment_yaml = "model/deployment.yaml",
+    generated_outputs = GENERATED_OUTPUTS,
+    visibility = ["//visibility:public"],
+    deps = ["//libs/bsw/middleware"],
+)
+
+# Drift guard: serialise the declared output list for comparison.
+write_file(
```
christophruethingbmw: We have the `middleware_codegen` as a macro to generate a `genrule` and `cc_library`. Would it make sense to also generate the `diff_test and what is necessary automatically? I think it is a kind of _must-have_ for every usage? Or at least provide a macro for this as well?

executables/referenceApp/middlewareConfiguration/BUILD.bazel
```diff
@@ -0,0 +1,86 @@
...
+    srcs = [
+        "model/deployment.yaml",
+        "//libs/bsw/middleware/tools/cpp_generator:generator",
+    ],
+    outs = ["generated_outputs_actual.txt"],
+    cmd = ("/opt/venv/bin/python3" +
+           " libs/bsw/middleware/tools/cpp_generator/jinja2cpp.py" +
+           " --input libs/bsw/middleware/tools/cpp_generator" +
+           " --deployment-yaml $(execpath model/deployment.yaml)" +
+           " --list-outputs | LC_ALL=C sort > $@"),
```
christophruethingbmw: Should we maybe add the sorting in the `jinja2cpp.py` directly instead of here using `sort`?

executables/referenceApp/middlewareConfiguration/BUILD.bazel
```diff
@@ -0,0 +1,86 @@
...
+    target_compatible_with = ["@platforms//os:linux"],
+)
+
+# Drift guard: fail if generated_outputs.bzl is out of sync with the generator.
+diff_test(
```
christophruethingbmw: This is executed at build time right? It's not necessary to call `bazel test` for it`?

christophruethingbmw: I remember this same approach from old CMake times where I did a very similar thing to make sure that the list of files written in the `CMakeLists.txt` was always up to date. Maybe we can think for the middleware and whatever we format we decide for in the future to be designed in a way that we can easily determine the output files from the input files, e.g. have a single input file per interface.

That's independent of this PR so I think it's fine like this, it is safe, avoids tree artifacts and should also be not a big performance penalty.

libs/bsw/middleware/tools/cpp_generator/README.md
christophruethingbmw: You proposal sounds good to me, I also would prefer to go with rules_python in the future and with bazel managed pips. I think that's the standard way bazel does it and if we have a pre-populated cache in the docker image we could even run potentially completely offline? I think once we are ready with the bazel implementation, we need to decide on the approach to follow here. Might be a longer discussion but I think it would be valuable.

Can you adapt the PR according to your proposal, then we can see if this would suit all needs.

libs/bsw/middleware/tools/cpp_generator/middleware_codegen.bzl
DominikAFischer: That's a leftover from a previous implementation. It definitely needs to be removed. Thank you for pointing it out!

executables/referenceApp/middlewareConfiguration/BUILD.bazel
DominikAFischer: makes sense, done

executables/referenceApp/middlewareConfiguration/BUILD.bazel
DominikAFischer: No not at build time. This just sets up a diff test, that will be executed in the CI's `bazel test //...` and will prevent merging of a PR that has a mismatch. I think that's a good enough gate for this. Do you agree?

executables/referenceApp/middlewareConfiguration/BUILD.bazel
DominikAFischer: makes sense, implemented

libs/bsw/middleware/tools/cpp_generator/README.md
DominikAFischer: I would propose to move this change to a separate PR since it makes some substantial changes. Would that be alright for you?

christophruethingbmw: Fine for me, what we have right now is at least working although maybe not perfect for all environments. Bazel support is anyway still preliminary, so I think its fine to improve this in a separate PR.

executables/referenceApp/middlewareConfiguration/BUILD.bazel
christophruethingbmw: I think it's the best option we have right now ;) I think in the future (as mentioned in the next comment) we need to find a better way for this. But the `yaml` we have here is anyway also preliminary at the moment. So whatever comes (whether it is FLYNC, ...) should be designed such that we can remove this :)

christophruethingbmw (approved):

Issue #567, opened 2026-08-26
Mock libraries declare their gmock dependency incorrectly


DominikAFischer:
**Describe the bug**
Some mock libs declare their gmock dependency incorrectly:
- **category A**: link `gmock_main` instead of `gmock`. A library should not pull in the main() test entry point.
- **category B**: use `<gmock/gmock.h>` in their headers but declare no gmock dependency.

Both compile only because the consuming test links `gmock_main` the mock's declared deps don't reflect what it actually uses.

category A:
- `libs/bsw/bsp/mock/CMakeLists.txt`: `bspMock`
- `libs/bsw/cpp2can/CMakeLists.txt`: `cpp2canMock`
- `libs/bsw/io/CMakeLists.txt`: `ioMock`
- `libs/bsw/logger/CMakeLists.txt`: `loggerMock`
- `libs/bsw/storage/CMakeLists.txt`: `storageMock`
- `platforms/s32k1xx/bsp/bspIo/CMakeLists.txt`: `bspIoMock`

category B:
- `libs/bsw/async/mock/CMakeLists.txt`: `asyncMock`
- `libs/bsw/util/CMakeLists.txt`: `utilMock`
- `libs/bsw/cpp2ethernet/CMakeLists.txt`: `cpp2ethernetMock`
- `libs/bsw/docan/CMakeLists.txt`: `docanMock`

**Expected behavior and actual behavior**
- Expected: each mock links against `gmock`. `gmock_main` is only used in test executables. E.g. `target_link_libraries(lifecycleMock INTERFACE gmock lifecycle)`.
- Actual: Some mock targets leak `main()` into every consumer and some mock targets do not declare what they consume but instead rely on the consumer to supply the needed dependency.

**Environment**
OpenBSW development docker container. Environment-independent.

**Additional context**

PR #584, opened 2026-09-04
Add Bazel support for remaining libs/bsw mock targets

SuhashiniNaik:
Add Bazel support for remaining libs/bsw mock targets

Migrated targets:

//libs/bsw/cpp2can:cpp2can_mock
//libs/bsw/cpp2ethernet:cpp2ethernet_mock
//libs/bsw/transport:transport_mock
//libs/bsw/docan:docan_mock
//libs/bsw/doip:doip_mock
//libs/bsw/uds:uds_mock
//libs/bsw/storage:storage_mock
//libs/bsp/bspInterrupts:bsp_interrupts_mock

edit: Addition of some more targets 
//libs/bsw/cpp2someip
//libs/bsw/middleware

PR #590, opened 2026-09-06
Fix gmock dependency declarations of mock libraries

ThePo1es:
Some mock libraries linked `gmock_main` instead of `gmock`, which pulls the test entry point into every consumer of the mock. Others used `<gmock/gmock.h>` in their headers without declaring any gmock dependency and relied on the consuming test to provide it.

Link `gmock` in all ten mock targets. Mocks whose headers include gmock declare it PUBLIC or INTERFACE so that consumers get the include paths.

Seven test executables had no `main()` of their own and obtained it transitively from a mock's `gmock_main`. They now link `gmock_main` explicitly: bspTest, cpp2canTest, ioTest, loggerTest, BxCanTransceiverTest, FdCanTransceiverTest and CanFlex2TransceiverTest.

Verified with the tests-posix-debug, tests-s32k1xx-debug and tests-stm32-debug presets: configure, build and ctest all pass (2564, 40 and 911 tests). `bspIoMock` is changed for consistency only; no preset configures it, because `platforms/s32k1xx/bsp/CMakeLists.txt` skips `bspIo` in unit test builds. `transportMock` shows the same pattern (PRIVATE gmock with gmock in its headers) but is not in the issue list and is left untouched.

CMake only. These mocks have no Bazel targets on main yet; #570 and #584 add them, and #584 currently declares `gtest_main` for `bsp_mock`, `cpp2can_mock` and `io_mock`.

Resolves:
#567: Mock libraries declare their gmock dependency incorrectly

### DIFF
--- a/libs/bsw/cpp2can/CMakeLists.txt
+++ b/libs/bsw/cpp2can/CMakeLists.txt
@@ -25,5 +25,8 @@ if (BUILD_EXECUTABLE STREQUAL "unitTest")
 
     target_include_directories(cpp2canMock PUBLIC mock/include)
 
-    target_link_libraries(cpp2canMock PRIVATE bspMock cpp2can gmock_main)
+    target_link_libraries(
+        cpp2canMock
+        PUBLIC cpp2can gmock
+        PRIVATE bspMock)
 endif ()
--- a/platforms/s32k1xx/bsp/bspIo/CMakeLists.txt
+++ b/platforms/s32k1xx/bsp/bspIo/CMakeLists.txt
@@ -15,5 +15,5 @@ if (BUILD_EXECUTABLE STREQUAL "unitTest")
 
     target_include_directories(bspIoMock PUBLIC mock/gmock/include)
 
-    target_link_libraries(bspIoMock PRIVATE gmock_main)
+    target_link_libraries(bspIoMock PUBLIC bspIo etl gmock)
 endif ()
--- a/platforms/s32k1xx/bsp/canflex2Transceiver/test/CMakeLists.txt
+++ b/platforms/s32k1xx/bsp/canflex2Transceiver/test/CMakeLists.txt
@@ -9,7 +9,7 @@ target_link_libraries(
             bspIo
             lifecycle
             bspMock
-            gmock)
+            gmock_main)
 
 gtest_discover_tests(CanFlex2TransceiverTest
                      PROPERTIES LABELS "CanFlex2TransceiverTest")
--- a/platforms/stm32/bsp/bxCanTransceiver/test/CMakeLists.txt
+++ b/platforms/stm32/bsp/bxCanTransceiver/test/CMakeLists.txt
@@ -9,7 +9,7 @@ target_link_libraries(
             bspIo
             lifecycle
             bspMock
-            gmock)
+            gmock_main)
 
 gtest_discover_tests(BxCanTransceiverTest PROPERTIES LABELS
                                                      "BxCanTransceiverTest")
--- a/platforms/stm32/bsp/fdCanTransceiver/test/CMakeLists.txt
+++ b/platforms/stm32/bsp/fdCanTransceiver/test/CMakeLists.txt
@@ -9,7 +9,7 @@ target_link_libraries(
             bspIo
             lifecycle
             bspMock
-            gmock)
+            gmock_main)
 
 gtest_discover_tests(FdCanTransceiverTest PROPERTIES LABELS
                                                      "FdCanTransceiverTest")
--- a/libs/bsw/async/mock/CMakeLists.txt
+++ b/libs/bsw/async/mock/CMakeLists.txt
@@ -2,7 +2,7 @@ add_library(asyncMock INTERFACE)
 
 target_include_directories(asyncMock INTERFACE gmock/include)
 
-target_link_libraries(asyncMock INTERFACE etl platform)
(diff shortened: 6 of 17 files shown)


platforms/s32k1xx/bsp/canflex2Transceiver/test/CMakeLists.txt
```diff
@@ -9,7 +9,8 @@ target_link_libraries(
             bspIo
             lifecycle
             bspMock
-            gmock)
+            gmock
+            gmock_main)
```
rolandreichweinbmw, line 12: Better drop gmock here since gmock_main depends on it already.

See also https://github.com/eclipse-openbsw/openbsw/pull/567

platforms/stm32/bsp/bxCanTransceiver/test/CMakeLists.txt
```diff
@@ -9,7 +9,8 @@ target_link_libraries(
             bspIo
             lifecycle
             bspMock
-            gmock)
+            gmock
+            gmock_main)
```
rolandreichweinbmw, line 12: Same as for platforms/s32k1xx/bsp/canflex2Transceiver/test/CMakeLists.txt

platforms/stm32/bsp/fdCanTransceiver/test/CMakeLists.txt
```diff
@@ -9,7 +9,8 @@ target_link_libraries(
             bspIo
             lifecycle
             bspMock
-            gmock)
+            gmock
+            gmock_main)
```
rolandreichweinbmw, line 12: Same as for platforms/s32k1xx/bsp/canflex2Transceiver/test/CMakeLists.txt

libs/bsw/cpp2can/CMakeLists.txt
```diff
@@ -25,5 +25,8 @@ if (BUILD_EXECUTABLE STREQUAL "unitTest")
 
     target_include_directories(cpp2canMock PUBLIC mock/include)
 
-    target_link_libraries(cpp2canMock PRIVATE bspMock cpp2can gmock_main)
+    target_link_libraries(
+        cpp2canMock
+        PUBLIC gmock
+        PRIVATE bspMock cpp2can)
```
rolandreichweinbmw: cpp2canMock scope is still wrong for cpp2can. Its public headers include can/transceiver/AbstractCANTransceiver.h, can/filter/IFilter.h, etc., so cpp2can belongs in PUBLIC, not PRIVATE. It only builds because every consumer links cpp2can directly.

platforms/s32k1xx/bsp/bspIo/CMakeLists.txt
```diff
@@ -15,5 +15,5 @@ if (BUILD_EXECUTABLE STREQUAL "unitTest")
 
     target_include_directories(bspIoMock PUBLIC mock/gmock/include)
 
-    target_link_libraries(bspIoMock PRIVATE gmock_main)
+    target_link_libraries(bspIoMock PUBLIC gmock)
```
rolandreichweinbmw: Headers include io/Io.h, bsp/Bsp.h and etl/singleton_base.h, but the target now links only gmock. PUBLIC bspIo etl gmock.

ThePo1es:
The run-pytest-on-posix failure is in the posix-threadx leg, docan/test_docan_dispatch_integration.py::TestGatewayAdversarial::test_r16_true_concurrent_burst_all_schemes: the EA and RangeExtended legs of the four-way burst hit the 3 s request timeout. The posix-freertos leg passed completely and the other 181 tests of the threadx leg passed. Both reference applications built and linked before pytest started. The CMake changes here are all inside unitTest blocks or test directories and are not configured in those builds.

The same job failed on main after #588 was merged (run 34023299598), also in the threadx leg, with a different timeout: console/test_lifecycle.py::test_lc_levelchange at wait_for_boot_complete. Both runs compiled the applications with GCC 13.3.0, so the container change is not the difference. I can't re-run the workflow from a fork.

rolandreichweinbmw:
> The run-pytest-on-posix failure is in the posix-threadx leg, docan/test_docan_dispatch_integration.py::TestGatewayAdversarial::test_r16_true_concurrent_burst_all_schemes: the EA and RangeExtended legs of the four-way burst hit the 3 s request timeout. The posix-freertos leg passed completely and the other 181 tests of the threadx leg passed. Both reference applications built and linked before pytest started. The CMake changes here are all inside unitTest blocks or test directories and are not configured in those builds.
> 
> The same job failed on main after #588 was merged (run 34023299598), also in the threadx leg, with a different timeout: console/test_lifecycle.py::test_lc_levelchange at wait_for_boot_complete. Both runs compiled the applications with GCC 13.3.0, so the container change is not the difference. I can't re-run the workflow from a fork.

Right: The test setup is broken regarding timeouts and we need to fix it separately. For now, typically just re-running tests in case of failure.

ThePo1es:
Applied all five review comments and force-pushed as c4d67a56: dropped the redundant gmock from the three transceiver tests, moved cpp2can to PUBLIC on cpp2canMock, and linked bspIo and etl on bspIoMock. tests-posix-debug, tests-stm32-debug and tests-s32k1xx-debug rebuilt and pass.

rolandreichweinbmw:
Thank you for the update. Looks good to me!

rolandreichweinbmw (approved):

matthiaskessler (approved):

Issue #583, opened 2026-09-04
LwipSocket: tcp_write() without tcp_output() causes indefinite send delay


raveenaKpit:
**Describe the bug**
In `LwipSocket::sendPendingData()`, data is enqueued into the lwIP TCP send buffer via
`tcp_write()` but `tcp_output()` is **never called afterward**. This means the enqueued
data is not immediately flushed to the network interface. Instead, lwIP holds the data
in its internal send queue and only transmits it when:

- A TCP ACK arrives from the peer (which triggers the `tcpSentCallback` path), **or**
- The lwIP TCP slow timer fires (every 500 ms by default), **or**
- Nagle's algorithm decides enough data has accumulated.

In practice this means that **short, single-packet responses — such as DoIP Routing
Activation Responses (`0x8001`) or SOME/IP SD offer messages — are silently delayed by
up to 500 ms or never transmitted at all during the session setup phase**. External
clients (e.g., a DoIP tester or an Android SOME/IP client) time out waiting for the
response and report the connection as failed.

This is consistent with the lwIP documentation and the recommended usage pattern:
> *"`tcp_write()` does not actually transmit the data but merely enqueues it for future
> transmission. You must call `tcp_output()` if you want the data to be transmitted
> immediately."*
> — lwIP wiki, [Raw TCP API](https://lwip.fandom.com/wiki/Raw/TCP)

---

**Steps to reproduce the bug**

*Environment prerequisites*

1. Build the OpenBSW POSIX reference application (preset `posix-freertos`):
   ```bash
   cmake --preset posix-freertos
   cmake --build --preset posix-freertos
   ```

2. Configure the host TAP interface (requires `sudo`):
   ```bash
   sudo ip tuntap add dev tap0 mode tap
   sudo ip addr add 192.168.0.1/24 dev tap0
   sudo ip link set tap0 up
   sudo ip route add 224.244.224.245 dev tap0   # SOME/IP SD multicast
   ```

3. Start the reference application:
   ```bash
   ./build/posix-freertos/app/common/application/Release/app.referenceApp.elf
   ```

*Reproducing the DoIP hang*

4. From the host, open a DoIP Routing Activation connection on TCP port 13400:
   ```bash
   python3 - <<'EOF'
   import socket, time

   # DoIP Routing Activation Request (SA=0x0E00, activation_type=0x00)
   doip_req = bytes([
       0x02, 0xFD,       # Protocol version
[...]

PR #595, opened 2026-09-07
Fix issue583

raveenaKpit:


rolandreichweinbmw:
As raised in https://github.com/eclipse-openbsw/openbsw/issues/583, the issue is not reproducable.

Further, there is sth. wrong with the premise: SOME/IP SD is UDP, therefore not relevant in the TCP send path.

Besides, doesn't explicit tcp_output() conflict with the concept of intentional coalescing of packets? Nagle and the purpose of flush()? This increases packet count beyond previously intended?

Would be interesting how you reproduce the issue since your IP address (.3) seems to be different from the original POSIX (.201) and S32K148 (.200).

libs/bsw/lwipSocket/src/lwipSocket/tcp/LwipSocket.cpp
```diff
@@ -171,6 +171,12 @@ err_t LwipSocket::sendPendingData(tcp_pcb* const pcb)
     err_t result = tcp_write(pcb, fPendingTcpData.data(), static_cast<uint16_t>(bytesToWrite), 1);
     if (result == ERR_OK)
     {
+        // Flush immediately: tcp_write() only queues data in lwIP's send buffer.
+        // Without tcp_output(), the data stays queued until the next TCP timer event,
+        // which can be hundreds of milliseconds later — causing DoIP routing activation
+        // responses (and all other TCP replies) to never reach the client in time.
+        (void)tcp_output(pcb);
```
matthiaskessler, line 178: It is true, that `tcp_write` only enqueues the data and it will be sent depending on multiple conditions, one of them being the timer event. This is how a TCP socket behaves by default. If you need to send your data faster, there are two tools:
- call the `setNagle(false)` or `disableNagleAlgorithm()` function to disable the Nagle algorithm for a socket instance
- call the `flush()` function manually

This change here would turn every socket into an immediately sending socket which is a restriction we should not do.

libs/bsw/lwipSocket/src/lwipSocket/tcp/LwipSocket.cpp
```diff
@@ -171,6 +171,12 @@ err_t LwipSocket::sendPendingData(tcp_pcb* const pcb)
     err_t result = tcp_write(pcb, fPendingTcpData.data(), static_cast<uint16_t>(bytesToWrite), 1);
     if (result == ERR_OK)
     {
+        // Flush immediately: tcp_write() only queues data in lwIP's send buffer.
+        // Without tcp_output(), the data stays queued until the next TCP timer event,
+        // which can be hundreds of milliseconds later — causing DoIP routing activation
```
matthiaskessler, line 176: IMHO, we shouldn't mention higher level protocols in generic layers.

matthiaskessler (commented):
I am against merging this change, since it affects the behavior in a way limiting the way TCP data stream can be controlled.

rolandreichweinbmw:
Can you please propose an update on this? Otherwise, we will probably close this isse.

PR #596, opened 2026-09-07
Add PLATFORM_SUPPORT feature flags to Bazel (3/5)

DominikAFischer:
Add PLATFORM_SUPPORT feature flags to Bazel

bazel/config/features:
- bool_flag, config_setting and cc_library per PLATFORM_SUPPORT flag
- feature_defines aggregate

bazel/platform:
- Add flags to s32k148 platform (mirrors Options.cmake PLATFORM_SUPPORT matrix)
- Add posix platform (parents = @platforms//host:host + posix feature matrix)
- posix is the DEFAULT platform: plain `bazel build //...` is the posix host build

.bazelrc:
- Replace per-flag approach with platform() flags
- Add build --platforms=//bazel/platform:posix as default

safeWatchdog:
- migrate inline defines to feature_defines

### DIFF
--- a/bazel/config/features/BUILD.bazel
+++ b/bazel/config/features/BUILD.bazel
@@ -1,0 +1,337 @@
+# *******************************************************************************
+# Copyright (c) 2026 Accenture
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+"""
+PLATFORM_SUPPORT feature flags
+
+Each flag is a self-contained triple: bool_flag + config_settings + defines carrier lib.
+All flags default to False. //bazel/platform platform() `flags` sets the matrix per platform.
+Override any flag individually, e.g. --//bazel/config/features:can=false
+
+feature_defines aggregates all carrier libs for the common dep.
+"""
+
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+bool_flag(
+    name = "can",
+    build_setting_default = False,
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "can_enabled",
+    flag_values = {":can": "true"},
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "can_disabled",
+    flag_values = {":can": "false"},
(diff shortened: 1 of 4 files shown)


bazel/config/features/BUILD.bazel
```diff
@@ -0,0 +1,337 @@
...
+All flags default to False. //bazel/platform platform() `flags` sets the matrix per platform.
+Override any flag individually, e.g. --//bazel/config/features:can=false
+
+feature_defines aggregates all carrier libs for the common dep.
+"""
+
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+
+bool_flag(
```
christophruethingbmw, line 24: For each flag we need currently 26 lines of code, would it make sense to define a macro to generate all of this and then have an easier to read list of all supported flags? Maybe something like (not tested):

```starlark
def feature_flag(name):
  bool_flag(
    name = "{}".format(name),
    build_setting_default = Flase,
  )
  config_setting(
    name = "{}_enabled".format(name),
    flag_values = {":{}".format(name): "true"},
  )

  config_setting(
    name = "{}_disabled".format(name),
    flag_values = {":{}".format(name): "false"},
  )

  cc_library(
    name = "{}_defines".format(name),
    defines = select({
        ":can_enabled": ["PLATFORM_SUPPORT_{}=1".format(name.upper()],
        "//conditions:default": [],
    }),
  )
```

DominikAFischer: Thanks for the suggestion! I understand where this is coming from but to be honest, I personally would prefer to keep it explicit:
- In general, that's the Bazel philosophy (explicit over implicit)
- This would add indirection and cognitive overhead that I personally think is not worth saving a few lines more for.
- The repetition here is mechanical and it's a big wall of text, but it is actually self-documenting and simple.

As I said, I can see both solutions have their merit though. What do you think?

christophruethingbmw: I was thinking the same, I personally also like explicit more typically. In this case I thought that it is really a lot of repeated code and the macro definition would be right at hand. 

Another option I see is a loop instead of a macro, not sure if this would improve readability. I am also fine with the longer version of writing down everything explicitly.

DominikAFischer: I think the loop approach is even less robust. It contains actually a lot of implicit assumptions (e.g. naming conventions) that could easily break and we don't have a way of easily enforcing.
So, all in all, I would still go with the explicit if that is alright with you.

christophruethingbmw (approved):

PR #597, opened 2026-09-07
Introduce suspend and resume for Async

christophruethingbmw:
Async acts as a lightweight abstraction layer for common RTOS functionality. Suspending and resuming a task is such a common functionality. It can be useful to explicitly disable tasks in case they are consuming too much runtime, accepting the penalty for single tasks but maintaining overall system stability.

### DIFF
--- a/executables/referenceApp/platforms/nucleo_f413zh/freeRtosCoreConfiguration/include/os/FreeRtosPlatformConfig.h
+++ b/executables/referenceApp/platforms/nucleo_f413zh/freeRtosCoreConfiguration/include/os/FreeRtosPlatformConfig.h
@@ -46,27 +46,23 @@ extern "C"
 #define configTIMER_TASK_STACK_DEPTH (configMINIMAL_STACK_SIZE)
 
 /* API function includes */
-#define INCLUDE_vTaskPrioritySet      1
-#define INCLUDE_uxTaskPriorityGet     1
-#define INCLUDE_vTaskDelete           1
-#define INCLUDE_vTaskCleanUpResources 1
-#define INCLUDE_vTaskSuspend          1
-#define INCLUDE_vResumeFromISR        1
-#define INCLUDE_vTaskDelayUntil       1
-#define INCLUDE_vTaskDelay            1
-#define INCLUDE_eTaskGetState         1
-#undef INCLUDE_uxTaskGetStackHighWaterMark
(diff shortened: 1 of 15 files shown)


executables/referenceApp/platforms/nucleo_f413zh/freeRtosCoreConfiguration/include/os/FreeRtosPlatformConfig.h
```diff
@@ -46,27 +46,23 @@ extern "C"
...
-#define INCLUDE_vTaskPrioritySet      1
-#define INCLUDE_uxTaskPriorityGet     1
-#define INCLUDE_vTaskDelete           1
-#define INCLUDE_vTaskCleanUpResources 1
-#define INCLUDE_vTaskSuspend          1
-#define INCLUDE_vResumeFromISR        1
-#define INCLUDE_vTaskDelayUntil       1
-#define INCLUDE_vTaskDelay            1
-#define INCLUDE_eTaskGetState         1
-#undef INCLUDE_uxTaskGetStackHighWaterMark
```
christophruethingbmw, line 58: These are already defined in the `FreeRTOSConfig.h` by async, therefore undefining and then redefining them with the same value makes not much sense, alternatively, we could also take them out into a separate block and comment it with this reason.

rolandreichweinbmw (approved):

libs/bsw/asyncThreadX/include/async/ThreadXAdapter.h
```diff
@@ -338,6 +342,18 @@ inline void ThreadXAdapter<Binding>::cancel(TimeoutType& timeout)
     }
 }
 
+template<class Binding>
+inline void ThreadXAdapter<Binding>::suspend(ContextType context)
+{
+    _taskContexts[static_cast<size_t>(context)].suspend();
```
christophruethingbmw, line 348: There is not bounds checking here to keep it consistent with the other implementations (e.g. `hreadXAdapter<Binding>::execute` or `hreadXAdapter<Binding>::schedule`. In case we want them we should do it separately in a common refactoring.

rolandreichweinbmw:
@christophruethingbmw and @matthiaskessler are discussing this concept separately.

matthiaskessler (approved):

PR #615, opened 2026-09-11
Hermetic Python toolchain for middleware codegen

DominikAFischer:
### Use rules_python 'library module with version constraints' pattern:
- A py_binary bundles CPython 3.12.3 and all pip packages (jinja2,
  jsonschema, pyyaml) as declared Bazel deps.
- Every dependency is a declared artifact, so remote caching and RE
  work correctly on any machine.
- Private Starlark rules are used instead of genrules so the tool
  attribute can declare cfg = "exec", keeping Python toolchain
  resolution on the host platform for all target configurations.
- Transparent to external Bzlmod consumers: does not interfere with
  their own Python toolchain configuration.
  
  ----

**NOTE: offline builds** 
The `bazel vendor` command can snapshot all external deps (CPython, pip wheels) for fully offline builds. Storage options:
- Baked into the Docker image
- Persistent volume mounted into the container
- Checked into the repo

**NOTE: security / supply chain:**
- All downloads are sha256-verified against committed lockfiles.
- Moving downloads into the Docker image build does not improve
  security. It only shifts when the network access occurs.
- Stronger guarantees require CVE scanning or a private artifact
  mirror.

**NOTE:** +737/-2,478 of the changes are attributed to the Bazel lockfile being changed


libs/bsw/middleware/tools/cpp_generator/middleware_codegen.bzl
```diff
@@ -10,71 +10,123 @@
...
 
+load("@bazel_skylib//lib:shell.bzl", "shell")
 load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
 load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
 
-# The generator package: script, templates and schemas are staged via this
-# filegroup, and the script/input dir are referenced by their execroot paths.
-_GENERATOR = "//libs/bsw/middleware/tools/cpp_generator:generator"
+_GENERATOR_TOOL = "//libs/bsw/middleware/tools/cpp_generator:jinja2cpp"
```
christophruethingbmw, line 24: Is there any advantage to define the variable here, it is only used in the private `_jinja2cpp` attribute of the rule?

libs/bsw/middleware/tools/cpp_generator/middleware_codegen.bzl
```diff
@@ -10,71 +10,123 @@
...
+
+# Templates and schemas staged as inputs so jinja2cpp.py can find them
+# at their source-tree paths inside the sandbox.
+_GENERATOR_TEMPLATES = "//libs/bsw/middleware/tools/cpp_generator:generator_templates"
+
+# Path passed to --input, which must match the package path of the generator so
+# jinja2cpp.py can locate templates/jinja/ and templates/schemas/ below it.
 _GENERATOR_INPUT = "libs/bsw/middleware/tools/cpp_generator"
```
christophruethingbmw, line 32: Hardcoding such a path feels a little bit unsafe? Can't we get this by a more defined way instead of just passing a string?

christophruethingbmw (commented):
Looks good to me :)

libs/bsw/middleware/tools/cpp_generator/middleware_codegen.bzl
DominikAFischer: It's used as the default in both of the rules `_codegen_srcs` and `_list_outputs`, so I would say having a shared variable is justified.

libs/bsw/middleware/tools/cpp_generator/middleware_codegen.bzl
DominikAFischer: Thank you for the suggestion. Deriving this from the generator's label like this makes more sense:
```python
_GENERATOR_INPUT = Label(_GENERATOR_TOOL).package
```

DominikAFischer: @christian-schilling Is this acceptable for you?

christophruethingbmw (approved):

christophruethingbmw:
in my opinion we are quite safe with all dependencies, even more than we are today, because the lock file precisely defines the hashes of all dependencies (the same way the requirements lockfile does), so we can make sure that we have a reproducible build. in case we really need offline build we can most likely (I did not test it out) pre-download all requirements to a docker image.

christian-schilling (approved):

PR #618, opened 2026-09-14
Fix Database search in skeleton and proxy

filipe-cuim:


rolandreichweinbmw (dismissed):

rolandreichweinbmw (approved):

rolandreichweinbmw:
Can you please elaborate about the actual fix? Why was NoClientsAvailable removed?

filipe-cuim:
> Can you please elaborate about the actual fix? Why was NoClientsAvailable removed?

Basically there can be several instance databases generated in the same common.cpp file. The instance databases are generated per cluster. If a cluster only contains proxy connections, it will still need the list of instanceIds that these proxies can connect to. This lead to a situation where when the skeleton tries to initialize it was stopping in the first instance database it found (in this case it was the one with only proxy connections). So I added the check in the skeleton to have a skeletonConnectionsRange != empty.
This way the skeleton will only select an instnace database with a skeleton connections range not empty, and with a valid instanceId.

This leads to the NoClientsAvailable becoming obsolete since it is not used anywhere else.

PR #620, opened 2026-09-14
Add pre-commit and gitlint configuration

simon-d-bmw:
## Purpose of this PR
- [x] New Feature
- [x] Documentation Update
- [x] Other: Developer tooling and pre-commit configuration

**Description**
Add a pre-commit configuration file (`.pre-commit-config.yaml`) and gitlint rules (`.gitlint`) matching the OpenBSW commit message guidelines.

Configure pre-commit hooks for file size, case conflict, merge conflict, YAML validation, `gitlint` (commit-msg stage), and `treefmt` (pre-commit stage). Sets `default_install_hook_types` so running `pre-commit install` enables both pre-commit and commit-msg hooks out of the box.

Update the developer guidelines in `doc/dev/guidelines/formatting/automatic.rst` and `doc/dev/guidelines/commit_message.rst` for automatic formatting and commit message validation.

**Related Issues**
None

**Breaking Changes**
- [ ] Yes
- [x] No

**Test Plan**
- Install pre-commit hooks: `pre-commit install`
- Run pre-commit stage checks: `pre-commit run --all-files`
- Run commit-msg gitlint check: `pre-commit run gitlint --hook-stage commit-msg --commit-msg-filename "$(git rev-parse --git-path COMMIT_EDITMSG)"`

**Regression Tests**
Have tests been added/updated? [ ] Yes [x] No

Tag: `no_hw_test_required`


.pre-commit-config.yaml
```diff
@@ -0,0 +1,40 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+default_install_hook_types:
+  - pre-commit
+  - commit-msg
+
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
```
matthiaskessler, line 16: What does this line do? Does it install sth. on the fly from the internet?

simon-d-bmw: This configuration is only loaded when you:
* installed pre-commit on your machine
* run pre-commit install

Then, it downloads the pre-commit hooks listed in lines 19-22 from this repository.

rolandreichweinbmw: We try to limit external downloads during development / build from external. Can that be done with explicit hooks? I guess it's only a few lines in the .git/hooks/ to install?

simon-d-bmw: pre-commit is only for local development, not for CI. 
The whole idea of pre-commit is to make the setup of git hook easier. Using explicit git hooks without pre-commit doesn't seem to bring benefits, because you still have to install the tools from somewhere, right?

rolandreichweinbmw: The hermeticity issue (i.e. prevent downloads) is not related to CI, but especially to the development use case.

Of course, we should make the user experience as comfortable as possible at the same time.

`pre-commit` is part of `pre-commit`, as available via `apt`, and the listed tools like `check-added-large-files` are part of `pre-commit-hook`. We could let the user install those 2 in their environment and the whole setup should work hermetically?

simon-d-bmw: Thanks for clarifying the background, Roland.

You're right that if we want **100% hermetic / offline execution with zero network calls**, we can use `repo: local` with `language: system` for all hooks.

Here is a quick breakdown of how that works and the trade-offs:

1. **Tool availability:**
   * `pre-commit` and `gitlint` are available via `apt` (or `pip`).
   * The standard check scripts (`check-added-large-files`, `check-yaml`, etc.) are part of the Python package `pre-commit-hooks` (`pip install pre-commit-hooks`), as there is no standalone `apt` package for them on Ubuntu.

2. **Options moving forward:**

   * **Option 1: Fully local (`repo: local`)**
     * We configure all hooks (`check-*`, `gitlint`, `treefmt`) as `repo: local` with `language: system`.
     * Pre-commit will **never** perform any network requests or clones.
     * We would add `pre-commit`, `pre-commit-hooks`, and `gitlint` to the development container (`Dockerfile` / `requirements.lock`) so it works out of the box inside Docker.

   * **Option 2: Standard pre-commit (`repo: https://github.com/...`) with pre-cached hooks**
     * Keeps remote hook definitions so external contributors don't need to manually install each tool in their global Python environment (pre-commit auto-provisions isolated venvs in `~/.cache/pre-commit/` once).
     * For hermetic environments/containers, we can run `pre-commit install-hooks` during the Docker image build to pre-populate the cache offline.

If the team prefers **Option 1 (fully local `repo: local`)**, I am happy to adjust `.pre-commit-config.yaml` and add the dependencies to the Docker setup. Let me know which direction you'd prefer!

rolandreichweinbmw: Regarding "Tool availability": The Debian and Ubuntu package "pre-commit-hooks" _is_ available.

Follows that "Options moving forward" includes "Option 0": Just install it via `apt` (locally, in docker , or CI) and use it.

simon-d-bmw: I have updated the setup to use **local-only hooks** (`repo: local` with `language: system`):

1. **`.pre-commit-config.yaml`**: All hooks (`check-added-large-files`, `check-case-conflict`, `check-merge-conflict`, `check-yaml`, `gitlint`, and `treefmt`) are now configured locally and will never attempt remote git clones or downloads.
2. **Development Container**: Added `pre-commit` and `gitlint` to the `apt-get` packages in `docker/development/Dockerfile`, and `pre-commit-hooks` to `docker/development/files/requirements.txt` / `requirements.lock`, so the full pre-commit tooling is available out of the box inside the container.

docker/development/Dockerfile
```diff
@@ -54,10 +54,12 @@ RUN apt-get update && apt-get install -y \
         g++-14 \
         gcc-14 \
         git \
+        gitlint \
         lcov \
         libc++-17-dev \
         libc++abi-17-dev \
         ninja-build \
+        pre-commit \
```
rolandreichweinbmw, line 62: Thank you for the update - that already helps!

How about updating above (top of the file) to image `ubuntu:resolute-20260912` or similar?

This one contains the package `pre-commit-hooks` already that we can add then here, besides `pre-commit`.

As a bonus, it is an Ubuntu LTS release.

rolandreichweinbmw: We can do it as a follow-up as well.

simon-d-bmw: ok, let's plan this as a follow-up PR. I guess more problems will pop up when doing this update.

rolandreichweinbmw (approved):

PR #623, opened 2026-09-17
Refactor Bazel platforms, constraints, config selection

DominikAFischer:
# Refactor Bazel platforms, constraints, config selection

## Problems with current design

- `bazel build //...` failed because test and production targets share one build graph. Test binaries pulled real FreeRTOS kernel symbols that only exist in the full reference app link.
- Several `label_flags` had no unit_test override so unit test builds used the real FreeRTOS kernel and real BSP.
- `alias + select()` defaults spread implementation selection across every library BUILD file. This scales poorly in case you want to add a new platform / app.
- `bsp_configuration` was defined in 8 local label_flags so a downstream consumer had to set 8 separate flags to override it.

## What changed

- New `build_variant` constraint with values `reference_app` and `unit_test` separates the build graphs. `posix_unit_test` inherits posix and adds explicit overrides for every label_flag that differs in unit_test. Mocks are reclassified from `testonly` to `target_compatible_with = [unit_test]` so they can be injected via label_flag while remaining excluded from production builds.

- Multiplexer alias defaults are replaced with `//bazel/config:no_implementation`, which is always incompatible. Each platform now owns its full implementation set via platform flags and no library BUILD changes are needed when adding a new platform. Platforms can define a parent platform to inherit its flags.

- `bsp_configuration` consolidated from 8 copies to a single `//bazel/config/bsp:bsp_configuration`.

## For downstream consumers

External consumers have multiple options to configure OpenBSW:
- The `build_variant` constraint is open so downstream projects can declare their own variant without modifying OpenBSW BUILD files.
- Feature flags, profiles and implementations can be overriden via label flags.
- Label flags can be set for example: 
   - using a custom platform mirroring the OpenBSW solution. 
   - globally via bazel.rc or even in a toolchain
   - via the commandline


bazel/platform/constraints/build_variant/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+# Separates production firmware builds from unit-test builds. Mock libs and
+# test targets declare incompatibility with reference_app so wildcard builds
+# skip them on production platforms.
+
+constraint_setting(
```
christophruethingbmw: Is there a reason we need the `build_variant` twice, once as `string_flag` and once as `constraint_setting`?

I imagine it is to use `target_compatible_with` to make sure targets are not used incorrectly in the wrong build, right?

I have/had two thinkings about this:

1. Since latest Bazel 9 version I think you can even use `config_setting`s inside `target_compatible_with`, at least it works for me locally. But since we do not want to only support Bazel 9 but also at least Bazel 8.6.0 (right now) this is not an option.
=> Instead, you could use a select on the config setting and map unsupported variants to `incompatible`. That's a little bit more writing but would work the same way.

2. Since Bazel 9 I think there was a change in the default test toolchain selection. There is the option `use_target_platform_constraints` introduced for the toolchain (see https://bazel.build/reference/be/platforms-and-toolchains#toolchain) which is set to true for the default test toolchain. This results in the fact, that the execution platform must fulfill all constraints of the target platform (since the default test has an implicit `exec_compatible_with` requiring the constraints of the target platform). This results in the fact, since our target platform specifies the `build_variant` constraint that the execution platform also need this. Since typically the execution platform is just the host, it does not have this set and then the test execution fails.
=> Thinking about this, I came to the conclusion that the `build_variant` is actually not really a constraint of the platform, it is a different kind of thing like `cpu` or `os`.
=> So my approach was to really not consider `build_variant` as a `constraint_setting` but only as a `string_flag`.
=> So I think we would run into the same issue here? Since setting the execution platform explicitly is quite uncommin in my opinion.

christophruethingbmw (commented):
Looks great, I applied exactly the same approach recently in another project and I think using platform flags and sepcifying the necessary `label_flags` but this is a quite readable and understandable approach :)

I just left a comment about some interesting (for me unexpected) observations which we might want to consider.

bazel/platform/constraints/build_variant/BUILD.bazel
DominikAFischer: Thank you for your feedback! Very interesting findings!
For 2:
>Thinking about this, I came to the conclusion that the build_variant is actually not really a constraint of the platform, it is a different kind of thing like cpu or os.

You're right, now that I think about it more. I will make the switch!

For 1:
I personally do not like these kinds of select blocks cluttered all over BUILD files. That said I understand that this is a real problem on Bazel 9.x.x.

Another solution that hides the select block would be to create helper rules:
```python
# bazel/config/build_variant/compatibility.bzl
def unit_test_only():
    return select({
        "//bazel/config/build_variant:unit_test": [],
        "//conditions:default": ["@platforms//:incompatible"],
    })

def reference_app_only():
    return select({
        "//bazel/config/build_variant:reference_app": [],
        "//conditions:default": ["@platforms//:incompatible"],
    }) 
```

then at the consumer side it would look like this:
```python
load("//bazel/config/build_variant:compatibility.bzl", "unit_test_only")
...
target_compatible_with = unit_test_only(),
```

instead of:
```python
target_compatible_with = select({
    "//bazel/config/build_variant:unit_test": [],
    "//conditions:default": ["@platforms//:incompatible"],
}),
```

Not sure what is really better here, open for both solutions.

bazel/platform/BUILD.bazel
```diff
@@ -31,6 +30,27 @@ platform(
         "--//bazel/config/features:transport=true",
         "--//bazel/config/features:uds=true",
         "--//bazel/config/features:watchdog=true",
+
+        # RTOS
+        "--//bazel/config/rtos=freertos",
```
christophruethingbmw: I would actually also set the `build_variant` flag here, although we set it to the default value. However, then we are more explicit and on the safe side in case the default value might be changed at some point in time.

bazel/platform/constraints/build_variant/BUILD.bazel
christophruethingbmw: I agree, I am fine with the implemented macros to simplify the usage in `target_compatible_with`. At some point in time when we migrate to bazel 9 or maybe bazel 8 get's this feature backportet, we can remove this.

PR #632, opened 2026-09-18
Add routing PDU <-> CAN Frame conversion methods

danielfaleirosantos:


libs/bsw/routing/include/routing/pduRouting.h
```diff
@@ -19,8 +19,36 @@
 
 #include <cstdint>
 
+namespace can
```
matthiaskessler: I think, this should go into a separate file, maybe a util file. The idea is, that the routing core itself works on PDUs only. The bus specifics should go into the adapters, which can even be project specific.

matthiaskessler (changes requested):

libs/bsw/routing/include/routing/pduRouting.h
danielfaleirosantos: Fine by me. This was done in our internal project and we agreed to open source this conversion, so it came from something project specific. I'll put it in a util file.

rolandreichweinbmw (approved):

